PR #2, opened 2015-12-16
Various fixes

bgamari:
I started to play around with this tonight against the current state of ghc's `master` branch. Here are fixes for a few issues I found.

I eventually stopped when I found that you expect hp2ps and genapply to both have Cabal files. I suspect we can make this happen but I need sleep.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,6 @@
+#!/bin/bash -e
+
+root=`dirname $0`
+mkdir -p $root/_shake
+ghc --make -Wall $root/src/Main.hs -i$root/src -rtsopts -with-rtsopts=-I0 -outputdir=$root/_shake -o $root/_shake/build
+$root/_shake/build --lint --directory $root/.. $@
--- a/cfg/system.config.in
+++ b/cfg/system.config.in
@@ -45,12 +45,12 @@ ar-supports-at-file = @ArSupportsAtFile@
 # Build options:
 #===============
 
-supports-package-key = @SUPPORTS_PACKAGE_KEY@
-solaris-broken-shld  = @SOLARIS_BROKEN_SHLD@
-split-objects-broken = @SplitObjsBroken@
-ghc-unregisterised   = @Unregisterised@
-ghc-source-path      = @hardtop@
-leading-underscore   = @LeadingUnderscore@
+supports-component-id = @SUPPORTS_COMPONENT_ID@
+solaris-broken-shld   = @SOLARIS_BROKEN_SHLD@
+split-objects-broken  = @SplitObjsBroken@
+ghc-unregisterised    = @Unregisterised@
+ghc-source-path       = @hardtop@
+leading-underscore    = @LeadingUnderscore@
 
 # Information about build, host and target systems:
 #==================================================
--- a/cfg/system.config.in
+++ b/cfg/system.config.in
@@ -116,3 +116,9 @@ iconv-lib-dirs     = @ICONV_LIB_DIRS@
 
 gmp-include-dirs   = @GMP_INCLUDE_DIRS@
 gmp-lib-dirs       = @GMP_LIB_DIRS@
+
+
+# Optional Dependencies:
+#=======================
+
+with-libdw         = @HaveLibdw@
--- a/src/Builder.hs
+++ b/src/Builder.hs
@@ -111,3 +111,4 @@ fixAbsolutePathOnWindows path = do
 -- Instances for storing in the Shake database
 instance Binary Builder
 instance Hashable Builder
+instance NFData Builder
--- a/src/GHC.hs
+++ b/src/GHC.hs
@@ -1,9 +1,9 @@
 module GHC (
     array, base, ghcBoot, binary, bytestring, cabal, compiler, containers,
     compareSizes, deepseq, deriveConstants, directory, dllSplit, filepath,
-    genapply, genprimopcode, ghc, ghcCabal, ghcPkg, ghcPrim, ghcPwd, ghcTags,
+    genapply, genprimopcode, ghc, ghcCabal, ghci, ghcPkg, ghcPrim, ghcPwd, ghcTags,
     haddock, haskeline, hsc2hs, hoopl, hp2ps, hpc, hpcBin, integerGmp,
-    integerSimple, mkUserGuidePart, parallel, pretty, primitive, process,
+    integerSimple, iservBin, mkUserGuidePart, parallel, pretty, primitive, process,
     runghc, stm, templateHaskell, terminfo, time, transformers, unix, win32, xhtml,
 
     defaultKnownPackages, defaultTargetDirectory, defaultProgramPath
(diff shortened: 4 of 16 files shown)


bgamari:
It also appears that you are expecting `ghc-cabal` to produce `LIB_NAME` fields in `package-data.mk`. Perhaps you have some `ghc` patches that have yet to be merged? We should make sure these get upstream soon if so.

bgamari:
Never mind, it seems these fields have been renamed.

bgamari:
For some reason `-j` appears to be broken. I'm seeing errors of the form,
```
No dependencies found for 'libraries/process/dist-install/build/libHSprocess-1.4.1.0.ghc7.11.20151216.so'.
```
but only when I build with `-jN` where N > 1.

bgamari:
For the record, the above depend upon some GHC commits that I'll try to push tomorrow.

snowleopard:
Awesome, many thanks Ben! Apologies about the current state of the codebase -- it's not in the best shape right now.

This forces me to switch from GHC 7.11 to a newer branch (I still have bin-package-db, LIB_NAME, etc). Obviously this must happen at some point, but the switch potentially exposes me to breakages unrelated to the build system. Any suggestions about the specific branch I should jump to, which is compatible with your changes and has no issues on Windows?

hvr:
@snowleopard what do you mean by "no issues on Windows"? afaik GHC HEAD builds fine on windows currently; if you want to avoid working on a moving target, maybe you could jump to the upcoming ghc-8.0 branch which by definition will be less of a moving target...

snowleopard:
@hvr Thanks, I wasn't sure about the current state of GHC HEAD. Yes, I think jumping to ghc-8.0 would be best.

@bgamari Do you think it will be possible to 'cabalise' hp2ps before ghc-8.0?

hvr:
> Do you think it will be possible to 'cabalise' hp2ps before ghc-8.0?

@snowleopard I doubt that... since that would require to rewrite `hp2ps` in Haskell... or alternatively artifically make a C program into a Haskell program which FFI-calls into the C programs `main` function... :stuck_out_tongue_winking_eye:

snowleopard:
@hvr Ah, it was a silly question on my part :) I forgot that `hp2ps` is a C program. I am currently working on a proper support for C-only packages for building `rts` (I am yet to implement build rules for it). It should also allow me to handle `hp2ps` more gracefully; at the moment I pretend that it has a cabal file.

bgamari:
For the record, these are the upstream changes which I was testing with,

 * https://phabricator.haskell.org/D1643
 * https://phabricator.haskell.org/D1642
 * https://phabricator.haskell.org/D1639

Also, @hvr has expressed concern that the fact that the shake build system running `configure` works against the usual mode of operation of autoconf and may be confusing to users. We could do something like this to address this,

 * https://phabricator.haskell.org/D1632

snowleopard:
Could you elaborate on the usual mode of operation of `configure` and `autoconf`? I'm not really familiar with these tools, so I assumed one would want to run them automatically when needed without bothering the user. Note, I modify the configure script by adding line 'AC_CONFIG_FILES([shake/cfg/system.config])' to handle the configuration settings.

We can certainly move `configure` and `autoconf` outside of shake or add a flag (initially switched off) that would enable the current functionality.

@bgamari I don't see how `D1632` is relevant to this question. Wrong link?

bgamari:
@snowleopard oops! Sorry about that,

 * https://phabricator.haskell.org/D1638

snowleopard:
Ah, that's perfect! This removes a temporary kludge I had in `Rules\Config.hs`. 

Is this the only problem with `configure` and `autoconf` or you meant something else?

hvr:
@snowleopard one problem is that Autoconf-gnostic developers know what to do when they see a `configure.ac` file. It's assumed that you just run `autoreconf` yourself to have a `configure` script generated. Fwiw, a non-generated `configure` can only happen when working directly from Git. In a released tar-ball you're not supposed re-generate the `configure` script anyway.

It's ok if the shake-buildsys is able to call `autoreconf` for you. However, it's not ok IMO, if you're not allowed to call `autoreconf` yourself anymore. Also the user ought to be allowed to call `./configure` and pass it the all sorts of env-vars and configure flags, and interact with --help.

Finally, `configure.ac` is tracked by Git, so you don't want to mutate it in-place.

bgamari:
I think with D1638 things will pretty much work the way you want them to.

snowleopard:
@hvr I no longer mutate `configure.ac` (see 5e7de75).

OK, I think I see a potential problem: if the user runs `configure` with special flags leading to a new `system.config` file, shake will undo this: it will rerun the rule for `system.config` executing `configure` without any flags and returning the system to a consistent state.

I added a flag `buildSystemConfigFile` initially set to `False` which guards the automatic invocation of `autoreconf/configure` (see 8ef67ed).

bgamari:
Right, for this reason it is probably best if we just make it the user's responsibility to run `configure`. Most users will already expect this to be necessary from their experience with the existing build system and I think in the case of build configuration there's no such think as too explicit. In my experience automagic behavior eventually blows up, no matter how safe it seems.

snowleopard:
@bgamari Regarding https://phabricator.haskell.org/D1638:

I forgot to mention that we actually decided not to put the new build system in `shake` folder because users may confuse it with the `shake` library. It is probably best to put all build-related stuff in folder `build`. Could you please rename `shake` to `build` in https://phabricator.haskell.org/D1638? I'll take care of all renamings on my side.

bgamari:
@snowleopard I'm a bit concerned that `build` may be a bit too general a name which may lead to a naming clash. Perhaps `shake-build`?

snowleopard:
@bgamari Yes, indeed let's stick to `shake-build`.

snowleopard:
@bgamari The commit you referenced uses `build` instead of `shake-build` -- it this a typo?

bgamari:

> @bgamari The commit you referenced uses `build` instead of `shake-build` -- it this a typo?
>
I'm about to push a commit updating this name.

bgamari:
@snowleopard the commit has been pushed. I'm now trying to work through the recent ghci changes.

snowleopard:
@bgamari Thanks! I noticed that in 5bee728 you didn't add `ghci` into `knownPackages` list. You need to do that, otherwise no rules for `ghci` will be generated. I'm unhappy about the fact that one need to add a new package to so many lists, but I might be able to fix this soon.

If you are currently busy with GHC 8.0, let me know I'll merge your pull request and fix the above myself.

bgamari:

> @bgamari Thanks! I noticed that in 5bee728 you didn't add `ghci` into
> `knownPackages` list. You need to do that, otherwise no rules for
> `ghci` will be generated. I'm unhappy about the fact that one need to
> add a new package to so many lists, but I might be able to fix this
> soon.
>
> If you are currently busy with GHC 8.0, let me know I'll merge your
> pull request and fix the above myself.

I actually have that fixed locally unfortunately there are further
issues. Indeed I am a bit busy with GHC 8.0 but I'll push my work. If
you want to work out what is going on that would be great!

Cheers,

- Ben

bgamari:
If you look at the `iserv-bin` issue, note that this package is handled a bit differently than others and in fact uses [different build directories](https://github.com/bgamari/ghc/blob/master/ghc.mk#L526) depending upon the way.

snowleopard:
Many thanks for your help Ben! I've merged this now.

I'll look into `ghci` & `iserv-bin` issues. It should be possible to allow different build directories depending on the way. The only difficulty for me is that `iserv` is not supposed to be build on Windows, but I'll try to mirror the `make` build system.

Issue #6, opened 2015-12-20
HsColour on GHC.Prim fails


bgamari:
I just encountered this while building `master`,
```
/--------
| Running GhcCabalHsColour with arguments:
|   hscolour
|   libraries/ghc-prim
|   dist-install
\--------
Running hscolour for ghc-prim-0.5.0.0...
Preprocessing library ghc-prim-0.5.0.0...
ghc-cabal: can't find source for GHC/Prim in ., dist-install/build/autogen
Error when running Shake build system:
* libraries/ghc-prim/dist-install/doc/html/ghc-prim/ghc-prim.haddock
user error (Development.Shake.cmd, system command failed
Command: /opt/exp/ghc/ghc/inplace/bin/ghc-cabal hscolour libraries/ghc-prim dist-install
Exit code: 1
Stderr:
ghc-cabal: can't find source for GHC/Prim in ., dist-install/build/autogen
)
```

snowleopard:
I tried to reproduce this on my Windows machine but `ghc-cabal hscolour libraries/ghc-prim dist-install` was successful, as well as all other invocations of `GhcCabalHsColour` builder.

I've pushed a small related fix to deal with absolute paths on Windows: 30d3d63. Currently `configure` sets the following path in `system.config`: `hscolour = /c/Users/[...]` which I replaced with `C:/Users/[...]` to be able to run `hscolour`. Maybe there is a better way to deal with this.

bgamari:
I can say that `Prim.hs` certainly does not exist where I would have expected it,
```
$ shake-build/build.sh
...
/--------
| Running GhcCabalHsColour with arguments:
|   hscolour
|   libraries/ghc-prim
|   dist-install
\--------
Running hscolour for ghc-prim-0.5.0.0...
Preprocessing library ghc-prim-0.5.0.0...
ghc-cabal: can't find source for GHC/Prim in ., dist-install/build/autogen
Error when running Shake build system:
* libraries/ghc-prim/dist-install/doc/html/ghc-prim/ghc-prim.haddock
user error (Development.Shake.cmd, system command failed
Command: /opt/exp/ghc/ghc/inplace/bin/ghc-cabal hscolour libraries/ghc-prim dist-install
Exit code: 1
Stderr:
ghc-cabal: can't find source for GHC/Prim in ., dist-install/build/autogen
)

$ ls libraries/ghc-prim/dist-install/build/autogen/
cabal_macros.h  Paths_ghc_prim.hs
[...]

Issue #5, opened 2015-12-20
Can't build ghc-cabal


bgamari:
The Shake build system is currently unable to build `inplace/bin/ghc-cabal` and therefore can't bootstrap a fresh source tree,
```
$ make distclean
$ ./boot
$ ./configure
$ shake-build/build.sh 
Reading shake-build/cfg/system.config...
Reading package dependencies...
Error when running Shake build system:
* ghc/stage1/build/tmp/ghc-stage1
* OracleQ (PackageDataKey ("ghc/stage1/package-data.mk","ghc_stage1_C_SRCS"))
* ghc/stage1/package-data.mk
* ghc/stage1/package-data.mk ghc/stage1/haddock-prologue.txt ghc/stage1/setup-config ghc/stage1/build/autogen/cabal_macros.h
* libraries/transformers/dist-boot/package-data.mk
* libraries/transformers/dist-boot/package-data.mk libraries/transformers/dist-boot/haddock-prologue.txt libraries/transformers/dist-boot/setup-config libraries/transformers/dist-boot/build/autogen/cabal_macros.h
* /opt/exp/ghc/ghc/inplace/bin/ghc-cabal
Error, file does not exist and no rule available:
  /opt/exp/ghc/ghc/inplace/bin/ghc-cabal
```

snowleopard:
Here is what is going on: I believe the Shake build system can build `utils/ghc-cabal/dist*/build/tmp/ghc-cabal`, but it just doesn't copy the resulting binary to `inplace/bin/ghc-cabal`.

At the moment this is a common issue for all program packages. I'll try to fix this today. I think it makes sense to place binaries directly to `inplace/bin` instead of building them in `[...]/build/tmp` first and then do the copying. I guess the old system had to do this because of the inability to properly handle dynamic dependencies with `make`. Please let me know if you know a reason to keep binaries in `[...]/build/tmp`.

Note, it is still unlikely that the build system will work on a fresh source tree after I do the fix because there are still some missing bits here and there. Currently the best way to hack on the build system is to run `make` first and then run the new build system on top of it.

snowleopard:
I've pushed the following fixes:
* Build some programs directly in `inplace/bin`: 663ad01.
* Bootstrap `ghc-cabal` via a fake `package-data.mk` file: c98eebc.

Please let me know if this works for you.

PR #8, opened 2015-12-20
Miscellany

bgamari:
Add details to readme, ignore `_shake` directory.


bgamari:
This ought to fix #6.

snowleopard:
Regarding 77d34c7:

Have a look at this commit: ddfe5bc. I decided to move the shake database inside `shake-build` and renamed it to `.db`. To make naming consistent I also renamed `_shake` to `.shake` (this also matches the naming convention for `.git` and `.arc-linters`. If you agree with this naming convention please amend 77d34c7.

snowleopard:
Regarding 576a06c:

I think this will become redundant once #5 is fixed. The `build` action already depends on the corresponding builder, so once I relocate all binaries to `inplace/bin` they will automatically be tracked.

bgamari:
Ahh, I see. Alright, perhaps I'll drop the `gitignore` change and "Documentation: Depend upon haddock for stage".

bgamari:
@snowleopard I think this is ready for another review.

snowleopard:
Excellent! This is now merged. I'll now look into #6.

PR #13, opened 2015-12-20
Consolidate box pretty-printing

bgamari:
And various other simple cleanups.


snowleopard:
Wonderful, many thanks!

I'll now switch to writing comments in proper Haddock syntax. Should have used it from start.

PR #17, opened 2015-12-21
[WIP] Make better use of types

bgamari:
Here is a patch which I've had in waiting. It's still a work-in-progress, but I thought I should at least make its existence know to solicit feedback on the approach.

It does two things,

 * Make `PackageName` into a proper type and add a number of associated types (e.g. `PackageId`)
 * Make `PackageData` into a typed interface

### DIFF
--- a/src/GHC.hs
+++ b/src/GHC.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE OverloadedStrings #-}
 module GHC (
     array, base, binary, bytestring, cabal, compiler, containers, compareSizes,
     deepseq, deriveConstants, directory, dllSplit, filepath, genapply,
@@ -103,11 +104,11 @@ defaultProgramPath :: Stage -> Package -> Maybe FilePath
 defaultProgramPath stage pkg
     | pkg == ghc     = Just . inplaceProgram $ "ghc-stage" ++ show (fromEnum stage + 1)
     | pkg == haddock || pkg == ghcTags = case stage of
-        Stage2 -> Just . inplaceProgram $ pkgName pkg
+        Stage2 -> Just . inplaceProgram $ pkgNameString pkg
         _      -> Nothing
     | isProgram pkg  = case stage of
-        Stage0 -> Just . inplaceProgram $ pkgName pkg
-        _      -> Just . installProgram $ pkgName pkg
+        Stage0 -> Just . inplaceProgram $ pkgNameString pkg
+        _      -> Just . installProgram $ pkgNameString pkg
     | otherwise = Nothing
   where
     inplaceProgram name = programInplacePath -/- name <.> exe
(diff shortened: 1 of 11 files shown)


snowleopard:
Many thanks, this looks excellent.

A couple of quick comments:
* `getPackageName (pkgName pkg)` is annoying, but I guess we can hide this in a function with a less tautological name.
* A large part of the build system is dedicated to dancing around `package-data.mk` files. My long term plan was to eliminate this and get rid of `ghc-cabal` altogether (see `Rules/Cabal.hs` where I use `Distribution.Package` instead). Therefore I'm not sure if we want to invest too much effort into prettifying this ugly corner of the codebase.

bgamari:
> getPackageName (pkgName pkg) is annoying, but I guess we can hide this in a function with a less tautological name.

I agree. `pkgNameString` or similar, perhaps.

>  My long term plan was to eliminate this and get rid of ghc-cabal altogether (see `Rules/Cabal.hs` where I use Distribution.Package instead). 

Ahh, I see. Indeed that is a very reasonable path forward.

snowleopard:
> I agree. pkgNameString or similar, perhaps.

Yes. By the way, it looks like one could also use `show pkg` in the current implementation. However, as I understand such `Show` instances are frowned upon (`show` is supposed to produce valid Haskell code?), so I guess we should derive Show instead.

snowleopard:
> Ahh, I see. Indeed that is a very reasonable path forward.

I added a new issue #18 to discuss potential ideas how to get rid of `ghc-cabal`.

bgamari:
> Yes. By the way, it looks like one could also use show pkg in the current implementation. However, as I understand such Show instances are frowned upon (show is supposed to produce valid Haskell code, right?), so I guess we should derive Show instead.

Indeed.

bgamari:
@snowleopard Should I rebase and clean this up so it can be merged?

snowleopard:
@bgamari Yes, please. Let's merge this.

bgamari:
Alright, here is a subset of the work. I've left the `PackageData` rework out for now, but I'll submit it soon.

bgamari:
I should note that I'm unable to test this at the moment due to #27.

snowleopard:
@bgamari Thanks! I'll do the testing after fixing #27. 

Could you please clarify what we gain by using the `OverloadedStrings` extension and deriving `IsString`? Does this allow us to pass a string, e.g. `"array"`, where a `PackageName` is expected?

bgamari:
> Could you please clarify what we gain by using the OverloadedStrings extension and deriving IsString? Does this allow us to pass a string, e.g. "array", where a PackageName is expected?

@snowleopard correct, it allows overloaded string literals.

snowleopard:
I see, thanks. At first I thought we were not actually using this, but I finally found ~50 uses in `GHC.hs` :-)

PR #25, opened 2015-12-23
Update README.md

angerman:
Adding missing prerequisites. These are probably installed if you build ghc often or use shake, but if not, these are missing.


snowleopard:
Thanks @angerman!

PR #34, opened 2015-12-24
Expression: Add Haddocks

bgamari:


snowleopard:
Awesome, many thanks @bgamari!

Issue #33, opened 2015-12-24
Simplify import conventions


bgamari:
Currently the codebase has many modules re-exporting other modules. Not only does this make it difficult for the developer to know the "proper" way to import a given definition, but it makes import cycles almost inevitable due to the more dense module dependency graph.

quchen:
A **very** big +1 to that.

snowleopard:
Sigh. 

I agree. I tried to minimise the import clutter at some point, but it just made things more complicated.

> I think a module should never export modules other than its own submodules.

So, is this proposal to get rid of re-exports completely or only partially?

bgamari:
I would say that re-exports should be limited to "children" of the re-exporting module. E.g. `Oracles` can re-export definitions from `Oracles.*`, but nothing else (without careful consideration, that is).

snowleopard:
@bgamari OK, this sounds reasonable. Would you like to give it a try?

bgamari:
Sure.

snowleopard:
I'll close this. Imports/reexports are much more sane now thanks to #36.

PR #36, opened 2015-12-24
Begin paring down reexports

bgamari:
This is the beginning of what appears to be a rather large undertaking. I began by eliminating some of the easier reexports. I expect we'll want to retain something like `Base`, which will act as a sort of project-local `Prelude` but otherwise re-exports will be limited to the sort of tree-like structures described in #33.

### DIFF
--- a/src/Base.hs
+++ b/src/Base.hs
@@ -12,7 +12,6 @@ module Base (
     module Development.Shake,
     module Development.Shake.Classes,
     module Development.Shake.FilePath,
-    module Development.Shake.Util,
 
     -- * Paths
     shakeFilesPath, configPath, sourcePath, programInplacePath,
@@ -38,7 +37,6 @@ import Data.Monoid
 import Development.Shake hiding (unit, (*>), parallel)
 import Development.Shake.Classes
 import Development.Shake.FilePath
-import Development.Shake.Util
 import System.Console.ANSI
 import qualified System.Directory as IO
 import System.IO
--- a/src/Base.hs
+++ b/src/Base.hs
@@ -39,7 +39,7 @@ import Data.Function
 import Data.List
 import Data.Maybe
 import Data.Monoid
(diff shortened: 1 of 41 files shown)


snowleopard:
> Base: Hide `parallel`
> Otherwise it is shadowed by `GHC.parallel`

Do we have different versions of Shake? As far as I can see `Development.Shake` doesn't export `parallel`: http://hackage.haskell.org/package/shake-0.15.5/docs/src/Development-Shake.html 

I already undid this change once, because I got a warning for hiding `parallel` unnecessarily.

bgamari:
Perhaps this is Linux specific? With `Shake-0.15.5`,
```
$ ghci
GHCi, version 7.10.2.20151118: http://www.haskell.org/ghc/  :? for help
λ> import Development.Shake
λ> :info parallel 
parallel :: [Action a] -> Action [a]
  	-- Defined in ‘shake-0.15.5:Development.Shake.Core’
```

bgamari:
The definition can be found here, https://github.com/ndmitchell/shake/blob/master/src/Development/Shake/Core.hs#L897

bgamari:
Ahh, my bad, I am actually using a git snapshot from a few days ago due to https://github.com/ndmitchell/shake/pull/365.

snowleopard:
Ah, that's the reason. Then it looks like we will be getting `parallel` at some point anyway, so let's hide it. I can live with a warning for a while.

snowleopard:
Speaking of `parallel`: do you know why we have packages like `parallel` in the GHC tree that we don't build? There are a few of them. If they are not dependencies of GHC it seems strange to have them in the codebase.

bgamari:
@snowleopard I believe this is for the testsuite and nofib.

snowleopard:
@bgamari Ah, I see! Thanks for clarifying. I haven't yet looked into testing at all.

snowleopard:
:+1:

Issue #43, opened 2015-12-26
Add CI


quchen:
Once we reach a half-decent state of stability, we should add continuous integration. This would provide us with a "standard" build that other builds can be compared to. Right now, we've got loosely coupled Windows, OSX, and Linux builds, and everyone is a little bit in his own bubble. Travis seems like the standard solution.

snowleopard:
Completely agree! I've never used Travis, so it'll take me some time to work it out. I'll appreciate any help with this.

angerman:
Looking at `shake`[1], we can apparently do `windows` ci on `appvoyer`?
Anyone experience with that?

[1]: https://github.com/ndmitchell/shake

snowleopard:
@angerman Thanks for the pointer. I'll look into this once more critical issues are out of the way.

angerman:
Maybe @quchen and @bgamari can take over? @quchen has this assigned already, and @bgamari has an open PR for travis.

To my understanding Travis can do Linux (and Mac - Beta Feature I think), and [appveyor](https://ci.appveyor.com) can do Window. So we should have the three most common platforms covered.
Not sure about architectures though.

quchen:
I'm only familiar with Travis, and @bgamari has done the required work already in his pull request. I've never used Windows CI.

snowleopard:
@quchen I believe @bgamari's pull request is still work in progress.

bgamari:
Andrey Mokhov <notifications@github.com> writes:

> @quchen I believe @bgamari's pull request is still work in progress.
>
Indeed it is. I haven't had time to keep poking away at it this week.

quchen:
#93 should do the trick for now.

PR #50, opened 2015-12-26
[WIP] Add Travis configuration

bgamari:
This is an attempt at resolving #43.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,22 @@
+sudo: false
+
+matrix:
+  include:
+    - env: CABALVER=1.22 GHCVER=7.10.3
+      addons: {apt: {packages: [cabal-install-1.22,ghc-7.10.3,zlib1g-dev,terminfo-dev],  sources: [hvr-ghc]}}
+
+before_install:
+ - export PATH=$HOME/.cabal/bin:/opt/ghc/$GHCVER/bin:/opt/cabal/$CABALVER/bin:$PATH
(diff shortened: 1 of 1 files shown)


snowleopard:
Many thanks. Let me know when it's ready for merge. 

I think it may be worth running a full `make` build before `./shake-build/build.sh`. Otherwise we are long way of seeing the build succeed.

angerman:
I'm in favor of omitting `make`, so that issues surface immediately and are not shadowed by prior `make` runs.
@bgamari can we run CI on Mac or Windows as well with Travis?

snowleopard:
Is it possible to have two different instances of Travis? There are benefits to both `from-the-scratch` and `from-make` builds.

bgamari:
Instead of doing a full `make` why don't we simply enumerate precisely which targets need to be built? Not only will this keep the build time down, but it also gives us a check-list for which functionality remains unimplemented.

angerman:
@bgamari :+1: I just never got the make system to only build certain targets, it always went ahead and built everything (or that's like how it felt).

snowleopard:
@bgamari Sounds good! Some targets are very close to the full build (e.g. the `ghc` package), but many are really small almost independent ones.

quchen:
I think this is now obsolete due to #93. @bgamari

snowleopard:
Thanks @bgamari and @quchen! I'll close this.

PR #52, opened 2015-12-26
Update LICENSE

angerman:
Be explicit about the license. Prevent others from having to lookup the license that matches this text.


snowleopard:
Fine. But I guess if one is really serious about licensing it is important to read it through, regardless of what the title says.

PR #61, opened 2015-12-30
Update Base.hs

angerman:
Tiny change to `renderBox` so that it looks a tiny bit rounder.

### DIFF
--- a/src/Base.hs
+++ b/src/Base.hs
@@ -164,9 +164,9 @@ renderBox ls =
                      , "│ "
                      , "╰──────────"
                      )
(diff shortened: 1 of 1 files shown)


snowleopard:
I gave it a try, but I don't like the asymmetry of this. Shall we postpone such cosmetic changes to a later stage? :-) By the way, I also have some changes of this sort; I keep them locally so that we could discuss all of them someday.

angerman:
Sure. Can do. I'll keep this locally for now. Just will have to make sure, it never ends up in a PR.

snowleopard:
:+1:

Issue #38, opened 2015-12-25
No support for `wrapper`


angerman:
We do not handle wrapper scripts at all so far. This I presume is the reason for #35 and maybe #27 as well?

How is this wrapping handled on Windows?

snowleopard:
It is a mystery for me why everything works on Windows without wrappers.

Can we get rid of wrappers? They seem to be just an unnecessary layer for massaging argument lists passed to a builder like `hsc2hs`. We can certainly achieve that in the new build system without any wrappers.

angerman:
Maybe @bgamari or @quchen can shed some light on the old build systems need for wrappers, might know who would know?

bgamari:
I would guess that the wrappers are necessary as we dynamically link on platforms other than Windows, meaning that we must set `LD_LIBRARY_PATH` to ensure that the dynamic linker can find the needed dynamic libraries.

angerman:
There seem to be the following wrapper scripts only in the ghc tree:
```
$ find . -name "*.wrapper"                                                                                                 
./ghc/ghc.wrapper
./utils/ghc-pkg/ghc-pkg.wrapper
./utils/haddock/haddock.wrapper
./utils/hpc/hpc.wrapper
./utils/hsc2hs/hsc2hs.wrapper
./utils/runghc/runghc.wrapper
```

I guess it boils down to figuring out if the build system initiates all invocations, or if other programs invoke the wrapped programs in question. If it's only the former case, shake could take care of providing the `LD_LIBRARY_PATH`, in the latter, we'll need them anyway.

snowleopard:
There is `rules/shell-wrapper.mk` which does a lot of things that can be moved to the new build system.

It also manipulates `LD_LIBRARY_PATH`, with help from `rules/library-path.mk`, but it seems strange that we can't deal with this without wrappers.

bgamari:
Keep in mind that you want to retain the ability to call these utilities outside of the build system. One way to accomplish this is to link executables with `-rpath` however I believe there may be portability issues here.

snowleopard:
@bgamari Yes, I see. We can at least minimise the amount of stuff that goes into wrappers. If we keep them only for passing `LD_LIBRARY_PATH` then they will be a lot easier to understand, maintain, and potentially remove.

I guess this issue is a prerequisite for #4.

angerman:
[...]

Issue #12, opened 2015-12-20
Rework `Package` to reflect structure of Cabal packages


bgamari:
The `Package` type currently doesn't really reflect the actual structure of Cabal packages. In particular `PackageType` forces a package to be either a `Library` or a `Program` whereas Cabal packages may be both. This works now but may become problematic in the future. Moreover, this requires that one explicitly define the type of each package in the build system.

One alternative would involve teaching `ghc-cabal` to emit more details about the components offered by the package, allowing us to read this information from `package-data.mk`.

angerman:
Wouldn't this relate to #18?

snowleopard:
@angerman No, this is separate. It's not clear how soon we want this. The build system can operate successfully without this, so I think we better not assign a milestone to this issue. As soon as someone presents a clean solution we'll merge this.

angerman:
Ohh. I'm sorry!

snowleopard:
No problem. Maybe @bgamari will pick this up once he gets some more spare time to hack on this project.

izgzhen:
Is this still relevant? What is the consequence of identifying `cabal` as a `Library` only?

alpmestan:
@hvr mentionned to me that he may need support for this soon and is looking into ways to have the make build system accept that. We might want to start discussing how we would achieve this ourselves, in Hadrian land :)

snowleopard:
@alpmestan Aha, interesting! I think it shouldn't be very difficult to support in Hadrian. If this becomes a priority, I'm happy to have a look at it.

Which package is this going to be?

alpmestan:
@hvr will hopefully jump in this thread at some point, but if I remember correctly, he was interested in simplifying the way we build haddock.

PR #62, opened 2015-12-30
Feature/wrapper

angerman:
This should help with #38, but is not yet ready to be pulled. There are a few issues, I'd appreciate help with:
- the `PackageWrapper` is a `String` for now, which is quite useless. I guess we would need something that can a) query the settings, and b) receive the actual binary name.
- I have not figured out how to obtain the `GhcSourePath` in the `Rules ()` to append to prepend to build the absolute path of the executable.

Also this feels like a very crude hack, hence any advice is appreciated.


snowleopard:
@angerman Thanks for giving it a try! Does the current implementation work for you? I guess it doesn't as I don't see any `-B` stuff yet.

> the `PackageWrapper` is a `String` for now, which is quite useless. I guess we would need something that can a) query the settings, and b) receive the actual binary name.

What is the intended purpose of `PackageWrapper`? Is this the full contents of a wrapper? If yes, this could be done similarly to how `Generators` work. Have a look at `Rules.Generators.ConfigHs`. 

> I have not figured out how to obtain the `GhcSourePath` in the `Rules ()` to append to prepend to build the absolute path of the executable.

It's not possible, because `GhcSourePath` cannot escape the `Action` monad. Why do you need an absolute path in `Rules ()`?

A couple of further comments:
* All this wrapper stuff should be disabled on Windows. The relevant condition to check is `windowsHost` from `Oracles.Config.Setting`.
* We need a data type `data Program = UnwrappedProgram | WrappedProgram Wrapper` or something similar to forbid cases when `pkgType == Library` and `pkgWrapper /= Nothing`.

angerman:
@snowleopard this **only** works, if I make the follow non-universal adjustments:
```
        let top = "/Users/angerman/Projects/haskell/ghc/"
        let wrapper = unlines [ "#!/bin/bash"
                              , "exec " ++ top ++ wrappedProgram ++ " -B" ++ top ++ " ${1+\"$@\"}"
                              ]
```
> It's not possible, because GhcSourePath cannot escape the Action monad. Why do you need an absolute path in Rules ()? 
I hope this becomes clear from the snipped above.

> What is the intended purpose of PackageWrapper? Is this the full contents of a wrapper? If yes, this could be done similarly to how Generators work. Have a look at Rules.Generators.ConfigHs.

Ideally the `PackageWrapper` should be in the `Action` monad, such that one can access the whole environment (e.g. read the `GhcSourcePath`, and potentially other information relevant to the build system). I envision the `PackageWrapper` to actually *build* the wrapped files content (e.g. produce what `wrapper` is above), that would give sufficient flexibility for all potential wrappers, I think.

> - All this wrapper stuff should be disabled on Windows. The relevant condition to check is `windowsHost` from `Oracles.Config.Setting`.

Yes, this clearly needs to boil down to a no-op on windows. I'm not sure how to do this elegantly.

> - We need a data type data `Program = UnwrappedProgram | WrappedProgram Wrapper` or something similar to forbid cases when `pkgType == Library` and `pkgWrapper /= Nothing`.

How about the following? Then I'm not so sure what you were debating about cabal package resemblance on IRC, which is why I went ahead with `pkgWrapper`.
```
data PackageType = WrappedProgram ProgramWrapper | Program | Library deriving Generic
```

snowleopard:
> I hope this becomes clear from the snipped above.

Hmm, why does the above snippet live in `Rules`? In `Rules` we only **define** targets or **match** targets, but this snippet **builds** the target, so it should live in `Action` where you have access to `GhcSourePath` and friends.

> Ideally the `PackageWrapper` should be in the `Action` monad, such that one can access the whole environment (e.g. read the `GhcSourcePath`, and potentially other information relevant to the build system). I envision the `PackageWrapper` to actually build the wrapped files content

Exactly! So why don't you turn `PackageWrapper` into an `Expr String`? Is this because of a potential import cycle?

> Yes, this clearly needs to boil down to a no-op on windows. I'm not sure how to do this elegantly.

OK, don't worry about this. I can add conditionals later.

> How about the following?
```haskell
data PackageType = WrappedProgram ProgramWrapper | Program | Library deriving Generic
```

I think this is semantically wrong: wrapped and unwapped programs are types of programs, not types of packages. We really should contain the wrapping flavour inside a `Program`.

> Then I'm not so sure what you were debating about cabal package resemblance on IRC, which is why I went ahead with `pkgWrapper`.

Ah, I see. That was about #12. At some point we might change `PackageType` from `Program | Library` to something like `PackageType [Program] [Library]` to reflect the fact that a single cabal package can mix programs and libraries. This shouldn't influence your work on wrappers.

angerman:
This now correctly generates the binary and wrapper for ghc-stage1. It's still a pretty big hack.

snowleopard:
Awesome!

Let me study your code and see whether it can be simplified if we switch to an alternative non-general solution.

snowleopard:
After experimenting with a couple of ways of adding wrappers I decided to take a non-general approach, see https://github.com/snowleopard/shaking-up-ghc/issues/38#issuecomment-168101977.

Many thanks for this pull request: I used it as a starting point, but merging it wasn't convenient.

PR #65, opened 2015-12-31
Fixes the -B path for the ghcWrapper.

angerman:
the -B option expects "inplace/lib", where takeDirectroy yields: "inplace/lib/bin"


snowleopard:
:+1:

PR #69, opened 2015-12-31
Improves matching filenames

angerman:
We currently only match on the file name. If multiple candidates exist,
this patch tries to compare their parent directry to find a better match.

### DIFF
--- a/src/Rules/Generate.hs
+++ b/src/Rules/Generate.hs
@@ -17,11 +17,6 @@ import Rules.Actions
 import Rules.Resources (Resources)
 import Settings
 
-import System.FilePath (takeBaseName, takeDirectory)
-
-parent :: FilePath -> String
-parent = takeBaseName . takeDirectory
-
 primopsSource :: FilePath
 primopsSource = "compiler/prelude/primops.txt.pp"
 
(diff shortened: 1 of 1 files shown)


angerman:
This is quite bad. Any suggestions / feedback how to do this better is welcome.

angerman:
Ideally I think the function should look like:
```
bestMatch :: FilePath -> [(FilePath, Builder)] -> (FilePath, Builder)
```

snowleopard:
@angerman Thanks! I'll experiment with this and see how your solution can be improved.

Does this solve the `unix` problem for you?

angerman:
Yes. Now stuck with `integer-gmp`, see #71.

angerman:
Improved the matching. We basically generate all possible `suffixes` and try to find the match with the longest suffix.

snowleopard:
@angerman Please test 6c80bd8. I think it is more robust because we know exactly what we match: generated file without the `buildPath` prefix. No need to try different possible suffixes. It's also simpler.

snowleopard:
Sorry, another small fix: 26cd11f.

snowleopard:
I think we can close this now.

PR #79, opened 2016-01-03
Adds Rules for IntegerGmp

angerman:
This should fix #71. We build the integer-gmp library similary to libffi now.

This *almost* works. We still need to call
```
./shake-build/build.sh libraries/integer-gmp/gmp/gmp.h
```
as I didn't know where to put the dependency into `integer-gmp` for the `gmp.h`.

And for some reason, we the include path doesn't contain `libraries/integer-gmp/gmp`, even though that's
what I thought
```
    includeGmp = ["-I" ++ pkgPath integerGmp -/- "gmp"]
```
would point to.

PR #80, opened 2016-01-03
Adds integer gmp path to the Gcc builder.

angerman:
This fixes the isse mentioned in #79, where `gmp.h` could not be found.


snowleopard:
:+1:

Issue #83, opened 2016-01-03
Implement support for unlit


angerman:
For #81 (`ghc-split`), we need #82 (`build-perl`), which in turn requires `unlit`. Hence we need `unlit`, the unliterizer tool (turns `lprl` literate perl files into `prl` files when used in `build-perl`).

snowleopard:
What is the purpose of `ghc-split`? Any chance to get rid of it and associated dependencies given that we now have the full power of Haskell in the build system?

snowleopard:
The `ghc-split.prl` file looks quite complicated, we probably don't want to do translate this in Haskell now. 

I think GHC uses `ghc-split` for splitting objects when executed with `-split-objs` flag.

angerman:
Yes. translating `ghc-split.lprl` looks like a non trivial task. Eep. to get it right :)

The best I found was this SO answer[1], which links to the GHC manual[2].

[1]: http://stackoverflow.com/questions/9198112/haskell-unnecessary-binary-growth-with-module-imports/9198223#9198223
[2]: https://downloads.haskell.org/~ghc/7.4.1/docs/html/users_guide/options-phases.html#id3081262

snowleopard:
This means there is a temporary workaround: you can set `splitObjects` to `return False` locally and see how far this will get you in the build.

angerman:
That's what I just did.

Now I'm stuck at `gmp` again(?). 
```
| Copy gmp.h -> libraries/integer-gmp/gmp/gmp.h
| Copy .libs/libgmp.a -> libraries/integer-gmp/gmp/libgmp.a
| Successfully build custom library 'integer-gmp'
/----------
| Running Gcc Stage1 (package = integer-gmp):
|     input: libraries/integer-gmp/cbits/wrappers.c
| => output: libraries/integer-gmp/stage1/build/cbits/wrappers.o
\----------

/----------
| Running Ld (stage = Stage1, package = integer-gmp):
|     input: libraries/integer-gmp/stage1/build/cbits/wrappers.o (and 5 more)
| => output: libraries/integer-gmp/stage1/build/HSinteger-gmp-1.0.0.0.o
\----------

shakeArgsWith         0.002s    0%                           
Function shake        0.028s    0%                           
Database read         0.320s    0%                           
Database compression  0.087s    0%                           
With database         0.009s    0%                           
Running rules       148.557s   99%  =========================
Total               149.002s  100%                           
Error when running Shake build system:
[...]

PR #85, opened 2016-01-03
Split libgmp.a

angerman:
Fixes the issue mentioned in #83
```
* utils/hpc/stage1/build/tmp/hpc-bin
* libraries/integer-gmp/stage1/build/libHSinteger-gmp-1.0.0.0.a
* getDirectoryFiles  [libraries/integer-gmp/gmp/objs/*.o]
libraries/integer-gmp/gmp/objs: getDirectoryContents: does not exist (No such file or directory)
```


snowleopard:
Thanks. We might be able to reuse the standard solution for running at least the `Ar` builder, but I'm not sure it is worth it.

This pattern seems useful too:
```haskell
builder <- pathBuilder Builder
cmd [builder]
```

PR #93, opened 2016-01-04
Travis CI using dirty tricks

quchen:
A rather hacky way to get CI going.

The problem is that Travis clones into a certain directory, but we'd like the Shake build system to be part of a second clone we have to do explicitly (GHC). The script does some dirty tricks to get it done™, but we should probably clean up the attempt a bit in the future.

CI build log: https://travis-ci.org/quchen/shaking-up-ghc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,32 @@
+sudo: false
+
+matrix:
+  include:
+    - env: CABALVER=1.22 GHCVER=7.10.3
+      addons: {apt: {packages: [cabal-install-1.22,ghc-7.10.3,zlib1g-dev,terminfo-dev],  sources: [hvr-ghc]}}
+
+before_install:
+
+    - PATH="$HOME/.cabal/bin:$PATH"
+    - PATH="/opt/ghc/$GHCVER/bin:$PATH"
+    - PATH="/opt/cabal/$CABALVER/bin:$PATH"
+    - export PATH
+
+    - .travis/print-env.sh
+
+install:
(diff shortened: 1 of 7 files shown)


snowleopard:
Cool, thanks! Let's try.

snowleopard:
It currently says "This is not an active repository": https://travis-ci.org/snowleopard/shaking-up-ghc

Do I have to do anything to activate it?

quchen:
@snowleopard Yes, you need to activate Travis. To do that, go to the Github repo -> (Project) Settings ->Webhooks & Services -> Add Service -> Travis CI -> Confirm. Your Github Repo now sends notices to Travis on each push. Now you need to let Travis know that you'd actually like to build the repo; for that, login on Travis-CI.org, go to the top right where your name is, click on it, then on Accounts. You should see a list of your projects; activate the one for GHC-Shake, and you should be good to go.

snowleopard:
@quchen Many thanks! I think it worked. At least the page is now alive: https://travis-ci.org/snowleopard/shaking-up-ghc.

snowleopard:
Right. Travis build failed at the very start:
```
* OracleQ (ConfigKey "host-os")
* shake-build/cfg/system.config
  shake-build/cfg/system.config
```
I believe the reason is the missing `./configure` step which produces `system.config` from `system.config.in`. Most likely `./boot` is missing too.

snowleopard:
OK, I attempted this: b978e17. Let's see if it works.

PR #101, opened 2016-01-05
Rework Travis CI script

quchen:
- Move all `.travis/*.sh` functionality directly into `.travis.yml`
- Install Alex/Happy from PPA instead of compiling them by hand
- Skip Shake progress indicators, as they're rather useless
  on clean-slate Travis builds, and clutter up the log
- Don't use a Cabal sandbox for building so Travis can share
  `$HOME/.cabal`
- `boot`/`configure` only after the Shake build system is in place,
  otherwise some required `*.in` files are not converted, and
  Shake fails


snowleopard:
:+1:

PR #108, opened 2016-01-05
[WIP] Use Cabal for package data

### LINKED ISSUE #18: Get rid of ghc-cabal and package-data.mk
A large part of the build system is dedicated to dancing around `package-data.mk` files containing package-related data (such as package name, version, dependencies, etc.). These files are generated by `utils/ghc-cabal` program, which also needs to be built in a non-trivial way. See `Rules/Data.hs` and `Oracles/PackageData.hs` in particular.

Our long term plan is to eliminate this and get rid of `ghc-cabal` and `package-data.mk` files altogether. See `Rules/Cabal.hs` where `Distribution.Package` is used instead of `ghc-cabal` to extract package dependencies directly from `.cabal` files.

A potential solution will need to be carefully thought through and discussed. This is the thread to do this.

Let me formulate the rationale behind this more clearly:
- _Separation of build artefacts from sources_: `ghc-cabal` creates `package-data.mk` files in a fixed location inside a package directory. This prevents us from moving build artefacts outside the source tree #113.
- _Performance_: `ghc-cabal` always runs expensive configure scripts, significantly affecting performance of the whole build system. On top of that, parallel invocations of `ghc-cabal` are currently broken.
- _Complexity_: Having a standalone `ghc-cabal` program built in a non-trivial way and `package-data.mk` files that need to be parsed significantly adds to the complexity of the build system and hence makes it more difficult to understand and maintain. Right now this is undoubtedly the most complicated and unreliable part of the build system.

bgamari:
This is a work-in-progress branch to fix #18 that I've had sitting around for lack of time. I believe much of the Cabal logic looks pretty good and it ought to produce everything that we need.

What it lacks is a good way to expose the various pieces of `PackageData` with an oracle such that we maintain minimal rebuilds.

I'm putting this here to provide a starting-point for someone wanting to take up this project before I am able to continue it myself.

### DIFF
--- a/src/Builder.hs
+++ b/src/Builder.hs
@@ -7,7 +7,9 @@ import Control.Monad.Trans.Reader
 
 import Base
 import GHC.Generics (Generic)
-import Oracles
+import Oracles.Config
+import Oracles.Config.Setting
+import Oracles.WindowsRoot
 import Stage
 
 -- | A 'Builder' is an external command invoked in separate process using 'Shake.cmd'
--- a/src/Oracles/GhcCabalData.hs
+++ b/src/Oracles/GhcCabalData.hs
@@ -1,0 +1,92 @@
+{-# LANGUAGE DeriveDataTypeable, GeneralizedNewtypeDeriving, GADTs #-}
+
+module Oracles.PackageData (
+    PackageData (..), PackageDataList (..),
+    pkgData, pkgDataList, packageDataOracle
+    ) where
+
+import Development.Shake.Config
+import Base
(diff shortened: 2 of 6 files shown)


snowleopard:
I think this is now on the critical path for finishing #113 properly, so I intend to give it a try tonight.

@bgamari it would be great if you could resolve merge conflicts.

snowleopard:
Never mind, I merged this into `ghc-cabal` branch. Now trying to compile.

I'll close this now -- many thanks again!

Issue #116, opened 2016-01-06
Switch off warnings during build by default.


angerman:
When building, we get *many* times warnings like:
```
This binding for '...' shadows the existing binding bound at ...
```
I think this should be a user configurable option to switch ann `-Wall` during ghc compilation.
Or even better on a per package option. As it is right now, we clutter the build log with massive
amounts of data, that is not necessarily related to the build. I don understand that one wants
all warnings, when hacking on ghc itself, but for the casual builder of ghc, this add's nothing,
but confusion, in my opinion; I also consider this to impact performance a little.

snowleopard:
I agree, we better switch warnings for now. However, the idiomatic way of doing this is as follows:
```haskell
userArgs = builderGhc ? remove ["-Wall", "-fwarn-tabs"]
```
I don't think we need a separate flag to control this.

@angerman Would you like to amend you PR #118?

PR #118, opened 2016-01-06
Removes -Wall and -fwarn-tabs

angerman:
-Wall and -fwarn-tabs flags to ghc can now be controlled from Settings/User.hs. This is a first stab at #116.


snowleopard:
@angerman Your last PR has some redundant code (`warnAll :: Predicate`, etc.), so I pushed my fix instead.

PR #120, opened 2016-01-06
Add IRC notifications

quchen:
... also cache `$HOME/.ghc`, maybe it speeds things up a bit.


snowleopard:
Let's chat to Travis on irc!

PR #122, opened 2016-01-06
Add documentation, small refactorings

quchen:
I tried documenting a couple of low-hanging fruits here.

One thing I could not wrap my head around is documentation of `WayUnit`s fields, I'd appreciate if someone more knowledgeable could have a look at that. `Way` confuses me!

### DIFF
--- a/src/Predicates.hs
+++ b/src/Predicates.hs
@@ -7,49 +7,60 @@ module Predicates (
 import Base
 import Expression
 
--- Basic predicates
+-- | Is the build currently in the provided stage?
 stage :: Stage -> Predicate
 stage s = fmap (s ==) getStage
 
+-- | Is a particular package being built?
 package :: Package -> Predicate
 package p = fmap (p ==) getPackage
 
--- For unstaged builders, e.g. GhcCabal
+-- | Is an unstaged builder is being used such as /GhcCabal/?
 builder :: Builder -> Predicate
 builder b = fmap (b ==) getBuilder
 
--- For staged builders, e.g. Ghc Stage
+-- | Is a certain builder used in the current stage?
 stagedBuilder :: (Stage -> Builder) -> Predicate
-stagedBuilder sb = (builder . sb) =<< getStage
+stagedBuilder stageBuilder = do
+    s <- getStage
+    builder (stageBuilder s)
 
+-- | Are we building with GCC?
 builderGcc :: Predicate
 builderGcc = stagedBuilder Gcc ||^ stagedBuilder GccM
 
+-- | Are we building with GHC?
 builderGhc :: Predicate
 builderGhc = stagedBuilder Ghc ||^ stagedBuilder GhcM
 
+-- | Does any of the output files match a given pattern?
 file :: FilePattern -> Predicate
 file f = fmap (any (f ?==)) getOutputs
 
+-- | Is the current build 'Way' equal to a certain value?
 way :: Way -> Predicate
 way w = fmap (w ==) getWay
 
--- Derived predicates
+-- | Is the build currently in stage 0?
 stage0 :: Predicate
 stage0 = stage Stage0
 
+-- | Is the build currently in stage 1?
 stage1 :: Predicate
 stage1 = stage Stage1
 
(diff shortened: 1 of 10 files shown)


snowleopard:
This looks great, many thanks. One thing I'd like to do a bit differently is `src/Main.hs`: can we use fully qualified names for imports? I think `Rules.Cabal.cabalRules` would look better than `RCabal.cabalRules`.

I'll do a full revision a bit later.

quchen:
Fully qualified names were a bit long when I tried them, but if you like it we can do that as well. The main point here was getting rid of the comments, because they described something easily expressible by code (and checked by the compiler).

snowleopard:
> The main point here was getting rid of the comments, because they described something easily expressible by code (and checked by the compiler).

Yes, absolutely, that's definitely a win on all accounts. I think fully qualified names make this even better as one doesn't need to look up the imports.

src/Predicates.hs
```diff
@@ -7,49 +7,60 @@ module Predicates (
...
 notStage0 :: Predicate
 notStage0 = notM stage0
 
+-- | Is a certain package /not/ built right now?
 notPackage :: Package -> Predicate
 notPackage = notM . package
 
--- TODO: Actually, we don't register compiler in some circumstances -- fix.
+-- | TODO: Actually, we don't register compiler in some circumstances -- fix.
 registerPackage :: Predicate
```
snowleopard: TODO should probably be outside haddock. The haddock comment can be:
-- | Do we need to run `ghc-pkg update` on the currently built package? See `Rules.Data.buildPackageData`

src/Target.hs
```diff
@@ -12,50 +16,53 @@ import Package
...
+--
+-- * the empty computation is the identity element of the underlying type
 -- * two computations can be combined by combining their results
 instance Monoid a => Monoid (ReaderT Target Action a) where
     mempty  = return mempty
     mappend = liftM2 mappend
 
--- PartialTarget is a partially constructed Target with fields Stage and
--- Package only. PartialTarget's are used for generating build rules.
+-- A partially constructed Target with fields 'Stage' and 'Package' only.
```
snowleopard: Should be "-- | A partially..." (missing '|')?

src/Way.hs
```diff
@@ -72,11 +80,23 @@ instance Read Way where
 instance Eq Way where
     Way a == Way b = a == b
 
-vanilla, profiling, logging, parallel, granSim :: Way
+-- | Build with no 'WayUnit's at all.
+vanilla :: Way
 vanilla   = wayFromUnits []
```
snowleopard: Why extra spaces? Also see logging, parallel, granSim.

snowleopard:
I've added some comments. I think your description of `WayUnit` is pretty close to how I understand it, although I have to admit I'm also a bit hazy on this topic. It's likely ways will be reworked in future.

Many thanks again!

quchen:
Just FYI, you can merge this whenever the build passes and you're happy with the changes.

I'll keep adding commits in the same spirit and rebase it onto your master periodically.

snowleopard:
There are some merge conflicts at the moment.

snowleopard:
I think the conflicts are because of this commit: c573af0.

@ndmitchell Could you please use pull requests, so I can keep track of what is happening? Otherwise you seem to have pushed exactly what @quchen has proposed in this PR :)

ndmitchell:
Opps, sorry for that! Yep, I'll go through pull requests. Perhaps remove my commit permissions? It's very easy to accidentally send to the wrong branch, and far easier to have GitHub stop me.

ndmitchell:
One easy option might be to revert my patch (just `git revert c573af0`), and then this branch should merge cleanly once more.

snowleopard:
@ndmitchell No problem! Yes, I'll do the revert. 

How do I remove your commit permissions without removing from collaborators? I'd like to keep you in collaborators because I can then assign issues to you :-)

snowleopard:
This is now merged.

ndmitchell:
Don't know how assigning really works - I don't think I've ever done it. I will just be careful then.

PR #123, opened 2016-01-06
Feature/fix clang

angerman:
Lookup raw commands in PATH.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,8 +51,7 @@ install:
     - ( cd ghc/shake-build && cabal install --only-dependencies )
 
     - ( cd ghc && ./boot )
-    - if [ $TRAVIS_OS_NAME == osx ]; then ( cd ghc && ./configure --with-gcc="$(which clang)"); fi
-    - if [ $TRAVIS_OS_NAME == linux ]; then ( cd ghc && ./configure ); fi
+    - ( cd ghc && ./configure )
 
 script:
     - ./ghc/shake-build/build.sh -j --no-progress
--- a/README.md
+++ b/README.md
@@ -31,9 +31,6 @@ git submodule update --init
 git clone git://github.com/snowleopard/shaking-up-ghc shake-build
 ./boot
 ./configure
-# or if you want to use clang (e.g. building on OS X)
-./configure --with-gcc=$(which clang) # See #26
-
 ```
 
 Now you have a couple of options:
--- a/shaking-up-ghc.cabal
+++ b/shaking-up-ghc.cabal
@@ -114,6 +114,7 @@ executable ghc-shake
                        , extra >= 1.4
                        , mtl >= 2.2
                        , shake >= 0.15
+                       , split >= 0.2
                        , transformers >= 0.4
                        , unordered-containers >= 0.2
     default-language:    Haskell2010
--- a/shaking-up-ghc.cabal
+++ b/shaking-up-ghc.cabal
@@ -22,12 +22,12 @@ executable ghc-shake
                        , Expression
                        , GHC
                        , Oracles
-                       , Oracles.AbsoluteCommand
                        , Oracles.ArgsHash
                        , Oracles.Config
                        , Oracles.Config.Flag
                        , Oracles.Config.Setting
                        , Oracles.Dependencies
+                       , Oracles.LookupInPath
                        , Oracles.ModuleFiles
                        , Oracles.PackageData
                        , Oracles.PackageDeps
--- a/shaking-up-ghc.cabal
+++ b/shaking-up-ghc.cabal
@@ -22,6 +22,7 @@ executable ghc-shake
                        , Expression
                        , GHC
                        , Oracles
+                       , Oracles.AbsoluteCommand
                        , Oracles.ArgsHash
                        , Oracles.Config
                        , Oracles.Config.Flag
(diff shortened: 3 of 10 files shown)


snowleopard:
Looks very nice, thanks! 

Unfortunately this doesn't go through yet:
```
* inplace/bin/ghc-pkg
* inplace/lib/bin/ghc-pkg
* OracleQ (PackageDataKey ("utils/ghc-pkg/stage0/package-data.mk","utils_ghc-pkg_stage0_C_SRCS"))
* utils/ghc-pkg/stage0/package-data.mk
* libraries/binary/stage0/package-data.mk
Command: which 
Exit code: 1
Stderr:
)
```

angerman:
Awesome. So $(which) is not available. Ahh well. Let's try something else then.

angerman:
@snowleopard the above issue is fixed, now mac fails with a `DerivedConstants.h` issue.

snowleopard:
@angerman There are a couple of issues with your approach:
* You are performing expensive lookupInPath on every invocation of a builder. This needs to be cached (that's why windowsRoot is an oracle).
* I don't think we need to depend on `split` library just for the sake of using `splitOn` here: we can do something like `words . replaceEq ':' ' '` instead.

Have a look at https://github.com/snowleopard/shaking-up-ghc/issues/124#issuecomment-169610804. Maybe this will solve the `clang` issue?

angerman:
We can clearly cache the lookupInPath! Will get at that.
Now replacing `:` with ` ` runs afoul for the following example:
```
PATH=/usr/bin:"/path/to/my binary folder"
```
I though `split` was fine as it was just a hidden package. I'll check that it doesn't introduce any additional dependency.

I'm afraid the fix with `specified` does not resolve the problem. As we check for `null`, but the value is `clang`.

In general, I think it's good to verify that the builder we use are actually *on* the PATH. Otherwise calling them is bound to fail anyway, hence we can extend them as well.

Another (cache less) solution would be to resolve the builder during initialization.

snowleopard:
> Now replacing : with  runs afoul for the following example:

Ah, what about `lines . replaceEq ':' '\n'`?

> In general, I think it's good to verify that the builder we use are actually on the PATH. Otherwise calling them is bound to fail anyway, hence we can extend them as well.

Yes, I think this is a good thing to do.

> Another (cache less) solution would be to resolve the builder during initialization

Could you elaborate on this?

angerman:
> Ah, what about lines . replaceEq ':' '\n'?
Well, as apparently everything except for `:` and `NUL` is allowed, that looked like a hack :-) Hope I found a better solution.

I've added the oracle, which was not so easy to wrap my head around, but I finally managed to!

>> Another (cache less) solution would be to resolve the builder during initialization
> Could you elaborate on this?

Sure. So we are (right now) doing this to find the tools autoconfig gave us in their fully expanded path. We are only considering elements from `system.config`. So we could parse `system.config` once, fix the missing paths and carry on.

But I guess that's pretty much identical to the oracle solution we have now, so the additional complexity might not be justified.

snowleopard:
> Well, as apparently everything except for : and NUL is allowed, that looked like a hack :-) Hope I found a better solution.

Nice one! I think `wordsWhen` is useful enough to be added to `Base.hs`. We may also need to rename `renameIf` to `renameWhen` for consistency. Do you know if `When` is a standard suffix in such cases?

> I've added the oracle, which was not so easy to wrap my head around, but I finally managed to!

Looks good. A couple of nitpicky comments:
* Shouldn't `absoluteCommand` have type `absoluteCommand :: FilePath -> Action FilePath`? We are dealing with `FilePath`s here.
* Why don't we call this `lookupInPathOracle`? This follows the convention used in other oracles and avoids having two names for essentially the same concept: `lookupInPath` and `absoluteCommand`.

> But I guess that's pretty much identical to the oracle solution we have now, so the additional complexity might not be justified.

I agree, let's use oracles.

angerman:
I'll give those changes a go after dinner. For some unrelated reason the mac build is still failing.

snowleopard:
@angerman See f187ca8. I think this fixes the `DerivedConstants.h` issue.

angerman:
> Nice one! I think wordsWhen is useful enough to be added to Base.hs. We may also need to rename renameIf to renameWhen for consistency. Do you know if When is a standard suffix in such cases?
I'll move this into Base. But will retain the name. It's technically `words` with a supplied predicate.

snowleopard:
> I'll move this into Base. But will retain the name. It's technically words with a supplied predicate.

Thanks. What I wonder is whether `replaceIf` should also become `replaceWhen`. I think it should (although it's not exported, so not big deal really). By the way, your `wordsWhen` function can probably be generalised to `(a -> Bool) -> [a] -> [[a]]`.

Please also change the module name, newtype, etc. `AbsoluteCommand` to `LookupInPath` for consistency.

angerman:
This compiles for me :)
And it's actually quite fast:
```
/------------------------------------------------------------------------------\
| Successfully built program 'ghctags' (Stage2).                               |
| Executable: inplace/bin/ghctags                                              |
| Program synopsis: A simple generator of vi- and emacs-compatible TAGS files. |
\------------------------------------------------------------------------------/
shakeArgsWith                         0.002s    0%                           
Function shake                        0.027s    0%                           
Database read                         0.006s    0%                           
With database                         0.000s    0%                           
Running rules                       800.131s   99%  =========================
Pool finished (25172 threads, 8 max)  0.000s    0%                           
Lint checking                         0.083s    0%                           
Total                               800.248s  100%                           
Build completed in 13:21m
```

snowleopard:
Nice! :-) Also Travis seems to be doing well. Let's wait for it to finish too.

angerman:
I'm sure travis will fail ultimately as this PR branch doesn't include the latest changes.

snowleopard:
OK, let me merge then.
Thanks again!

PR #136, opened 2016-01-08
Fix unnecessary import hiding of "parallel"

quchen:
This has been issuing "unnecessary import" warnings for quite some time for me.


snowleopard:
@quchen OK, let's fix this. This was actually added by @bgamari who uses a newer version of Shake which does export `parallel`. So I guess we will have to bring this back at some point in future.

Issue #84, opened 2016-01-03
-split-objs not supported (yet)


angerman:
I just want to hold this here. We currently do *not* support the `-split-objs` feature of build `ghc`s. 
This depends on #83, #82 and #81. See also the Linkage Optimization [1] item in the wiki.

[1]: https://wiki.haskell.org/HSFFIG/Linkage_optimization

angerman:
I belive, over the long run we will need to support `-split-objs` and all the dependencies that come with it: `unlit`, `perl`, ...

snowleopard:
I added a proper way to control `splitObjects` in `Settings/User.hs`:
```haskell
-- To switch off split objects change to 'return False'
splitObjects :: Predicate
splitObjects = return False -- FIXME: should be defaultSplitObjects, see #84.
```

PR #138, opened 2016-01-09
Feature/UtilUnlit

### LINKED ISSUE #82: build-perl not supported
We do not yet support the `build-perl` macro, e.g. used for `driver/split`. See also: #81

angerman:
Fixes #82, #83, and #84.

But I’m running into an issue with the split objs.

```
/usr/bin/ar: createProcess: runInteractiveProcess: exec: resource exhausted (Argument list too long))
```

I’m also fairly certain this needs some adaptations for Windows.

I’ll leave this here for review.

### DIFF
--- a/src/GHC.hs
+++ b/src/GHC.hs
@@ -6,7 +6,7 @@ module GHC (
     haddock, haskeline, hsc2hs, hoopl, hp2ps, hpc, hpcBin, integerGmp,
     integerSimple, iservBin, libffi, mkUserGuidePart, parallel, pretty,
     primitive, process, rts, runGhc, stm, templateHaskell, terminfo, time,
-    touchy, transformers, unix, win32, xhtml,
+    touchy, transformers, unlit, unix, win32, xhtml,
 
     defaultKnownPackages, defaultTargetDirectory, defaultProgramPath
     ) where
@@ -28,7 +28,7 @@ defaultKnownPackages =
     , ghcTags, haddock, haskeline, hsc2hs, hoopl, hp2ps, hpc, hpcBin, integerGmp
     , integerSimple, iservBin, libffi, mkUserGuidePart, parallel, pretty
     , primitive, process, rts, runGhc, stm, templateHaskell, terminfo, time
-    , touchy, transformers, unix, win32, xhtml ]
+    , touchy, transformers, unlit, unix, win32, xhtml ]
 
 -- Package definitions (see "Package")
 array, base, binary, bytestring, cabal, compiler, containers, compareSizes,
@@ -37,7 +37,7 @@ array, base, binary, bytestring, cabal, compiler, containers, compareSizes,
     haddock, haskeline, hsc2hs, hoopl, hp2ps, hpc, hpcBin, integerGmp,
     integerSimple, iservBin, libffi, mkUserGuidePart, parallel, pretty,
     primitive, process, rts, runGhc, stm, templateHaskell, terminfo, time,
-    touchy, transformers, unix, win32, xhtml :: Package
+    touchy, transformers, unlit, unix, win32, xhtml :: Package
 
 array           = library  "array"
 base            = library  "base"
@@ -85,11 +85,12 @@ terminfo        = library  "terminfo"
 time            = library  "time"
 touchy          = utility  "touchy"
 transformers    = library  "transformers"
+unlit           = utility  "unlit"
 unix            = library  "unix"
 win32           = library  "Win32"
 xhtml           = library  "xhtml"
 
--- TODO: The following utils are not implemented yet: unlit, driver/ghc-split
+-- TODO: The following utils are not implemented yet: driver/ghc-split
 -- TODO: The following utils are not included into the build system because
 -- they seem to be unused or unrelated to the build process: checkUniques,
 -- completion, count_lines, coverity, debugNGC, describe-unexpected, genargs,
@@ -112,8 +113,11 @@ defaultProgramPath stage pkg
     | pkg == haddock || pkg == ghcTags = case stage of
         Stage2 -> Just . inplaceProgram $ pkgNameString pkg
         _      -> Nothing
-    | pkg == touchy = case stage of
+    |  pkg == touchy = case stage of
         Stage0 -> Just $ "inplace/lib/bin" -/- pkgNameString pkg <.> exe
+        _      -> Nothing
+    | pkg == unlit = case stage of
+        Stage0 -> Just $ "inplace/lib" -/- pkgNameString pkg <.> exe
         _      -> Nothing
     | isProgram pkg = case stage of
         Stage0 -> Just . inplaceProgram $ pkgNameString pkg
--- a/src/GHC.hs
+++ b/src/GHC.hs
@@ -113,11 +113,8 @@ defaultProgramPath stage pkg
     | pkg == haddock || pkg == ghcTags = case stage of
         Stage2 -> Just . inplaceProgram $ pkgNameString pkg
         _      -> Nothing
-    |  pkg == touchy = case stage of
+    | pkg `elem` [touchy, unlit] = case stage of
         Stage0 -> Just $ "inplace/lib/bin" -/- pkgNameString pkg <.> exe
-        _      -> Nothing
-    | pkg == unlit = case stage of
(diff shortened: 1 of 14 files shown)


ndmitchell:
For Windows, long ar lines need to use an @foo file - a file where the arguments are written one per line. withTempFile in Shake makes it pretty easy.

angerman:
> For Windows, long ar lines need to use an @foo file - a file where the arguments are written one per line. withTempFile in Shake makes it pretty easy. 
@ndmitchell Thanks for the hint, let's see if I can find that somewhere in the existing `make` system.

ndmitchell:
The other (far worse) solution is to merge the files piece by piece, so you create an ar from the first 100, then merge the next 100. No idea what GHC currently does, we use @ at work.

angerman:
@ndmitchell this whole split objects business seems shaky to me anyway. But as ghc allows for split objc, I think `shaking-up-ghc` should as well.

src/GHC.hs
```diff
@@ -112,9 +113,12 @@ defaultProgramPath stage pkg
     | pkg == haddock || pkg == ghcTags = case stage of
         Stage2 -> Just . inplaceProgram $ pkgNameString pkg
         _      -> Nothing
-    | pkg == touchy = case stage of
+    |  pkg == touchy = case stage of
```
snowleopard: Accidental extra space

src/GHC.hs
```diff
@@ -112,9 +113,12 @@ defaultProgramPath stage pkg
...
         Stage0 -> Just $ "inplace/lib/bin" -/- pkgNameString pkg <.> exe
         _      -> Nothing
+    | pkg == unlit = case stage of
+        Stage0 -> Just $ "inplace/lib" -/- pkgNameString pkg <.> exe
```
snowleopard: The old build system put `unlit` into `inplace/lib/bin`, same as `touchy` (not `inplace/lib`).

src/Rules/Generate.hs
```diff
@@ -80,7 +81,7 @@ compilerDependencies stage =
 
 generatedDependencies :: Stage -> Package -> [FilePath]
 generatedDependencies stage pkg
-    | pkg   == compiler = compilerDependencies stage
+    | pkg   == compiler = compilerDependencies stage ++ ["inplace/lib/bin/ghc-split"]
```
snowleopard: `["inplace/lib/bin/ghc-split"]` should better be moved to `compilerDependencies`, so we have everything in one place.

src/Rules/Generate.hs
```diff
@@ -177,6 +185,31 @@ generateRules = do
...
+    -- this is a hack :-/
+    "//*.prl" %> \out -> do
+        let src = out -<.> "lprl"
+        path <- builderPath Unlit
+        need [path]
+        unit $ cmd [path] [src] [out]
+
+    -- ghc-split is only a perl script.
+    let ghcSplit = "inplace/lib/ghc-split" -- See system.config
+    let ghcSplitBin = "inplace/lib/bin/ghc-split" -- See ConfigHs.hs
```
snowleopard: Why the two stage process? The old build system appears to only keep `inplace/lib/bin/ghc-split`?

src/Rules/Generators/GhcSplit.hs
```diff
@@ -0,0 +1,25 @@
...
+import Base
+import Expression
+import Oracles
+import Settings.User
+
+generateGhcSplit :: Expr String
+generateGhcSplit = do
+    let yesNo = lift . fmap (\x -> if x then "YES" else "NO")
+    perl <- getBuilderPath Perl
+    let script = "driver" -/- "split" -/- "ghc-split.prl"
```
snowleopard: In the rest of the codebase we simply write `"driver/split/ghc-split.prl"`. We use forward slashes everywhere, even on Windows, so I don't see a benefit using `-/-` here.

src/Rules/Generate.hs
```diff
@@ -177,6 +185,31 @@ generateRules = do
   where
     file <~ gen = file %> \out -> generate out emptyTarget gen
 
+-- | Generate scripts the build system requires. For now we generate the
+-- @ghc-split@ script from it's literate perl source.
+generateScripts :: Rules ()
+generateScripts = do
+    -- how to translate literate perl to perl.
+    -- this is a hack :-/
+    "//*.prl" %> \out -> do
```
snowleopard: I think the rules responsible for working with `perl` should live in a separate file `Rules/Perl.hs`. Semantically they are not generators, they are invocations of a builder `Unlit` on a source file to produce an output file.

src/Rules/Generate.hs
```diff
@@ -177,6 +185,31 @@ generateRules = do
...
+        let src = out -<.> "lprl"
+        path <- builderPath Unlit
+        need [path]
+        unit $ cmd [path] [src] [out]
```
snowleopard: There is a `runBuilder` function which does this, see `Rules/Actions.hs`.

src/Rules/Generate.hs
angerman: That got me very confused as well. `cfg/system.config` has `unlit` in `inplace/lib`. Yet we need it in `inplace/lib/bin` as well. Something fishy is going on here :(

src/Rules/Generators/GhcSplit.hs
angerman: Good to know. Was wondering about the use of `"x/y/z"` and `"x" -/- "y" -/- "z"`.

snowleopard:
@angerman Many thanks, this is a big one!

I added a bunch of comments. If you don't feel up to handling some of them (e.g. moving rules around), let me know and I'll be happy to do the changes myself.

angerman:
> I added a bunch of comments. If you don't feel up to handling some of them (e.g. moving rules around), let me know and I'll be happy to do the changes myself.

I'll get right at it. Please advice on the `inplace/lib`, `inplace/lib/bin` issue, which get's me really confused, because there seem to be two valid places.

I'll update with the fixes to `ar` from #141.

snowleopard:
> Please advice on the `inplace/lib`, `inplace/lib/bin` issue, which get's me really confused, because there seem to be two valid places.

What do you mean? On Windows, `ghc-split`, `touchy`, and `unlit` all live in `inplace/lib/bin`.

angerman:
> What do you mean? On Windows, `ghc-split`, `touchy`, and `unlit` all live in `inplace/lib/bin`.

This is what my `system.config` reads:
```
unlit            = inplace/lib/unlit
ghc-split        = inplace/lib/ghc-split
```

snowleopard:
Ah, I see! This is just a mistake. Add `bin` to unlit and remove the `ghc-split` line altogether (we don't register it as a `Builder`).

angerman:
Ugh. It's hardcoded, I should have checked that. I assumed it was `@UnlitCmd@`. Alright.

snowleopard:
Or I should have said, _we shouldn't register it as a Builder_. Remove corresponding lines from `Builder.hs` too (we never call `ghc-split` directly).

snowleopard:
> Ugh. It's hardcoded, I should have checked that.

By the way, we should refactor this at some point. No need to carry hardcoded paths through `system.config` files.

angerman:
Alright, I hope I caught all issues.

snowleopard:
Thanks! Please have a look at this line in `Settings/Packages.hs`:
```haskell
    , notM windowsHost ? notM (anyHostOs ["ios"]) ? append [terminfo] ]
```
I thought `anyHostOs ["ios"]` does what you need for `osxHost` function, but apparently I was wrong. But then it's unclear: does the above line in `Settings/Packages.hs` make sense? It is a translation of this `ghc.mk` snippet:
```
ifeq "$(Windows_Host)" "NO"
ifneq "$(HostOS_CPP)" "ios"
PACKAGES_STAGE0 += terminfo
endif
endif
```
I guess `ios` stands for the mobile operating system iOS: do we ever build GHC on iOS?

angerman:
I'll give this a look tomorrow. We *can* build iOS cross compiler thanks to @luxi.

angerman:
See also: https://github.com/ghc-ios/ghc-ios-scripts

snowleopard:
@angerman Cool! I'll add `iosHost` function to tidy up things.

PR #140, opened 2016-01-09
Use Extra library

### LINKED ISSUE #130: Use the extra library
A number of functions in Base.hs are very similar to ones in the extra library, for example chunksOfSize is [chunksOf](https://hackage.haskell.org/package/extra-1.4.3/docs/Data-List-Extra.html#v:chunksOf), replace is there as [replace](https://hackage.haskell.org/package/extra-1.4.3/docs/Data-List-Extra.html#v:replace), wordsWhen is [wordsBy](https://hackage.haskell.org/package/extra-1.4.3/docs/Data-List-Extra.html#v:wordsBy). I suggest you use the extra library because:

* It's a dependency of Shake, so you don't increase the transitive set of things that the build system depends on.
* It has barely any dependencies itself, so doesn't cause issues, e.g. with old-time or similar.
* It has lots of useful methods you might not have thought of abstracting, or which might not be useful for just this one project.
* Everything is tested and (to some extent) optimised.

If you decide to use it, I'll send a few pull requests switching things over. If you decide against it then that's fine too - keeping dependencies down is a good plan.

angerman:
- replaces `wordsWhen` with `wordsBy`
- replaces `replace` with `replace`
Fixes #130

### DIFF
--- a/src/Base.hs
+++ b/src/Base.hs
@@ -22,9 +22,9 @@ module Base (
     putColoured, putOracle, putBuild, putSuccess, putError, renderBox,
 
     -- * Miscellaneous utilities
-    bimap, minusOrd, intersectOrd, replaceEq, replace, quote, chunksOfSize,
+    bimap, minusOrd, intersectOrd, replaceEq, quote, chunksOfSize,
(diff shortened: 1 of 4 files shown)


snowleopard:
:+1:

Issue #110, opened 2016-01-06
CI: Windows


angerman:
Setup CI for Windows builds using [AppVeyor](https://www.appveyor.com), which is Travis for Window I think.

snowleopard:
OK, I'll look into this. Judging by @ndmitchell's `appveyor.yml` script for Shake this is not very difficult.

snowleopard:
On a second thought, Shake is tested on Windows natively, while GHC is built in MSYS2. I'll google for possible CI solutions for Windows + MSYS2.

ndmitchell:
Appveyor comes with MSYS2, and even if it doesn't, you can install it yourself - I don't think that will be a problem.

ndmitchell:
In addition, Stack sets up MSYS2 properly and installs a GHC compiler - I don't think it should be too much more work to have Stack know how to compile GHC in some way. There's a ticket at https://github.com/commercialhaskell/stack/issues/1051 which might turn out to be the easiest route, and has lots of other benefits.

For info, if you did fix the Stack integration, things like https://github.com/ndmitchell/shake/issues/400 would be a lot less scary for me :)

snowleopard:
> Appveyor comes with MSYS2

Cool, I'll check it out. My familiarity with CI is about zero, so it might take a while. But it's useful to learn.

> In addition, Stack sets up MSYS2 properly and installs a GHC compiler

Right, that's another interesting approach. But then I don't know much about Stack either (but I heard it's awesome). It will also take some time to learn :-)

ndmitchell:
I might be able to help on the Stack side, since that's probably the way I'll have a go at following your instructions.

snowleopard:
Found this on `appveyor` website (https://www.appveyor.com/pricing):
```
> Are there any build time restrictions?
All plans have maximum build job execution time of 60 minutes.
```
At the moment we are already over 60 minutes on Travis, and we are still not building a lot of stuff (only vanilla way, no split objects, no haddock, etc.), so I guess `appveyor` is not an option.

Thoughts?

ndmitchell:
Can you build a subset? Perhaps just stage0? That gives you some confidence that it works a bit.

angerman:
I'd say, we should still give it a shot. Maybe it's faster on their machines? And if that still doesn't work, maybe talking to them helps ;)

snowleopard:
OK, agreed. We'll try a full build first and then may start pruning what we build.

[...]

PR #142, opened 2016-01-09
CI: Shallow clone GHC from Github instead of Haskell.org

quchen:
The shallow clone brings down cloning GHC to 10 megs instead of almost 100, giving us a small CI speedup.

(For reasons unknown to me, the Haskell.org Git instance does not support shallow clones.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,12 @@ install:
     - alex --version
     - happy --version
 
-    - travis_retry git clone git://git.haskell.org/ghc --recurse-submodules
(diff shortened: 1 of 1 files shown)


snowleopard:
Thanks, looks good. Could you clarify why we need the `git config --global` stuff?

snowleopard:
Once I get the appveyor build going (#110) I'll use this to accelerate it as well.

quchen:
Github and Haskell.org have different package naming conventions, and the config settings are necessary so that submodules can be initialized correctly. https://ghc.haskell.org/trac/ghc/wiki/Building/GettingTheSources#CloningfromGitHub

snowleopard:
I see, thanks!

Issue #143, opened 2016-01-09
Turn off variable shadowing warning


ndmitchell:
I've always felt this warning was actively harmful. Looking at the code, I can see places where you have actively harmed your code to adhere to it :(

    digest list = case list of
        []  -> "none"
        [x] -> x
        xs  -> head xs ++ " (and " ++ show (length xs - 1) ++ " more)"

In any normal code I would expect `digest xs`, but then you get name shadowing - a harmless situation, but which leads to a warning.

I appreciate this is personal preference, but I always turn off this warning.

snowleopard:
To be honest I never found it difficult to fix this warning if need be. I do agree that in most cases it doesn't actually highlight an error, although on rare occasions I found it useful.

In the above example, I find that `digest list` actually reads better, but only marginally, so I don't mind to accept the change you proposed in #144.

> I've always felt this warning was actively harmful.

Can you elaborate? I may underestimate the harm. My current understanding is that it only forces us to hide unused imports, which is indeed annoying.

ndmitchell:
I often use variable shadowing to reduce the possibility of error. For example, in digest, if you use xs as the scrutinee and the test then you can guarantee in the alternative that you don't refer to the wrong variable, since it's shadowed. It's even more useful in large monadic blocks to avoid things you have "updated".

I generally find it to be annoying and make my code less robust, so always turn it off. But it's ultimately up to your preferences and I'll follow.

angerman:
The two sides of this coin we are discussing are the following, if I understand correctly:

- One the one side (and I think this is the intended warning for the warning), you want to prevent using variables you think refer to something in a more global scope but you accidentally shadowed it by a local binding you might not have noticed to have done so actually. In this case it prevents a case of hard to debug errors, but also requires that both bindings are of the same type!
- On the other side, we have the issue of explicitly finding new names for local variables that would be more concise using a common naming scheme. Or the inverse case of the first side, which @ndmitchell alludes to, in that you *want* the local binding, but accidentally obtain the global binding because you forgot to rename your variable in your subscope.

If you stick to the idea of having a consistent naming scheme, you should not run afoul of the second point, as a binding will always refer to the most recent one. But then the shadowing warning makes no sense, as shadowing is part of your style.

I don't see a clear stronger side here. Both do not prevent you from using the wrong name. The first one though keeps *all* bindings in scope, while the second one can purposefully reduce the number of bindings in scope by actively using shadowing.

[...]

Issue #113, opened 2016-01-06
Move build products out of the ghc tree


angerman:
We've discussed this in #32 already when debating how to properly reset the build. So far we have depended on the build products being built _inside_ of the ghc tree, such that we could potentially use makes build artifacts during the build. As we now can build on Windows, OS X and Linux.

snowleopard:
Agreed, it's time to start moving things around. Let me take care of this.

snowleopard:
I moved most build artefacts to `buildRootPath = ".build"` which can be changed in `Settings/User.hs`. An example command invocation: 
```
/-----------------------------------------------\
| Run Ghc Stage0 (package = ghc)                |
|     input: compiler/main/GHC.hs               |
| => output: .build/stage0/compiler/build/GHC.o |
\-----------------------------------------------/
```
This is still work in progress and there are a few temporary hacks.

The remaining bits are: `gmp`, program binaries, package configuration folders such as `libraries/bootstrapping.conf`, generated files in `includes`, and files produced by `ghc-cabal` (e.g. `package-data.mk`). Should be able to finish this tomorrow, although the last one seems tricky. We may first need to tackle the `ghc-cabal` issue #18.

snowleopard:
Coming across some limitations unfortunately. For example:
```
Exit code: 1
Stderr:
In file included from compiler\nativeGen\AsmCodeGen.hs:14:0:

compiler\HsVersions.h:23:0:
     fatal error: ../includes/ghcautoconf.h: No such file or directory
compilation terminated.
```
For some reason `compiler\HsVersions.h` includes a generated file directly `../includes/ghcautoconf.h`, not relying on `-I` paths. This forces me to keep a copy of generated file in `includes` folder.

Does anyone know the reason for `#include "../includes/ghcautoconf.h"`? Shall this be fixed in GHC?

snowleopard:
And another one :(
```
Exit code: 1
Stderr:
compiler\main\DynFlags.hs:127:0:
     fatal error: ../includes/dist-derivedconstants/header/GHCConstantsHaskellExports.hs: No such file or directory
compilation terminated.
```

snowleopard:
Another mystery. When invoking the following command:
```
inplace/bin/ghc-stage1.exe
-hisuf hi
-osuf o
-hcsuf hc
-static
-hide-all-packages
-no-user-package-db
-this-package-key rts
-i
-i.build/stage1/rts/build
-i.build/stage1/rts/build/autogen
[...]

PR #144, opened 2016-01-09
General refactorings

ndmitchell:
These patches first set up the code so it can work with my style of development (shake HEAD, ghcid).

Next I refactor the use of extra a bit. It seems that generally the approach is to make Base export all of the module, so I did that, fixing up warnings etc (most of which were down to variable shadowing, see #143).

It's usually a bad idea to import directly from `Extra`, but instead of things like `Data.List.Extra`. The real purpose of `Extra` is to let people see all the functions at once. With this patch that no longer happens.

As a future question, what's your preference - lots of small branches, or just one big branch with my consolidated changes in?

### DIFF
--- a/src/GHC.hs
+++ b/src/GHC.hs
@@ -8,12 +8,10 @@ module GHC (
     primitive, process, rts, runGhc, stm, templateHaskell, terminfo, time,
     touchy, transformers, unlit, unix, win32, xhtml,
 
-    defaultKnownPackages, defaultTargetDirectory, defaultProgramPath
+    defaultKnownPackages
     ) where
 
-import Base
 import Package
-import Stage
 
 -- | These are all GHC packages we know about. Build rules will be generated for
 -- all of them. However, not all of these packages will be built. For example,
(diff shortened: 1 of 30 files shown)


snowleopard:
Thanks! It's all fine, just one comment.

It looks like you are using a non-released version of Shake that exports `parallel`, which forced you to add some qualifications. I'd prefer to hide `parallel` when it's going to come in official version, not to add confusion to the collision of GHC.parallel and Way.parallel, which I'd prefer to keep unqualified whenever possible.

I'm not hiding `parallel` yet as it causes an annoying warning.

snowleopard:
> As a future question, what's your preference - lots of small branches, or just one big branch with my consolidated changes in?

I don't have a strong opinion on this, but I generally prefer to review PR's that are focused on one particular issue (like this one). Not sure this answers your question really, so please clarify if need be.

ndmitchell:
Yep, I appreciate you'd prefer the hiding of parallel - but the Shake release this is the only warning free form that works with both released and HEAD Shake, which seems valuable (as I'm not the only one who has hit it).

ndmitchell:
The alternative is to hide parallel and just disable the warning in that module. It's a smaller change when we upgrade to a new Shake.

snowleopard:
> The alternative is to hide parallel and just disable the warning in that module. It's a smaller change when we upgrade to a new Shake.

Sounds reasonable. I let you add this change and I'm ready to merge the PR.

src/GHC.hs
```diff
@@ -4,7 +4,7 @@ module GHC (
     deepseq, deriveConstants, directory, dllSplit, filepath, genapply,
     genprimopcode, ghc, ghcBoot, ghcCabal, ghci, ghcPkg, ghcPrim, ghcTags,
     ghcSplit, haddock, haskeline, hsc2hs, hoopl, hp2ps, hpc, hpcBin, integerGmp,
-    integerSimple, iservBin, libffi, mkUserGuidePart, parallel, pretty,
+    integerSimple, iservBin, libffi, mkUserGuidePart, GHC.parallel, pretty,
```
bgamari: IMHO it would be better to just hide `parallel` from the `Shake` import and accept the warning for a while than uglify the export lists for eternity.

ndmitchell:
Done, as well as #143 - all good to go now I think (after the CI passes it)

ndmitchell:
It's failing the tests - @snowleopard, something that looks like my regression, or something you are aware of already?

ndmitchell:
Both buildbots are failing with:

    * inplace/bin/ghc-stage1.exe
    * OracleQ (PackageDataKey (".build/stage0/ghc/package-data.mk",".build_stage0_ghc_C_SRCS"))
    * .build/stage0/ghc/package-data.mk
    * .build/stage0/compiler/package-data.mk
    * includes/dist-derivedconstants/header/GHCConstantsHaskellExports.hs
    * inplace/bin/deriveConstants.exe
    * .build/stage0/utils/deriveConstants/build/Main.o
      .build/stage0/utils/deriveConstants/build/Main.o

snowleopard:
Thanks @ndmitchell! I'm pretty sure the failure is my fault (consequences of work in progress on #113). I'll have another look and merge.

ndmitchell:
I added another patch in - now you can run `shake-build/build.bat` without first changing into `shake-build`.

snowleopard:
:+1:

snowleopard:
@ndmitchell Apparently it was your fault! :-)
This change:
```diff
-        units  = map uniqueReads . words . replaceEq '_' ' ' $ s
+        units  = map uniqueReads . splitOn "_" $ s
```
The difference appears on the empty string:
```haskell
splitOn "_" "" == [""]
(words . replaceEq '_' ' ') "" == []
```

ndmitchell:
Thanks for finding that! I think I owe you a test (see #151) and a beer.

snowleopard:
No problem, and thanks for adding tests! Regarding beer, I guess now I just owe one fewer? Or maybe we are both in luck because beers are non-cancellable ;)

ndmitchell:
Indeed, beers are non-rehypothecable :)

snowleopard:
> non-rehypothecable
Can't even pronounce this without a beer :)

PR #151, opened 2016-01-11
Add tests

ndmitchell:
Only real decision was whether to put the Arbitrary instances in the same module or in a separate module, with orphan instances. Generally orphans should be avoided, but here we're writing an exe and the testing stuff is very much incidental, so I used orphans to keep the main system simple and permit avoiding QuickCheck if we need to.

I added tests for chunksOfSize, and they promptly failed taking up all memory and looping forever. So I rewrote chunksOfSize and now it works and passes all the tests.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -30,4 +30,5 @@ install:
 
 build_script:
     - cd C:\msys64\home\ghc\shake-build
+    - echo "" | stack --no-terminal exec -- build.bat selftest
     - echo "" | stack --no-terminal exec -- build.bat -j --no-progress inplace/bin/ghc-stage1.exe
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,6 +62,7 @@ install:
 
 script:
     - ( cd ghc/shake-build && cabal haddock --internal )
+    - ./ghc/shake-build/build.sh selftest
     - ./ghc/shake-build/build.sh -j --no-progress $TARGET
 
 cache:
--- a/shaking-up-ghc.cabal
+++ b/shaking-up-ghc.cabal
@@ -100,6 +100,7 @@ executable ghc-shake
                        , Settings.Ways
                        , Stage
                        , Target
+                       , Test
                        , Way
 
     default-extensions:  BangPatterns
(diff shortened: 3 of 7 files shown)


ndmitchell:
I added the scripts on appveyor but not Travis, since it wasn't immediately obvious where to do so.

angerman:
@ndmitchell I'd put them between `L64` and `L65`, https://github.com/snowleopard/shaking-up-ghc/blob/master/.travis.yml#L63-L65

ndmitchell:
@angerman - done, that seems right.

snowleopard:
Many thanks, that looks great! I agree, it seems orphan instances is the right way to go here.

I'm yet to grok the new implementation of `chunksOfSize` though :-) Looks impressive. Do you know it's time & memory complexity? This function may sometimes get very large inputs, so if the complexity is quadratic we might want to optimise it.

ndmitchell:
Should be O(n) time just like the old one, and also O(n) space. If the inputs were producing streaming (which they aren't) and consumed streaming (which they aren't) then this one would force a whole chunk to be in memory at once, but otherwise they are equivalent.

    chunksOfSize n = repeatedly f
        where f xs = splitAt (max 1 $ length $ takeWhile (<= n) $ scanl1 (+) $ map length xs) xs

`repeatedly` is just do this operation until you have consumed everything. `splitAt` is simple, just peel off the first number. `max 1` is to make sure we always peel of at least one and thus terminate. `length` is to convert the list coming in back to an Int for the splitAt - if we had lazy Int's that would be identical to the solution before, but since we have strict Int that forces each chunk. `scanl1` just gives us the running total, and takeWhile keeps taking from that total while we are below the size. In a strict language the repeated `map length xs` would give us O(n^2), but in Haskell we're fine.

snowleopard:
I see, very nice! Many thanks for the explanation.

ndmitchell:
I decided to blog the explanation: http://neilmitchell.blogspot.co.uk/2016/01/a-simple-haskell-function.html

snowleopard:
@ndmitchell `chunksOfSize` turned out to be surprisingly interesting :-)

PR #152, opened 2016-01-11
Report success on IRC only if the build was fixed

quchen:
We've reached a somewhat stable level of green-ness on Travis, so I think we can reduce the number of messages to the more relevant ones.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,7 +72,7 @@ cache:
 
 notifications:
     irc:
-        on_success: always # always/never/change
+        on_success: change # always/never/change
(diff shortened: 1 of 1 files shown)


snowleopard:
> Pull requests that have a failing status can’t be merged from a phone.

Wat.

snowleopard:
Thanks!
(Had to pretend I'm on a desktop.)

ndmitchell:
I regularly suffer from the failing test requires desktop - most annoying to switch modes each time. No good way around it I've found.

Issue #134, opened 2016-01-08
Advanced render box styles


angerman:
Currently, when using render box, we are bound to use five lines for each invocation. I propose we abstract render box in a way such that we can either render the box as is right now, or produce a single line condensed output. E.g.
```
> Run Ghc Stage1 (package = gcc-bin) ghc/stage1/build.o (and 5 more) => inplace/lib/bin/ghc-stage2
```
And preferably let this be controlled through a `--style=short` or `--style=box` or similar. That way we could also reduce the logs travis produces.

snowleopard:
Yay, the output style inventions continue! :+1:  
If we combine this proposal with @bgamari's one in #92 we get three potential styles already:
* `--style=normal`: the current implementation, default.
* `--style=brief`: the single line approach proposed here.
* `--style=pony`: @bgamari's messenger pony mode.

@angerman Feel free to self-assign :-)

angerman:
It will need some rewiring though, such that the semantics of the box are captured to be formatted according to style.

snowleopard:
I edited my comment: we actually have 4 styles if #135 is taken into account. I also think `--build-info` is probably a better name (this can be finalised later of course).

angerman:
The big issue with `none` (#135) is, that you still want to capture enough output in case something goes horribly wrong,  so you have an idea where to look. For this the error messages need to be informative enough. This is going to be a fun issue. I suggest we go wit `--build-info` `normal`, `brief`, `pony` first and then try to see how to do `none`. I always wonder if `pony` would be the error case though ;-)

snowleopard:
Yes, it's unclear at this point how to handle `none` properly.

By the way, I also added style `dot` in my comment to have everything in one place.

angerman:
I'll add this. To first-shake, as I have some of this already working in a branch.

snowleopard:
Surely this feature is nice to have, although I don't think it is critical for any milestone. However, if you feel you'll manage to complete this by `first-shake`, let's aim for this!

angerman:
Well, I got this one and the documentation issue assign to me so far. I sure hope, I'll manage :-)

snowleopard:
:+1:

angerman:
This is mostly implemented in #154. Even @bgamari unicorn!

`dot` and `none` are going to be a bit tricky though, as we currently send `"."` and `""` down.  Yet `dot` should not send a newline, and `none` should actually be a no-op instead of an empty string.

[...]

PR #154, opened 2016-01-12
Add Advanced render box styles

angerman:
Adds support for `--build-info` flags as proposed in #134. Also adds our all time favorite unicorn.

### DIFF
--- a/src/Oracles/Config/CmdLineFlag.hs
+++ b/src/Oracles/Config/CmdLineFlag.hs
@@ -1,33 +1,23 @@
-{-# LANGUAGE DeriveGeneric, DeriveDataTypeable, GeneralizedNewtypeDeriving #-}
-module Oracles.Config.CmdLineFlag (buildInfo, cmdLineOracle, flags, BuildInfoFlag(..)) where
+module Oracles.Config.CmdLineFlag (putOptions, buildInfo, flags, BuildInfoFlag(..)) where
 
-import GHC.Generics (Generic)
-
-import Development.Shake hiding (Normal)
-import Development.Shake.Classes
 import Data.Char (toLower)
 import System.Console.GetOpt
 
+import System.IO.Unsafe (unsafePerformIO)
+import Data.IORef
+
 -- Flags
 
-data BuildInfoFlag = Normal | Brief | Pony | Dot | None deriving (Eq, Show, Generic)
-
-instance Hashable BuildInfoFlag
-instance NFData BuildInfoFlag
-instance Binary BuildInfoFlag
+data BuildInfoFlag = Normal | Brief | Pony | Dot | None deriving (Eq, Show)
 
 data CmdLineOptions = CmdLineOptions {
     flagBuildInfo :: BuildInfoFlag
-} deriving (Eq, Show, Generic)
+} deriving (Eq, Show)
 
 defaultCmdLineOptions :: CmdLineOptions
 defaultCmdLineOptions = CmdLineOptions {
     flagBuildInfo = Normal
 }
-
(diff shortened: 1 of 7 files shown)


angerman:
@ndmitchell This doesn't work all the time, as I'm running into:
```
* inplace/bin/genprimopcode
* OracleQ (PackageDataKey (".build/stage0/utils/genprimopcode/package-data.mk",".build_stage0_utils_genprimopcode_C_SRCS"))
* .build/stage0/utils/genprimopcode/package-data.mk
* .build/stage0/utils/genprimopcode/package-data.mk utils/genprimopcode/stage0/package-data.mk
Build system error - cannot currently call askOracle:
  Reason:          Within withResource using Resource ghc-cabal
  Question type:   OracleQ CmdLineFlags
  Question value:  OracleQ (CmdLineFlags ())
Move the askOracle call earlier/later
```

could you lend me a hand here?

ndmitchell:
If you put the render box style in an oracle, then if someone decides to switch from `dot` to `unicorn` then suddenly its going to rebuild everything just so you can see lots of unicorns. I think the issue here is that this flag has gone in the oracle at all.

snowleopard:
Looks nice! Although I expected unicorns to be different each time... :D 
Once the oracle issue is resolved we'll merge this.

snowleopard:
@ndmitchell Agreed, a change in the output style shouldn't lead to rebuilds.

ndmitchell:
Best suggestion is just to shove the render box style in a global IORef with the typical `unsafePerformIO $ newIORef ...` code. With HEAD Shake you could shove it in shakeExtra, but I'd still opt for an IORef here.

angerman:
> Looks nice! Although I expected unicorns to be different each time... :D 

That's up to @bgamari :p I was only provided with one.

> @ndmitchell Agreed, a change in the output style shouldn't lead to rebuilds.

Right!

> Best suggestion is just to shove the render box style in a global IORef with the typical unsafePerformIO $ newIORef ... code. With HEAD Shake you could shove it in shakeExtra, but I'd still opt for an IORef here.

Wha! And I worked so hard to get the oracles right ;-)

src/Oracles/Config/CmdLineFlag.hs
```diff
@@ -0,0 +1,64 @@
...
+    deriving (Show, Typeable, Eq, Hashable, Binary, NFData)
+
+buildInfo :: Action BuildInfoFlag
+buildInfo = do
+    opts <- askOracle $ CmdLineFlags ()
+    return $ flagBuildInfo opts
+
+cmdLineOracle :: [CmdLineOptions -> CmdLineOptions] -> Rules ()
+cmdLineOracle opts = do
+    cache <- newCache $ \_ -> return $ foldl (flip id) defaultCmdLineOptions opts
```
ndmitchell: I don't think the cache helps here since the oracle itself is cached

ndmitchell:
The Oracle stuff looks good - I reviewed it and made a single minor remark for next time you have to go near oracles :)

angerman:
> The Oracle stuff looks good - I reviewed it and made a single minor remark for next time you have to go near oracles :)

Thanks. Much appreciated!

angerman:
Alright, now using an `IORef`. This feels dirty.  Once we add more command line options, we might want to distinguish between those that influence the build process and should end up in an Oracle, and those that should live in the `IORef`.

snowleopard:
@angerman Thanks, I'll merge this.

I don't like using `unsafePerformIO` so I hope we'll figure out a nicer solution eventually.

Issue #156, opened 2016-01-12
ffi build failes when /tmp is different device


kgardas:
Hello,
my attempt to use shake to build GHC on Solaris fails on building ffi with:
```
Reading shake-build/cfg/system.config...
| Successfully generated 'includes/ghcversion.h'.
| Successfully generated 'includes/ghcplatform.h'.
| Successfully generated 'includes/ghcautoconf.h'.
/------------------------------------------------\
| Run Tar (stage = Stage0, package = libffi)     |
|     input: libffi-tarballs/libffi-3.2.1.tar.gz |
| => output: /tmp/extra-dir--929703821           |
\------------------------------------------------/
/---------------------------------------------------\
| Move directory                                    |
|     input: /tmp/extra-dir--929703821/libffi-3.2.1 |
| => output: .build/stage0/libffi                   |
\---------------------------------------------------/
shakeArgsWith   0.002s    0%                           
Function shake  0.078s   31%  ===========              
Database read   0.001s    0%                           
With database   0.000s    0%                           
Running rules   0.165s   67%  ========================
Total           0.246s  100%                           
Error when running Shake build system:
* .build/stage1/rts/build/libHSrts.a
* OracleQ (PackageDataKey (".build/stage1/rts/package-data.mk",".build_stage1_rts_C_SRCS"))
* .build/stage1/rts/package-data.mk
* .build/stage1/rts/build/ffitarget.h
* .build/stage1/rts/build/ffi.h .build/stage1/rts/build/ffitarget.h
/tmp/extra-dir--929703821/libffi-3.2.1: rename: unsupported operation (Cross-device link)
```
the last message probably suggest that shake does not like to have /tmp mounted from different device than on home where it tries to build GHC...

snowleopard:
Thanks for reporting! We haven't tried building GHC on Solaris yet. The build system does rely on creating temporary files and directories.

kgardas:
And does it run for you on either Linux or Mac OS X when /tmp is tmpfs or different drive? That's the point IMHO here, otherwise Solaris is not that different...

snowleopard:
@kgardas I'm using Windows myself, so can't check.

@angerman Could you check on OS X?

kgardas:
Btw, when I moved the build tree to /tmp as a whole then this runs fine -- i.e. was able to rename unpacked ffi tree to stage0 sub dir. My guess is, the issue should be also preserved on Linux too...

ndmitchell:
[...]

PR #161, opened 2016-01-12
Cleanups in libffi

ndmitchell:
Following on from #156, I think a little more needs to be in the `actionFinally`, so that if things fail while extracting the tarball we still clean up properly.


ndmitchell:
Please note, untested, other than that it typechecks.

snowleopard:
:+1:

Issue #155, opened 2016-01-12
doc issue, missing shake install


kgardas:
Hello,

README is not clear about it, certainly it looks like you need to cabal install shake in order to give a shakized ghc a try. README mentions that what you needed with make for building GHC is enough, but this is not true probably.

kgardas:
Another try: it also looks like you also need QuickCheck installed.

kgardas:
And a last addition: you also need ansi-terminal

snowleopard:
Thanks, you are right, it's better to be explicit about dependencies. I've updated README.

angerman:
Haha we come full circle. Not sure how @quchen feels about that ;-)

Anyway. The QuickCheck Dependeny irks me a little. Shouldn't that be a `Test` only dependency? I though cabal allowed for Test only dependencies? cc: @ndmitchell

snowleopard:
@angerman It's the second time this issue is raised. We already discussed adding such a line here: https://github.com/snowleopard/shaking-up-ghc/issues/37#issuecomment-167225959. This appears to have never been actioned.

quchen:
`build.cabal.sh` installs all the deps automatically, why ping me?

angerman:
Sorry @quchen, will not happen again.

ndmitchell:
@angerman you can have multiple exe's in Cabal, and have different dependencies for each. We have one single exe though (which seems to suit us best), so all dependencies are required for everything.

PR #162, opened 2016-01-13
#155, fix the name of the QuickCheck package

ndmitchell:
It's package name is camel case rather than lowercase (unlike most packages)


snowleopard:
:+1:

PR #166, opened 2016-01-13
Ghc cabal

angerman:
This should bring us towards fixing #18! :tada:


snowleopard:
I've removed `ghc-cabal` package (woo-hoo!) and added support for `hp2ps`. Everything looks very nice compared to the previous stringy mess. Let's see how far I can go today!

snowleopard:
This is taking longer than expected, so I am thinking that we might have to postpone it until a later milestone. @angerman Do you agree?

angerman:
@snowleopard Yes. I think we tried to byte of a bit too much to chew.

angerman:
I don't think this is going anywhere. Let's shave this yak some other time.

snowleopard:
@angerman Yes, this turned out to be quite a bit more complicated than we hoped to. Anyway, thank you and @bgamari for the efforts!

snowleopard:
I think we should also delete the `ghc-cabal` branch. It's badly out of date and is likely to confuse people.

Issue #159, opened 2016-01-12
Build ignores --with-* configure params


kgardas:
It looks like shake completely ignores the fact that I've configured GHC to use system supplied GMP library (with --with-gmp-includes/--with-gmp-libraries) and tries to build its own GMP library build for GHC.
This is also the case of --with-ffi-* params since even if I use it shake still insist on building its own ffi library. I guess very same is also with iconv and ncurses related parameters.

snowleopard:
We do have some functionality to detect installed `gmp`, but it hasn't been tested properly. And at the moment we always build `ffi`. Feel free to pick up this issue if you like.

Did the building of in-tree `gmp` and `ffi` work for you?

kgardas:
FFI build seems to be fine after moving to /tmp to work around #156 the problem is gmp build which fails with
```
Error when running Shake build system:
* utils/mkUserGuidePart/stage1/build/tmp/mkUserGuidePart
* OracleQ (PackageDataKey (".build/stage1/utils/mkUserGuidePart/package-data.mk",".build_stage1_utils_mkUserGuidePart_C_SRCS"))
* .build/stage1/utils/mkUserGuidePart/package-data.mk
* .build/stage1/utils/mkUserGuidePart/package-data.mk utils/mkUserGuidePart/stage1/package-data.mk
* .build/stage1/compiler/package-data.mk
* .build/stage1/compiler/package-data.mk compiler/stage1/package-data.mk
* .build/stage0/gmp/include/ghc-gmp.h
user error (Development.Shake.cmd, system command failed
Command: bash configure
Current directory: /tmp/shake-test/libraries/integer-gmp
Exit code: 1
Stderr:
configure: error: Cannot find gmp.h
)
```

snowleopard:
It seems to me that it fails right at the stage of determining whether there is an installed `gmp`. This is done by running `configure` in `libraries/integer-gmp`, which produces file `config.mk` with `HaveFrameworkGMP = YES`.

Could you show previously executed commands?

Second question: how do you typically run `configure` to build GHC using `make` and use installed `gmp`?

snowleopard:
To whoever would like to pick this up: the right files to change are: `Rules/Gmp.hs` and `Rules/Libffi.hs`.

I won't be able to tackle this in the coming days.

kgardas:
[...]

PR #168, opened 2016-01-13
fix handling of --with-gmp-* configure arguments

kgardas:
This fixes part of issue #159.

### DIFF
--- a/src/Rules/Gmp.hs
+++ b/src/Rules/Gmp.hs
@@ -57,6 +57,19 @@ configureArguments = do
            , "--host=" ++ hostPlatform
            , "--build=" ++ buildPlatform]
 
+configureIntGmpArguments :: Action [String]
+configureIntGmpArguments = do
+    includes      <- settingList GmpIncludeDirs
+    libs          <- settingList GmpLibDirs
+    return ([]
+            ++ (if (not (null includes))
+                then map ((++) "--with-gmp-includes=") includes
+                else [])
+            ++ (if (not (null libs))
+                then map ((++) "--with-gmp-libraries=") libs
+                else [])
+           )
+
 -- TODO: we rebuild gmp every time.
 gmpRules :: Rules ()
 gmpRules = do
@@ -105,7 +118,8 @@ gmpRules = do
         runConfigure libPath envs args
 
         -- TODO: currently we configure integerGmp package twice -- optimise
-        runConfigure (pkgPath integerGmp) [] []
+        intGmpArgs <- configureIntGmpArguments
+        runConfigure (pkgPath integerGmp) envs intGmpArgs
 
         createDirectory $ takeDirectory gmpLibraryH
         -- check whether we need to build in tree gmp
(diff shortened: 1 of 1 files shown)


snowleopard:
Many thanks @kgardas!

I have a question about this piece of code:
```haskell
configMk <- liftIO . readFile $ gmpBase -/- "config.mk"
if "HaveFrameworkGMP = YES" `isInfixOf` configMk
then do
    putBuild "| GMP framework detected and will be used"
    copyFile gmpLibraryFakeH gmpLibraryH
else if "HaveLibGmp = YES" `isInfixOf` configMk
     then do
       putBuild "| GMP detected and will be used"
       copyFile gmpLibraryFakeH gmpLibraryH
     else do [...]
```
It looks like the first two checks can be combined into one:
```haskell
configMk <- liftIO . readFile $ gmpBase -/- "config.mk"
if "HaveFrameworkGMP = YES" `isInfixOf` configMk || "HaveLibGmp = YES" `isInfixOf` configMk
then do
    putBuild "| GMP framework detected and will be used"
    copyFile gmpLibraryFakeH gmpLibraryH
else do [...]
```
Do you know if these two checks should indeed be handled in the same way? If yes, perhaps you can merge them as above to simplify the code.

Which of the two checks is activated on your machine? I presume the one with `HaveLibGmp`.

kgardas:
@snowleopard yes, checks can be combined, but then for initial testing I kept this in a way to see the difference in detection of frameworks-based or plain GMP. Honestly "framework GMP" is IIRC Mac OS X specific thing.
And yes, you are right on Solaris and everywhere else except probably Mac OS X, HaveLibGmp is YES when the GMP library is detected.
OK! Will merge the checks now...

snowleopard:
> I kept this in a way to see the difference in detection of frameworks-based or plain GMP

Another question (sorry, I have very little understanding of GMP): what is the difference between "plain" and "framework-based" GMP?

ndmitchell:
My rough understanding is that Mac is too cool to use libraries like everyone else, so they use frameworks, which I've always viewed as an alternative to a set of dll/so but with more metadata. See http://stackoverflow.com/questions/4618027/what-exactly-is-a-framework-in-mac-os-x-framework-folders.

kgardas:
@snowleopard don't know any information about framework GMP except that the term is used in GHC and the check is done in a different way, see the code in libraries/integer-gmp/aclocal.m4:
```
AC_DEFUN([LOOK_FOR_GMP_FRAMEWORK],[
    if test "$HaveLibGmp" = "NO"
    then
        case $target_os in
        darwin*)
            AC_MSG_CHECKING([for GMP.framework])
            save_libs="$LIBS"
            LIBS="-framework GMP"
            AC_TRY_LINK_FUNC(__gmpz_powm_sec,
                             [HaveFrameworkGMP=YES; GMP_FRAMEWORK=GMP])
            LIBS="$save_libs"
            AC_MSG_RESULT([$HaveFrameworkGMP])
            ;;
        esac
    fi
```
and in configure.ac in the same dir:
```
AC_ARG_WITH([gmp-framework-preferred],
  [AC_HELP_STRING([--with-gmp-framework-preferred],
    [on OSX, prefer the GMP framework to the gmp lib])],
    [GMP_PREFER_FRAMEWORK=YES],
    [GMP_PREFER_FRAMEWORK=NO])
```
looks like pure Mac-ism...

kgardas:
@ggreif silly me, indeed map is working also on empty lists. @snowleopard I'll provide simplified version based on @ggreif code...

snowleopard:
Many thanks all! Looking forward to the last revision, ready to merge.

snowleopard:
Let me hold on with the merge while I'm checking how to fix the Travis regression which happens when GMP is already installed. I wonder why @kgardas hasn't experienced the same error...

kgardas:
@snoleopard the build fails when gmp is installed/detected and fails with:
```
* utils/compare_sizes/stage1/build/tmp/compareSizes
* .build/stage1/libraries/integer-gmp/build/libHSinteger-gmp-1.0.0.1.a
* getDirectoryFiles  [.build/stage0/gmp/objs/*.o]
.build/stage0/gmp/objs: getDirectoryContents: does not exist (No such file or directory)
```
that probably means that some from the infrastructure still counts with the object file from in tree built GMP and would like to do something with it.

ndmitchell:
So seems @kgardas is hit by this bug. It's a Shake bug :( Workaround is to `doesDirectoryExist` first, which is deeply sad, as I previously told @snowleopard it definitely wasn't required (and thus my beer count increases further).

snowleopard:
Let me merge this and temporary fix the bug with an extra `doesDirectoryExist` check.

snowleopard:
With the above fix we no longer have the `getDirectoryContents` error. However, on my machine I get a linking error when using installed GMP. Will investigate now.

snowleopard:
I found the following comment in the old build system:
```
# Apparently building on Windows fails when there is a system gmp
# available, so we never try to use the system gmp on Windows
```
I think we can mirror the old build system here. Maybe we will come back to fixing system GMP on Windows in future.

snowleopard:
Done.

PR #171, opened 2016-01-14
advanced render box

angerman:
The addition of `dropEscSequence` is rather unfortunate.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -30,6 +30,7 @@ install:
     - bash -lc "cd /home/ghc; mkdir -p ghc-tarballs/mingw-w64/x86_64/"
     - bash -lc "cd /home/ghc; mkdir -p ghc-tarballs/perl/"
     - bash -lc "cd /home/ghc; exec 0</dev/null; ./configure --enable-tarballs-autodownload"
+    - bash -lc "cat /home/ghc/shake-build/cfg/system.config"
     - ghc   --version
     - stack --version
     - alex  --version
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,6 +59,7 @@ install:
 
     - ( cd ghc && ./boot )
     - ( cd ghc && ./configure )
+    - cat ghc/shake-build/cfg/system.config
 
 script:
     - ( cd ghc/shake-build && cabal haddock --internal )
--- a/cfg/system.config.in
+++ b/cfg/system.config.in
@@ -1,5 +1,6 @@
-# Edit 'user.config' to override these settings.
-#===============================================
+# This file is processed by the configure script
+# See 'src/Settings/User.hs' for user-defined settings
+#=====================================================
 
 # Paths to builders:
 #===================
@@ -39,6 +40,7 @@ nm               = @NmCmd@
 objdump          = @ObjdumpCmd@
 ranlib           = @REAL_RANLIB_CMD@
 tar              = @TarCmd@
+patch            = @PatchCmd@
 perl             = @PerlCmd@
 
 # Information about builders:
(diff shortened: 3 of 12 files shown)


snowleopard:
This feels very wrong:
* Why do we drop `none` but keep `dot`? I think `none` is much more useful than `dot`, especially taking into account Shake's built-in progress reporting feature. Dots are very primitive indicators of progress: one dot might take minutes when running something expensive like `configure`. Why don't we rely on Shake to report progress instead? I don't remember who proposed printing dots initially: what was the rationale? Is this feature still wanted?
* We don't want to pollute `Main.hs` with low-level functions like `dropEscSequence`.
* I don't understand why all the complexity. I imagined we will simply have a function `putBuildInfo` which respects the `--build-info` flag:
```haskell
putBuildInfo msg = do
    style <- [...]
    when style /= None $ putBuild msg
```
All print statements that want to respect the `--build-info` flag should switch to using `putBuildInfo` instead of lower-level `putBuild`. (`putSuccess` may become something like `putBuildInfo Green`.)

angerman:
The way the adjusted code works is that `none` would essentially be `-q` for shake. I guess we kind of talked about something different here. Your suggesting sounds more sensible though. I'll adjust the PR accordingly.

snowleopard:
@angerman Thanks! I think we need to clarify our understanding of what exactly is controlled by `--build-info` flag. One simple way to define it is: `--build-info` controls the output of `putBuildInfo`. I think this is clear and precise.

Why doesn't `--build-info` control all output? Probably because there may be other classes of messages that don't fall into the build info category. For example, we might want to print a warning, or a rough progress prediction, e.g.: "Stage1 compiler must be rebuilt, this will take awhile.".

Maybe the name `--build-info` is not the best choice after all, because technically everything printed by a build system is build info. Shall we come up with a better name?

angerman:
@snowleopard how about `--action-style=` with `None | Brief | Box | Unicorn`, where `Box` is default?
We usually use the box to print information about performed actions, right?

snowleopard:
@angerman In most cases yes, but sometimes we do `putSuccess`. Also there are messages like:
```haskell
putBuild "| GMP framework detected and will be used"
```
I think they all are related to _build progress_, so maybe `--progress-info` would be better? Note they are not always 'boxes', so I think `None | Brief | Normal | Unicorn` is more precise.

angerman:
@snowleopard alright, let me know what you think :)

snowleopard:
@angerman Thanks, that's much cleaner! :+1: 

There are some remaining clean-ups, which I'm fine doing myself if you can't stand this PR any more :) 
* Given the new name of the flag, we should rename `putBuildInfo` to `putProgressInfo`.
* Similarly, shall we rename `renderPony` to `renderUnicorn`?
* You removed spaces after `input:` and `output:` in `renderAction` (again), please bring them back.
* Shall we move `putProgressInfo` to `Rules/Actions.hs`? Arguably it should only be called by build rules, so placing `putProgressInfo` closer to call sites seems reasonable.

angerman:
@snowleopard I'll see this through :) Ahh that space thing. That probably happened during the inclusion of the master branch. And I failed to catch that in the merge.

Those requests sound reasonable. Will get at them tomorrow.

angerman:
@snowleopard ok. Just figured we have a scheduled blackout tomorrow from 8am to 2pm for a transformer replacement. So I went ahead to fixed your raised concerns.

snowleopard:
@angerman :+1: Awesome, many thanks!

angerman:
Well well, I still screwed up :weary: I'll do my very best to prevent this from happening again :pray:

snowleopard:
I don't see where you screwed up! Thanks to you we now have a nice `--progress-info` feature. Just bumped at a couple of corners on the way due to an incomplete initial specification.

PR #175, opened 2016-01-16
Angerman/feature/doc

angerman:
Let's _not_ merge this now. But keep this PR as a basis for discussion on the high level documentation.  My intention with this is to make navigation of the code base to people new to this project (and potentially shake) easier and explain how the parts are supposed to play together.


snowleopard:
Thanks @angerman! I've been watching this for some time now and it's a very good start.

I think it will be easier to find in `doc` folder though. We may also think about a more informative name than `README`, to hint that this is a high-level overview.

Maybe `doc/overview.md`?

snowleopard:
I hope later we can add more focused docs, such as `doc/oracles.md`, `doc/expressions.md`, etc.

angerman:
I'd like to keep them as `README.md` in the folders primarily, because that way github shows their content rendered when browsing the source on github. Also I'd like to keep the source structure defined by the folders to mean something :)

It might be a good idea to replace them with symlinks and have the actual documentation in the `doc` folder. My hope is that by being very visible, it ends up being corrected more likely once someone sees a bug.

I certainly agree, that oracles and expressions should get their own subsection. I was planning on getting at the `README.md`s in Oracles and then in Settings, as I hoped that Rules are pretty self explanatory.  
Not sure where to put Expressions though.

Let me know what you think!

snowleopard:
> I'd like to keep them as README.md in the folders primarily, because that way github shows their content rendered when browsing the source on github. Also I'd like to keep the source structure defined by the folders to mean something :)

Aha! I see, this makes sense, although feels unusual.

> It might be a good idea to replace them with symlinks and have the actual documentation in the doc folder. 

I really want to have all docs available in one place. People usually expect them in `doc` folder, other locations may be surprising, and surprises are typically bad in design :-) Does github support symlinks? If yes, this may be a good compromise.

> I was planning on getting at the README.mds in Oracles and then in Settings

This feels fragile to me and adds weight for maintenance. Whenever things are coupled in this way, it is almost inevitable that some time in future they will desynchronise when someone forgets to update one after updating the other. What if we decide to refactor codebase into a different, better directory structure? 

> Not sure where to put Expressions though.

I'm afraid Expressions is not the only problem. There are other important concepts that are worth documenting, which are spread over the codebase and figuring out a proper doc location for them would be difficult.

Having said that, I see value in having `README` files in source directories, because they are shown automatically by github and hence may indeed help someone new to make sense of the codebase. We could provide brief descriptions of the local folder structure and link to appropriate documentation sections in the `doc` directory. This will also be fragile, but will be easier to keep in sync: we'll keep these `README`s very brief and link to stable documentation, which preferable should be as decoupled from folder structure as possible. For the documentation which _must be tightly coupled with code_ we should use Haddock, I think.

Then there is also github wiki...

Let's keep discussing.

angerman:
@snowleopard Thanks for the feedback, keep it coming!

I know it might be unusual, but I've wanted this for many projects many times already. Being somehow taken by the hand and guided through the structure and the project. Say a double decker city tour :grinning: 

I guess the fragility depends a lot on the details put into those docs. I'm actually aiming for a very high level documentation without going too much into detail.  If I went too much into detail somewhere please highlight it, so I can drop some details.

I basically see three layers of documentation:
- High level guidance about the projects structure and main ideas. (`README.md`(s))
- Detailed discussion of the projects goals and inner mechanics. (Read the docs, Man page, ...)
- Code level documentation, explaining the intent behind each function. (Haddocks)

As I've said elsewhere I'm not a big fan of github wiki pages. They feel like an after thought and just tacked onto the projects. They are also not part of source tree, which I think the documentation should be; even if it's the source for some documentation generation. E.g. cloning the source should give me the documentation as well, even though I might have to use tool `T` to produce the documentation from the information contained within the source repository. (E.g. for the code level documentation, that would be generation of the haddocks).

Hence my initial plan of attack was to build the very high level guidance through the source code, and then see where and how to add a deeper layer of detailed documentation, which I guess can perfectly live in `/doc`; I've hence tried to see if we can just symlink the `README.md` experimentally into the `doc` folder, but github doesn't play really nice here. As per explanation above, symlinking might actually not be a good solution.

If you consider the current README's good enough for `/doc`, we could move them there under different names; if my preference of one README per folder is considered harmful rather than helpful, I'm willing to rest my case on this; I see this only succeeding if everyone considers this useful, and hence will report inaccuracies.

snowleopard:
> I know it might be unusual, but I've wanted this for many projects many times already. Being somehow taken by the hand and guided through the structure and the project. 

This does sound good!

> I basically see three layers of documentation:
- High level guidance about the projects structure and main ideas. (`README.md`(s))
- Detailed discussion of the projects goals and inner mechanics. (Read the docs, Man page, ...)
- Code level documentation, explaining the intent behind each function. (Haddocks)

I like these three layers. Let's give it a try! Please keep `README`s as brief and high-level as possible, linking to appropriate sections of `doc` files for any further details.

I think it is also important that `doc` files are independent of `README`s. That is: one can get all the information about the build system by looking only into `doc` folder, without the need to search for `README`s scattered in the source directories.

@angerman To sum up, let's go ahead with your approach!

ndmitchell:
Just my opinion, but if someone wants to know stuff, they want to know stuff, they don't usually have the context to know what they want to know. As a result, usually a single file with sections is best - if the build system is complex enough to require lots of files, its probably too complex. Having a README (for people who want to use) vs a Dev document (for people who want to hack) might be a good idea.

angerman:
@ndmitchell fair points. I might be overshooting with some of this. But would like to see how the one (rather short) README per folder plays out. It is in no case supposed to substitute a more in depth development documentation, but aid in superficially browsing the source. I consider myself quite a novice, and often am lost in developer documentation as I'm missing some prior knowledge. While I suspect someone more knowledgable would find a superficial overview either insulting, or not helpful.

I feel we have the *user* REAMDE.md already at the top level.  A proper developer documentation augmenting the haddocks will follow.  I'm trying to target somewhere in between the *user* README and the developer documentation for now. My target audience are those, who know a little about haskell, and would like to understand the big picture about how ghc is built and give enough guidance to be able to make some sense of the code that is used to produce the final product. I do not intend to make someone proficient in hacking this project. That's what the developer documentation will be for.

As others have mentioned already, heavy refactoring can invalidate the developer documentation pretty fast. I hope to achieve some high level overview, that would be immune to most of those refactoring, but still provide some high level insight.

Taking a text adventure as an example, I'd like to provide the setting. You can move around, and I'm able to tell you what the scene looks like. But you are deprived of any `Look` command that would allow you to focus on details; you'd have to install the developer extension for that.

angerman:
And yes. Please! If I'm going overboard and too much into detail. Tell me, so we can move that into the developer docs.

snowleopard:
@angerman That last bit with library descriptions (2f05d11) is actually very useful for me :) I have little knowledge of what most of the libraries/packages are for, and having all packages briefly described in one place would be great.

angerman:
@snowleopard ha! It's not just me! :grinning: I'll keep grinding at this branch and will hopefully pick up some speed through this week.

angerman:
This has been dead for almost two years now. Let's just nuke it.

snowleopard:
@angerman Yes, it's got a bit outdated already, but there are still some useful bits. Let's keep your branch, maybe we'll recover some text from it someday!

PR #182, opened 2016-01-17
fix handling of FFI library configure params

kgardas:


snowleopard:
Thanks @kgardas!

Several comments:
* I think `UseSystemFfi` should live in `Oracles.Config.Flag` with other YES/NO flags.
* Configure variables have singular names `@FFIIncludeDir@` and `@FFILibDir@` yet you are using plural labels `ffi-include-dirs` and `ffi-lib-dirs`, and store them as multi-valued settings, which later forces you to use `head`. I suggest to give them type `Setting` instead of `SettingList`.
* What is the purpose of `FfiLibDirs`? At the moment it is not used at all.

kgardas:
@snowleopard setting -> flag is done. With dirs I've followed established practice of system.config.in for gmp and iconv libraries:
```
iconv-include-dirs = @ICONV_INCLUDE_DIRS@
iconv-lib-dirs     = @ICONV_LIB_DIRS@

gmp-include-dirs   = @GMP_INCLUDE_DIRS@
gmp-lib-dirs       = @GMP_LIB_DIRS@
```
anyway, the configure itself seems to talk about singular (dir):
```
$ ./configure --help|grep gmp
  --with-gmp-includes     directory containing gmp.h
  --with-gmp-libraries    directory containing gmp library
karel@silence:/tmp/shake-test$ ./configure --help|grep iconv
  --with-iconv-includes   directory containing iconv.h
  --with-iconv-libraries  directory containing iconv library
karel@silence:/tmp/shake-test$ ./configure --help|grep ffi
  --with-system-libffi    Use system provided libffi for RTS [default=no]
  --with-ffi-includes=ARG Find includes for libffi in ARG [default=system
  --with-ffi-libraries=ARG
                          Find libffi in ARG [default=system default]
```
so perhaps we can also later change gmp/iconv variables? (in another branch/pull/commit of course!)

snowleopard:
@kgardas If each of these settings is supposed to be a single directory only then I suggest to switch to `Setting` with all of them (and rename all labels accordingly). Yes, I'd appreciate a separate PR for gmp/iconv variables.

kgardas:
@snowleopard dirs to dir change is done. Still remains:
- pass dir values to HsCpp
- find out where linker passes -L<FFILibDir> -- and set that too correctly

snowleopard:
Thanks @kgardas. Just to help you find relevant files in the codebase:
* `Settings.Builders.HsCpp` is used to pass flags to `HsCpp`.
* I think linker options are set in `Settings.Builders.Ghc`.

kgardas:
@snowleopard I'm done with this one, please review and if ok merge. Thanks!

snowleopard:
@kgardas Everything looks fine. Just one comment: you moved
```haskell
when trackBuildSystem $ need [sourcePath -/- "Rules/Libffi.hs"]
```
into an else branch, which is wrong I think. The purpose of this statement is to force `libffiRules` to rerun whenever anyone changes this particular source file. Otherwise it is difficult to ensure that the build system is in a consistent state. So, this line should always be executed, and the best placement for it is in the very beginning of the rule.

kgardas:
@snowleopard I'll fix that immediately and thanks for the review.

snowleopard:
Many thanks @kgardas! This is now merged.

PR #183, opened 2016-01-19
fix "dirs" to "dir" for gmp and iconv libraries

kgardas:
@snowleopard this is requested dirs to dir pull request. I'm trying to make it as simple as possible and so putting dir into list whenever applicable to keep code intact. Tested on Solaris 11 with system supplied gmp and ffi libraries.


snowleopard:
Thank you. Looks fine!

PR #185, opened 2016-01-19
do not pass --with-intree-gmp to configure when system gmp is used

kgardas:
@snowleopard this PR fixes issue of ghc-cabal configuring integer-gmp with intree gmp even when user prefers to use system supplied gmp.

### DIFF
--- a/src/Settings/Packages/IntegerGmp.hs
+++ b/src/Settings/Packages/IntegerGmp.hs
@@ -4,7 +4,6 @@ import Base
 import Expression
 import GHC (integerGmp)
 import Predicates (builder, builderGcc, package)
-import Settings.Builders.Common
 import Settings.Paths
 import Oracles.Config.Setting
 
--- a/src/Settings/Packages/IntegerGmp.hs
+++ b/src/Settings/Packages/IntegerGmp.hs
@@ -15,17 +15,13 @@ import Oracles.Config.Setting
 integerGmpPackageArgs :: Args
 integerGmpPackageArgs = package integerGmp ? do
     let includeGmp = "-I" ++ gmpBuildPath -/- "include"
-    gmp_includedir <- getSetting GmpIncludeDir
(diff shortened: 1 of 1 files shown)


snowleopard:
@kgardas :+1: 

Some comments:
* Let's stick to `lowerCamelCase` naming convention as in the rest of the build system: rename the variables to `gmpIncludeDir` and `gmpLibDir`.
* import Settings.Builders.Common seems to be redundant?
* I think this can be simplified as follows using operator `?`:
```haskell
    let includeGmp = "-I" ++ gmpBuildPath -/- "include"
    gmpIncludeDir <- getSetting GmpIncludeDir
    gmpLibDir     <- getSetting GmpLibDir
    mconcat [ builder GhcCabal ? mconcat
              [ (null gmpIncludeDir && null gmpLibDir) ?
                arg "--configure-option=--with-intree-gmp"
              , appendSub "--configure-option=CFLAGS" [includeGmp]
              , appendSub "--gcc-options"             [includeGmp] ]

            , builderGcc ? arg includeGmp ]
```

kgardas:
@snowleopard Interesting, if ? does what it looks so, then indeed. W.r.t names, will do. W.r.t. import, I'll verify this. Thanks for the review!

snowleopard:
@kgardas Here is how `?` works:
```haskell
true  ? args == args
false ? args == mempty
```
where `true` and `false` can be of type `Bool`, `Action Bool` or `Predicate`. Note, `Args` is a Monoid, with `mempty` being equivalent to a no-op. Furthermore, `arg` appends a single argument to `Args`, `append` appends a list of arguments to `Args`, `remove` removes arguments from `Args` if they are present, etc.

kgardas:
@snowleopard thanks for explanation, it looks natural, but underlying tech is still over my head. Anyway, code should be fixed based on your review. Retested on Solaris.

snowleopard:
Many thanks. It also passed all CI checks. Merged now.

PR #190, opened 2016-01-22
Allow building ghc-shake with stack

joehillen:
I've found stack to be more reliable than cabal sandboxes.

This PR adds `build.stack.sh` based on `build.cabal.sh`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ cfg/system.config
 cabal.sandbox.config
 dist/
 .cabal-sandbox/
+.stack-work/
--- a/build.cabal.sh
+++ b/build.cabal.sh
@@ -28,8 +28,8 @@ function rl {
     echo "$RESULT"
 }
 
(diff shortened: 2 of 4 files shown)


snowleopard:
Thanks @joehillen! I spotted a typo which seems to have come from `build.cabal.sh`: `absoltueRoot`. Could you fix this in both scripts?

I can't test this myself, but I hope @ndmitchell could test it when he is back as he is also keen on using Stack. It would be nice to have `.bat` equivalents for all `.sh` scripts too, but this could be done separately.

ndmitchell:
Not sure I'll be able to test much as I'm on Windows and using the .bat, but it all looks plausible. I have been using stack as the installed environment, but global cabal packages to build this project.

joehillen:
@snowleopard I fixed the typos

@ndmitchell It should work in cygwin (or whatever people use nowadays).

snowleopard:
@joehillen Thanks again! This is now merged.

PR #192, opened 2016-01-24
Make .cabal meta-data more accurate

hvr:
We need this so cabal (this is even more important with the upcoming
nix-style cabal features) can do a proper job so this is a pre-requisite
for the new build-system being used by default for GHC anyway, as we
need to be as accurate as possible with the build specification to give
`git bisect` a chance of remaining usable.

### DIFF
--- a/shaking-up-ghc.cabal
+++ b/shaking-up-ghc.cabal
@@ -107,26 +107,24 @@ executable ghc-shake
                        , Test
                        , Way
 
-    default-extensions:  BangPatterns
-                       , LambdaCase
-                       , MultiWayIf
-                       , TupleSections
+    default-language:    Haskell2010
     other-extensions:    DeriveDataTypeable
                        , DeriveGeneric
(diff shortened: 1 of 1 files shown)


snowleopard:
Thanks @hvr! One question: do you think it makes sense to add `alex` and `happy` into dependencies? Obviously they are not direct dependencies, but the build system is useless without them.

hvr:
@snowleopard well, you *could* add dependencies to `alex` and `happy` as `buildtools` with proper version bounds (rather than as `build-depends` which would be wrong and may not even work with future cabal versions). However, this (i.e. automatic installation of build-tools) is one area that's still in flux in cabal. Adding as `build-tools` won't hurt however.

snowleopard:
Thanks @hvr! It seems less straightforward than I thought, so let's postpone adding `alex` and `happy` to the cabal file.

This is now merged.

PR #199, opened 2016-02-01
Clear PWD

### LINKED ISSUE #189: configure fails
I checked out a fresh GHC and tried to build from scratch using `--configure=--enable-tarballs-autodownload`. It failed with.

    configure: error: in `C:/Neil/ghc':
    configure: error: C compiler cannot create executables
    See `config.log' for more details

I then ran `./configure --enable-tarballs-autodownload` by hand and it said:

    configure: Checking for Windows toolchain tarballs...
    checking for gcc... C:/Neil/ghc/inplace/mingw/bin/gcc.exe
    checking whether the C compiler works... no
    configure: error: in `/c/Neil/ghc':
    configure: error: C compiler cannot create executables
    See `config.log' for more details

In neither case did `config.log` say anything interesting. I cleaned everything and reran `boot`/`configure` by hand and it worked. I then ran `shake-build/build.bat -j` and it said:

    Configuration file shake-build/cfg/system.config is missing.
    Run the configure script either manually or via the build system by passing --configure[=ARGS] flag.`

That seems to be no longer true, as I've already run `configure`. I then passed `--configure=--enable-tarballs-autodownload` and it now seems to be working (at least it's got a lot further). 32bit Windows.

ndmitchell:
Your guess was spot on, if you have PWD set then `pwd` uses that as the current path. If not, it uses the current path, and translates it into the appropriate Unix style in MinGW, fixing configure.

Fixes #189

### DIFF
--- a/src/Environment.hs
+++ b/src/Environment.hs
@@ -11,6 +11,11 @@ setupEnvironment = do
     -- ghc-cabal refuses to work when GHC_PACKAGE_PATH is set (e.g. by Stack)
     unsetEnv "GHC_PACKAGE_PATH"
 
+    -- in MinGW if PWD is set to a Windows "C:\\" style path then configure
+    -- `pwd` will return the Windows path, and then modifying $PATH will fail.
+    -- See https://github.com/snowleopard/shaking-up-ghc/issues/189 for details.
+    unsetEnv "PWD"
+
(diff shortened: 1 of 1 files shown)


snowleopard:
Awesome!

PR #202, opened 2016-02-03
Update stack.yaml to lts-5.1

joehillen:
I don't plan to change this file every time there is a new stackage lts, but lts-4.x has a bug with aeson.


snowleopard:
Thanks! As far as I am aware we don't have many users of your stack script yet, so feel free to change it the way it works best for you.

PR #208, opened 2016-02-17
Ensure that if ghc compilation fails then we return a non-zero error code from build.bat

ndmitchell:
At the moment if compiling the build script fails the build.bat returns 0, it should return 1 to indicate failure. I checked it still returns 0 on success and 1 on compilation failure.

### DIFF
--- a/build.bat
+++ b/build.bat
@@ -21,8 +21,8 @@
 
 @ghc %ghcArgs%
 
-@if %ERRORLEVEL% EQU 0 (
-    @rem Unset GHC_PACKAGE_PATH variable, as otherwise ghc-cabal complains
-    @set GHC_PACKAGE_PATH=
-    @.shake\build %shakeArgs%
-)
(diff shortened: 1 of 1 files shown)


snowleopard:
Thanks! Not sure what happened with Travis though... but for some reason `build.sh` appears broken?

ndmitchell:
No idea - this patch only touches build.bat. There is a merge commit in there, but the merge doesn't touch build.sh either. Perhaps take the diff from this patch and apply it by hand?

ndmitchell:
And also perhaps rerun the last Travis state? See if the broken sh is somewhere weird, like a Travis version upgrade (I've been bit by those at various points).

snowleopard:
So, it seems to have gone through on AppVeyor, so let me merge this. I'll fix the Travis issue later.

snowleopard:
So, the Travis issue got fixed by itself. So I guess it was Travis fault.

PR #212, opened 2016-03-03
Add docs for how to compile on Windows

ndmitchell:
As per https://github.com/snowleopard/shaking-up-ghc/issues/110#issuecomment-191954062


snowleopard:
@ndmitchell Great, thanks! It would be nice to mirror your build script in AppVeyor to continuously test it.

ndmitchell:
Yep, mirroring in AppVeyor sounds good - although that starts from MinGW already on the path so is a bit different.

PR #230, opened 2016-04-22
Add shake-0.15.6 to stack extra-deps

cocreature:
Since you bumped the lower bound, this is necessary for `stack build` to work.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -8,7 +8,8 @@ packages:
 - '.'
 
 # Packages to be pulled from upstream that are not in the resolver (e.g., acme-missiles-0.3)
-extra-deps: []
+extra-deps:
+- shake-0.15.6
 
 # Override default flag values for local packages and extra-deps
 flags: {}


ndmitchell:
This should become unnecessary as soon as the Stackage LTS rolls around? Do we know how soon that will be?

snowleopard:
I don't mind to merge this but as @ndmitchell points out this is likely to be a temporary change and will soon have to be undone. If Stackage LTS updates quickly enough it may be better to just wait a little.

ndmitchell:
Or it may be better for this project to delay upgrading Shake for the week or so necessary to hit Stackage LTS in future? Not really sure.

cocreature:
I think you should either delay shake upgrades until the reach LTS, use extra-deps or drop stack support. Having a supported stack configuration that sometimes doesn’t work is pretty bad imho.

snowleopard:
@ndmitchell I wanted to do a couple of fixes I had in mind that require the new Shake, so I'd prefer to switch asap.

@cocreature OK, I'll merge this. It's not such a big deal to undo this later.

ndmitchell:
I agree that having Stack broken for some time period is not desirable.

Perhaps the Cabal constraints should be adjusted to >= 0.15.6, so Cabal will pick the latest and Stack will pick the last LTS. In most cases you can then upgrade, and it's only when you want the new features you have to bump the Cabal constraint.

snowleopard:
@ndmitchell Good idea! Done in the above commit.

PR #256, opened 2016-05-22
Set `THREADS` when running validate

michalt:
GHC testsuite uses the `THREADS` env variable (and not the make's `-j`
setting) to control the parallelism. This commit sets THREADS to the
value of `shakeThreads`.

### DIFF
--- a/src/Settings/Builders/Make.hs
+++ b/src/Settings/Builders/Make.hs
@@ -11,4 +11,4 @@ makeBuilderArgs = do
     mconcat
         [ builder (Make gmpBuildPath     ) ? append ["MAKEFLAGS=" ++ j]
         , builder (Make libffiBuildPath  ) ? append ["MAKEFLAGS=" ++ j, "install"]
-        , builder (Make "testsuite/tests") ? arg "fast" ]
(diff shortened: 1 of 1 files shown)


snowleopard:
@michalt Thanks! Now merged.

By the way, are you just playing with Hadrian or actually using it during GHC development? How is your experience so far?

michalt:
@snowleopard I'm not really a regular GHC contributor, so you could say that I'm just playing with both GHC and Hadrian ;-) So far Hadrian works pretty well (although, I've used it mostly for pretty basic stuff, e.g., I've tried `validate` only yesterday and noticed the problem with threads). Compared to the old `make`-based system, it's pretty nice that I can actually have a vague idea of how things work and try to solve a problem (instead of "`make distclean` and try again"). Also, it feels that Hadrian re-compiles *far less* code on small code changes. So my plan is to try to avoid the old system and use Hadrian instead. :-)

snowleopard:
Thanks @michalt! Great to hear that Hadrian is already usable for basic tasks :-) Please keep reporting issues that are most annoying during GHC development. I don't work on GHC myself, so have little idea how to prioritise.

PR #257, opened 2016-05-25
Packages/Rts: add `linker` to RTS directories

michalt:
Recent commit split off the m32 allocator to `rts/linker/`, which
broke the build using Hadrian (since it didn't know about the new
directory). This fixes it.


snowleopard:
Many thanks @michalt!

For some reason Travis OSX timed out, so I've restarted it. Probably just a glitch.

snowleopard:
Merged!

PR #258, opened 2016-05-30
Enable optional UserSettings.hs file

### LINKED ISSUE #247: Do not check src/Settings/User.hs into the repository
Any changes a user of Hadrian makes to that file should not get included into their GHC commits (by accident).

Possible solution:
* `git mv src/Settings/User.hs src/Settings/DefaultUser.hs` 
* In `build.sh`, if `src/Settings/User.hs` is missing, either abort or show a warning and `cp src/Settings/DefaultUser.hs src/Settings/User.hs`.

Checking out old revisions of Hadrian will cause some problems though.

KaiHa:
Fix #247

The defaul user settings are stored in ./src/UserSettings.hs. If the
user want to override these settings, he can copy this file into ./ and
make the desired changes to ./UserSettings.hs.


snowleopard:
Great, thanks!

The only thing to discuss: do we want users to copy `src/UserSettings.hs` or, as also mentioned previously, rename `./UserSettings.hs.sample`? The latter may be more familiar, as users currently rename `build.mk.sample`. The sample file may actually differ from `src/UserSettings.hs` and contain several common settings commented out.

snowleopard:
@KaiHa @ndmitchell @thomie @Ericson2314 Any thoughts on the above comment?

KaiHa:
> The only thing to discuss: do we want users to copy src/UserSettings.hs or, as also mentioned previously, rename ./UserSettings.hs.sample?

Both approaches are fine for me.

Ericson2314:
If I understand correctly, this uses the precedence of '.' over 'src/' to work?

KaiHa:
> If I understand correctly, this uses the precedence of '.' over 'src/' to work?

Correct

Ericson2314:
Anything with `.sample` would be more complex then. My vote is for this. Good work!

ndmitchell:
Looks very good to me. Thanks @KaiHa.

snowleopard:
OK, merged as is. Thanks again!

PR #260, opened 2016-06-03
Merge build.cabal-new.sh into build.cabal.sh

hvr:
The script now detect the cabal version and uses either the robust and
fast 'new-build'-logic (for version 1.24 or later), or falls back to the
fragile sandbox-based legacy logic.


snowleopard:
Thanks! :+1: 

I like the autodetection and that we have only one Cabal script now.

Can you explain why running Hadrian via `build.cabal.sh` is generally better than via `build.sh`? Is the situation similar on Windows?

We can edit the `README` to point out the recommended script, but hitting the users with a warning each time they are running `build.sh` seems excessive. If we clearly explain the difference between the scripts and their usage in `README` there should be no reason for adding the warning and complicating `build.sh` with Cabal version detection logic. I'd prefer to keep these basic `build.*` scripts as simple as possible so that mere mortals like me can understand them.

hvr:
@snowleopard well, for one, `build.sh` requires you to have installed the right package versions of hadrian's build-deps, most importantly `shake` and `ansi-terminal` and `QuickCheck` (I just tried, I needed 3 attempts, each time guessing which package to install next, until `build.sh` finally worked!) needs to be installed which can't be assumed to exist in general.

Also, `build.sh` assumes that you can actually `cabal install shake` which is an anti-pattern that future cabal's are trying to discourage/get-rid-of. So `build.sh` is using an inconvenient and deprecated method. If you *need* `build.sh` it means you don't have `cabal` at your disposal, but then you suffer from having to manually bootstrap `shake` and all its dependencies.

However, I recommend `build.cabal.sh` only when cabal-1.24+ is detected, as only then I consider it robust enough and superior to the alternative. There's also a bit of an agenda here: I want to have the new nix-local-build feature tested as widely as possible and break the chicken&egg issue, and hadrian/GHC is a perfect place for this, as any remaining issues discovered would be addressed quickly by me.

Also, cabal 1.24's `new-build` doesn't suffer anymore from corrupting your package database. I.e. you'll never see any message about `--force-reinstall` ever again, as it's not possible anymore to run into that state. Also, `cabal 1.24`

Finally, with cabal 1.24 `build.cabal.sh --help` is faster by an order of magnitude when there's nothing to compile!

```
$ time ./build.sh --help  > /dev/null

real	0m0.865s
user	0m1.972s
sys	0m1.168s
```

vs

```
$ time ./build.cabal.sh --help  > /dev/null

real	0m0.063s
user	0m0.024s
sys	0m0.024s
```

That's how good `cabal new-build` is! :-)

In short, there's IMO no reason *when* you have cabal-1.24+ to want to use the plain `build.sh`. And warning every time is the only way I can think of to reliably reach everybody to tell them there's a way better way given their current environment (not everybody reads READMEs carefully, and if there's a `build.sh` that's the one ppl are going to pick by default)

snowleopard:
Many thanks for the explanation @hvr!

A couple of follow-up comments:

* I presume this code is only temporary until `cabal` supports `new-run`? (It looks sadly complicated.)

    ```diff
    +    PKGVER="$(awk '/^version:/ { print $2 }' hadrian.cabal)"
    +    "./dist-newstyle/build/hadrian-${PKGVER}/build/hadrian/hadrian" \
    +        --lint                         \
    +        --directory "$absoluteRoot/.." \
    +        "$@"
    ```
* The speedup you refer to is due to the fact that `ghc --make` is so much slower than `cabal new-build`? Or are there any other hidden variables? (Also, what about the speed of actually building Hadrian? Will it also improve with `cabal new-build`?)

* As a Windows user I'd appreciate if you also provide the `bat` versions too as soon as possible, to avoid having different instructions on different platforms. I presume HP is not necessary to try out the new Cabal. Why wait for it? Not all Windows users rely on HP. Note, I'm running the build scripts on Windows _very often_ so I can discover some issues with the new Cabal features, helping your agenda :-)

PR #262, opened 2016-06-08
Note on nm on OS X

alexbiehl:
This places a note which shows how to mitigate #11744 with Hadrian.

I found it non obvious on which parameter to use with the `Configure` constructor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,13 @@ runs the `boot` and `configure` scripts automatically on the first build, so tha
 need to. Use `--skip-configure` to suppress this behaviour (see overview of command line
 flags below).
 
+* Also note on OS X newer versions of XCode ship with a broken `nm` tool ([#1174](https://ghc.haskell.org/trac/ghc/ticket/11744)). To mitigate the problem place something like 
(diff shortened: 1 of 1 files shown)


snowleopard:
Thanks @alexbiehl!

Out of curiosity, could you check if `--with-nm=nm-classic` works too? Hadrian does try to look up executables in `PATH` if no specific path is given. If it does, perhaps we should use this simpler version.

alexbiehl:
It actually depends on what your "active development directory" is. That depends which versions of the SDKs you installed.
For my machine it prints `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/nm-classic` which is certainly not in `PATH` (and thus the hadrian build failed).

 I think to be safe we should use the `xcrun` command. Thats also whats advised in the trac ticket.

snowleopard:
I see, thanks! Merged now.

snowleopard:
I think I'll rearrange README slightly to have a section on OS-specific notes, moving your note there.

PR #266, opened 2016-06-18
Add actions copyDirectoryContent and runBuilderWith

KaiHa:
These new functions will be helpful when implementing the 'sdist' and
'install' rules.

### DIFF
--- a/src/Oracles/DirectoryContent.hs
+++ b/src/Oracles/DirectoryContent.hs
@@ -1,0 +1,31 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+module Oracles.DirectoryContent (
+    getDirectoryContent, directoryContentOracle, Exclude(..), ExcludeNot(..)
+    ) where
+
+import Base
+import System.Directory.Extra
+
+newtype DirectoryContent = DirectoryContent (Exclude, ExcludeNot, FilePath)
+    deriving (Binary, Eq, Hashable, NFData, Show, Typeable)
+newtype Exclude = Exclude [FilePattern]
+    deriving (Binary, Eq, Hashable, NFData, Show, Typeable)
+newtype ExcludeNot = ExcludeNot [FilePattern]
+    deriving (Binary, Eq, Hashable, NFData, Show, Typeable)
+
+-- | Get the directory content. 'Exclude' and 'ExcludeNot' are a list of file
+-- patterns matched with '?=='.
+getDirectoryContent :: Exclude -> ExcludeNot -> FilePath -> Action [FilePath]
+getDirectoryContent exclude excludeNot dir =
+    askOracle $ DirectoryContent (exclude, excludeNot, dir)
+
+directoryContentOracle :: Rules ()
+directoryContentOracle = void $ addOracle oracle
+  where
+    oracle :: DirectoryContent -> Action [FilePath]
+    oracle (DirectoryContent (Exclude exclude, ExcludeNot excludeNot, dir)) =
+        liftIO $ filter test <$> listFilesInside (return . test) dir
+      where
+        test a = include' a || not (exclude' a)
+        exclude' a = any (?== a) exclude
+        include' a = any (?== a) excludeNot
--- a/src/Oracles/DirectoryContent.hs
+++ b/src/Oracles/DirectoryContent.hs
@@ -1,31 +1,39 @@
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving, DeriveGeneric #-}
 module Oracles.DirectoryContent (
-    getDirectoryContent, directoryContentOracle, Exclude(..), ExcludeNot(..)
+    getDirectoryContent, directoryContentOracle, Match(..)
     ) where
 
 import Base
+import GHC.Generics
 import System.Directory.Extra
 
-newtype DirectoryContent = DirectoryContent (Exclude, ExcludeNot, FilePath)
-    deriving (Binary, Eq, Hashable, NFData, Show, Typeable)
-newtype Exclude = Exclude [FilePattern]
-    deriving (Binary, Eq, Hashable, NFData, Show, Typeable)
-newtype ExcludeNot = ExcludeNot [FilePattern]
+newtype DirectoryContent = DirectoryContent (Match, FilePath)
     deriving (Binary, Eq, Hashable, NFData, Show, Typeable)
 
--- | Get the directory content. 'Exclude' and 'ExcludeNot' are a list of file
--- patterns matched with '?=='.
-getDirectoryContent :: Exclude -> ExcludeNot -> FilePath -> Action [FilePath]
-getDirectoryContent exclude excludeNot dir =
-    askOracle $ DirectoryContent (exclude, excludeNot, dir)
+data Match = Test FilePattern | Not (Match) | And [Match] | Or [Match]
+    deriving (Generic, Eq, Show, Typeable)
+instance Binary Match
+instance Hashable Match
+instance NFData Match
+
+matches :: Match -> FilePath -> Bool
+matches (Test m) f = m ?== f
+matches (Not m) f = not $ matches m f
+matches (And [])     _               = True
+matches (And (m:ms)) f | matches m f = matches (And ms) f
+                       | otherwise   = False
+matches (Or [])     _               = False
+matches (Or (m:ms)) f | matches m f = True
+                      | otherwise   = matches (Or ms) f
+
+-- | Get the directory content recursively.
+getDirectoryContent :: Match -> FilePath -> Action [FilePath]
+getDirectoryContent expr dir =
+    askOracle $ DirectoryContent (expr, dir)
 
 directoryContentOracle :: Rules ()
 directoryContentOracle = void $ addOracle oracle
   where
     oracle :: DirectoryContent -> Action [FilePath]
-    oracle (DirectoryContent (Exclude exclude, ExcludeNot excludeNot, dir)) =
-        liftIO $ filter test <$> listFilesInside (return . test) dir
-      where
-        test a = include' a || not (exclude' a)
-        exclude' a = any (?== a) exclude
-        include' a = any (?== a) excludeNot
+    oracle (DirectoryContent (expr, dir)) =
+        liftIO $ filter (matches expr) <$> listFilesInside (return . matches expr) dir
--- a/src/Rules/Actions.hs
+++ b/src/Rules/Actions.hs
@@ -2,7 +2,7 @@ module Rules.Actions (
     build, buildWithCmdOptions, buildWithResources, copyFile, fixFile, moveFile,
     removeFile, copyDirectory, copyDirectoryContent, createDirectory,
     moveDirectory, removeDirectory, applyPatch, runBuilder, runBuilderWith,
-    makeExecutable, renderProgram, renderLibrary, Exclude(..), ExcludeNot(..)
+    makeExecutable, renderProgram, renderLibrary, Match(..)
     ) where
 
 import qualified System.Directory.Extra as IO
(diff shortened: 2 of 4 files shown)


src/Rules/Actions.hs
```diff
@@ -1,12 +1,14 @@
...
-    applyPatch, runBuilder, makeExecutable, renderProgram, renderLibrary
+    removeFile, copyDirectory, copyDirectoryContent, createDirectory,
+    moveDirectory, removeDirectory, applyPatch, runBuilder, runBuilderWith,
+    makeExecutable, renderProgram, renderLibrary
     ) where
 
 import qualified System.Directory       as IO
 import qualified System.IO              as IO
 import qualified Control.Exception.Base as IO
+import qualified System.Directory.Extra as X
```
snowleopard: I suggest to use 'IO' for the qualification here as well.

src/Rules/Actions.hs
```diff
@@ -126,6 +128,18 @@ copyDirectory source target = do
     putProgressInfo $ renderAction "Copy directory" source target
     quietly $ cmd cmdEcho ["cp", "-r", source, target]
 
+-- | Copy the content of the source directory into the target directory. Only
+-- the files and directories for which the predicate returns True are copied.
+copyDirectoryContent :: (FilePath -> IO Bool) -> FilePath -> FilePath -> Action ()
```
snowleopard: Most probably it would be more convenient for callers if the predicate has `FilePath -> Action Bool` type.

src/Rules/Actions.hs
```diff
@@ -152,12 +166,16 @@ applyPatch dir patch = do
     quietly $ cmd Shell cmdEcho [Cwd dir] [path, "-p0 <", patch]
 
 runBuilder :: Builder -> [String] -> Action ()
-runBuilder builder args = do
+runBuilder =
+    runBuilderWith []
```
snowleopard: I'd use a single line here `runBuilder = runBuilderWith []`.

src/Rules/Actions.hs
```diff
@@ -126,6 +128,18 @@ copyDirectory source target = do
...
     quietly $ cmd cmdEcho ["cp", "-r", source, target]
 
+-- | Copy the content of the source directory into the target directory. Only
+-- the files and directories for which the predicate returns True are copied.
+copyDirectoryContent :: (FilePath -> IO Bool) -> FilePath -> FilePath -> Action ()
+copyDirectoryContent test source target = do
+    putProgressInfo $ renderAction "Copy directory" source target
+    liftIO $ X.listFilesInside test' source >>= mapM_ cp
+  where
+    target' a = target -/- fromJust (stripPrefix source a)
```
snowleopard: Sad that we have to use `fromJust`, but I can't see an easy way to avoid that...

src/Rules/Actions.hs
snowleopard: On a second thought, perhaps this should be simplified even further? Do we really need to test filenames in `IO` or `Action`? The only (so far) use case in `sdist` rule is just a pure function.

src/Rules/Actions.hs
```diff
@@ -126,6 +128,18 @@ copyDirectory source target = do
...
+    test' a = ifM (test a) (mkdir a >> return True) (return False)
+    mkdir a = IO.createDirectoryIfMissing True $ target' a
+    cp a = whenM (test a) $ IO.copyFile a $ target' a
```
snowleopard: By using system `copyFile` function you are not tracking the files being copied, that is, if they change and you rerun the `sdist` rule, Hadrian will fail to copy updated files. We either need to add `need` here, or use a tracking version of `copyFile`. However, there is an even deeper problem: we also need to depend on the result of `listFilesInside`, because some files may be added or deleted! I don't think this is currently tracked.

I think the right way to do something like this in Shake is to use Shake's `getDirectoryFiles` function. However, the problem may be that `getDirectoryFiles` does not provide a way to exclude certain files and/or directories... Maybe it should!

src/Rules/Actions.hs
ndmitchell: No need to import System.Directory.Extra and System.Directory since the former reexports the latter.

src/Rules/Actions.hs
ndmitchell: GetDirectoryFiles functions are a pain -
If you want more control use an oracle and write the traversal yourself or use the functions in extra. Pushing an arbitrary predicate into the tracked Shake ones is unfortunately impossible and what is most convenient.

snowleopard:
Thanks @KaiHa! (I've finally found some time for coding, hurrah!)

I added a bunch of comments above. Most of them a minor things, but there seem to be one important problem: tracking the files that are copied by `copyDirectoryContent`. At the moment it is clear that this function does not track the files, but the interface with a predicate `FilePath -> IO Bool` suggests that the caller will not track the files either. But somebody has to!

I start to think that we need an equivalent of `getDirectoryFiles` that will have an additional argument for excluding certain `FilePattern`'s during the search. @ndmitchell Do you think such a function will be generally useful in Shake? Or will you suggest a better way?

src/Rules/Actions.hs
snowleopard: @ndmitchell I think we actually don't need a predicate. We only need to exclude certain `FilePattern`'s. So, it's like going from
```haskell
getDirectoryFiles :: FilePath -> [FilePattern] -> Action [FilePath]
getDirectoryFiles dir include = do ...
```
to
```haskell
getDirectoryFiles' :: FilePath -> [FilePattern] -> [FilePattern] -> Action [FilePath]
getDirectoryFiles' dir include exclude = do ...
```
We can certainly try to implement it in Hadrian directly and see how it goes.

KaiHa: > Hadrian will fail to copy updated files. We either need to add need here, or use a tracking version of copyFile. However, there is an even deeper problem: we also need to depend on the result of listFilesInside, because some files may be added or deleted! I don't think this is currently tracked.

My plan for the 'sdist' and 'install' rules was to make them _phony_, so I have not to consider if something I have copied has changed. Is this an option?

snowleopard: @KaiHa Consider the following scenario:
```bash
$ build sdist # This builds a source distribution tarball
$ vi libraries/base/Prelude.hs # Edit sources
$ build sdist # Rebuild the source distribution tarball
```
Presumably, in the second invocation of `build sdist` you expect the tarball to be updated automatically, and that nothing should be done if `libraries/base/Prelude.hs` was actually unchanged. Always removing the tarball may consume more time than necessary. It doesn't matter whether the rule is _phony_, Hadrian can still provide certain correctness and performance guarantees, e.g. _Running `build sdist` guarantees that the source tarball is up-to-date and we only rebuilt what was necessary_. This is not that hard to achieve with Shake: just don't forget to track dependencies with `need` and friends.

KaiHa:
@snowleopard just like ndmitchell has suggested I have implemented the `copyDirectoryContent` function with an oracle.

src/Oracles/DirectoryContent.hs
```diff
@@ -0,0 +1,31 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+module Oracles.DirectoryContent (
+    getDirectoryContent, directoryContentOracle, Exclude(..), ExcludeNot(..)
+    ) where
+
+import Base
+import System.Directory.Extra
+
+newtype DirectoryContent = DirectoryContent (Exclude, ExcludeNot, FilePath)
```
snowleopard: Any particular reason for `ExcludeNot` instead of a more natural `Include`? Is it a name clash?

snowleopard: I see you even call it `include` later on in the code, which seems strange...

src/Oracles/DirectoryContent.hs
```diff
@@ -0,0 +1,31 @@
...
+    askOracle $ DirectoryContent (exclude, excludeNot, dir)
+
+directoryContentOracle :: Rules ()
+directoryContentOracle = void $ addOracle oracle
+  where
+    oracle :: DirectoryContent -> Action [FilePath]
+    oracle (DirectoryContent (Exclude exclude, ExcludeNot excludeNot, dir)) =
+        liftIO $ filter test <$> listFilesInside (return . test) dir
+      where
+        test a = include' a || not (exclude' a)
```
snowleopard: I'm a bit confused: shouldn't it be `&&` instead of `||`? What is the semantics? I thought it is _include these files unless they are explicitly excluded_, but it doesn't work with `||`.

KaiHa: > I'm a bit confused: shouldn't it be && instead of ||? What is the semantics? I thought it is include these files unless they are explicitly excluded, but it doesn't work with ||.

It is copy everything, only exclude the files that are matched by _Exclude_ unless they are matched by _ExcludeNot_.

snowleopard:
@KaiHa Great! Everything looks good, but see two comments above about `ExcludeNot` and `||`.

src/Oracles/DirectoryContent.hs
snowleopard: Ah, I see! That's more complicated than I thought :-) Please add this as a comment to the top function. 

Maybe we should at least rename `include'` to `excludeNot'` to avoid possible confusion.

snowleopard:
@KaiHa Once you are happy with this PR let me know and I'll merge it and test on my machine.

KaiHa:
> Great! Everything looks good, but see two comments above about ExcludeNot and ||.

The reason for the _Exclude_ and _ExcludeNot_ is that I want to be able to do something like this:
```hs
copyDirectoryContent (Exlcude ["//*.o"]) (ExcludeNot ["//HS*.o"]) "."
```

```
ghc
├── apps
│   └── bar
│       ├── bar     # copied
│       ├── bar.hs  # copied
│       └── bar.o   # NOT copied
└── libs
    └── foo
        ├── foo.hs  # copied
        └── HSfoo.o # copied
```

Therefore it is named _ExcludeNot_ and not _Include_. 

As it is not possible to use a predicate with oracles, this is the best solution I have found. Maybe I should Improve the documentation, or do you see any better (more intuitive) way of achieving the above?

KaiHa:
> Once you are happy with this PR let me know and I'll merge it and test on my machine.

I will improve the documentation and will let you know then I am satisfied.

ndmitchell:
I would have been tempted to define a small abstract syntax tree:

```
data Match a = Test a | Not (Match a) | And [Match a] | Or [Match a]

copyDirectoryContent :: Match FilePattern -> FilePath -> Action ()
```

That makes the semantics much clearer and more flexible. Only downside is that you have to write a `Binary` instance for `Match`, but with Generic deriving maybe that's not too bad either. It's also more flexible, if you ever need the power.

snowleopard:
@KaiHa Yes, I now see why you need `ExcludeNot`.

One possible alternative, which may be clearer, is to copy required files explicitly. Something like:
```haskell
copyContext :: Context -> FilePath -> Action ()
copyContext context dir = do
    -- Haskell sources
    srcs <- haskellSources context
    forM_ srcs $ \src -> copyFile src (dir -/- src)
    -- Library
    lib <- pkgLibraryFile context
    copyFile lib (dir -/- lib)
    -- plus a bunch of other files
```
This may be more work, but I like this approach because it is very explicit: we know and specify exactly what we need to copy instead of relying on `Exclude`/`ExcludeNot` or more general matching mechanism proposed by @ndmitchell.

This could be made even clearer by pulling all files into a list:
```haskell
sdistFiles :: [Context -> Action FilePath]
sdistFiles = [ haskellSources -- Haskell sources
             , pkgLibraryFile -- Library
             -- plus a bunch of other files
             ]
```
Now this works both as specification and documentation at the same time. Isn't it nice?

KaiHa:
@snowleopard I agree, that sounds nice. Unfortunately my [attempt to implement it](https://github.com/KaiHa/hadrian/compare/copyDirectoryContent...KaiHa:wip/sdist-ghc_explicit-copy) looks a lot messier. 

- How do I prevent generated files from sneaking in?
- Getting the `haskellSources` feels clumsy, am I doing something wrong?
- Are there functions to get the paths to all the other needed files or must I list them like I have done for the _hadrian_ files?

snowleopard:
@KaiHa Yes, it does look messier than I thought... all those extra source files (like `README`) and Hadrian sources are not easy to deal with. 

> How do I prevent generated files from sneaking in?

I think we should actually use `contextFiles` instead of `haskellSources`. The former returns actual source files (e.g. `*.x` files if need be) instead of possibly generated Haskell files.

> Getting the `haskellSources` feels clumsy, am I doing something wrong?

Do you perhaps mean Hadrian sources here? I agree, listing them explicitly is clumsy, but at the moment I don't see a clear alternative. We could in principle include `hadrian` as a `Package` in the build system and ask Hadrian to compute its own sources, but this feels crazy :)

> Are there functions to get the paths to all the other needed files or must I list them like I have done for the hadrian files?

I think `Settings/Paths.hs` and `Oracles/ModuleFiles.hs` contain all functions that compute package files. There are no others.

----

To sum up: I think your initial approach with `copyDirectoryContent` may be a more sensible way forward at the moment. The alternative seems not as clear as I thought. My mistake.

KaiHa:
I have implemented a small AST as proposed by @ndmitchell, that was easier than expected. Thanks for the hint on Generic deriving! @snowleopard if you find the result acceptable, please merge.

snowleopard:
Many thanks @KaiHa! I've merged this. I may do some minor revision later, hope you don't mind.

By the way, for some reason AppVeyor build failed, but this seems to be unrelated to your changes.

KaiHa:
Feel free to revise. I have no clue why AppVeyor is reporting a complie error.

PR #271, opened 2016-07-22
Change the stack configuration to use the local Cabal lib

michalt:
This seems to fix using `build.stack.sh` for me (I haven't finished building the whole GHC yet though).

I'm also not sure about the version bound on `Cabal` in the `hadrian.cabal`. What do you think?


snowleopard:
:+1:

snowleopard:
I'm not sure about the Cabal version too. Let's keep this for now.

PR #273, opened 2016-07-22
Fix inability to find gmp.h when building concurrently

mpickering:
There were situations when building concurrently when we would
request `gmp.h` before it had been built (or copied).

This was occuring when we generated the list of dependents for
the c files in the `integer-gmp` folder. Thus, when generating the
dependents for this library we now require `gmp.h`.


snowleopard:
Thank you!

PR #275, opened 2016-07-23
Split CompilerMode for GHC and CC

michalt:
I haven't completed a full build with this change yet, but the CI bots should pick up the PR .

### DIFF
--- a/src/Builder.hs
+++ b/src/Builder.hs
@@ -1,6 +1,6 @@
 {-# LANGUAGE DeriveGeneric, LambdaCase #-}
 module Builder (
-    CompilerMode (..), Builder (..),
+    CcMode (..), GhcMode (..), Builder (..),
     builderPath, getBuilderPath, builderEnvironment, specified, needBuilder
     ) where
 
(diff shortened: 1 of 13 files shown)


src/Builder.hs
```diff
@@ -19,10 +19,13 @@ import Stage
...
-                  | FindDependencies
-                  | Link
-                  deriving (Eq, Generic, Show)
+-- We have CcMode for CC and GhcMode for GHC.
+
+data CcMode = CcCompile | CcFindDependencies
+    deriving (Eq, Generic, Show)
+
+data GhcMode = GhcCompile | GhcFindDependencies | GhcLink
+    deriving (Eq, Generic, Show)
```
snowleopard: I think I have a better idea for naming modes:
```haskell
data CcMode = CompileC | FindCDependencies

data GhcMode = CompileHs | FindHsDependencies | LinkHs
```
We'll later add `CompileCmm` and `CompileS` to the latter.

michalt:
@snowleopard Done.

snowleopard:
Great, thanks @michalt!

PR #276, opened 2016-07-23
Use nm-classic instead of nm when host is Darwin

wereHamster:


src/Settings/Builders/Configure.hs
```diff
@@ -23,4 +25,11 @@ configureBuilderArgs = mconcat
...
                , "--enable-shared=no" -- TODO: add support for yes
                , "--host=" ++ targetPlatform ]
+
+    -- On OS X, use "nm-classic" instead of "nm" due to a bug in the later.
+    -- See https://ghc.haskell.org/trac/ghc/ticket/11744
+    , builder (Configure ".") ?
+        if System.os == "darwin"
+            then arg "--with-nm=$(xcrun --find nm-classic)"
+            else mempty
     ]
```
snowleopard: There is a more concise way to express this:
```haskell
System.os == "darwin" ? arg "--with-nm=$(xcrun --find nm-classic)"
```

snowleopard:
Looks good, but please see my comment.

wereHamster:
Updated the commit to use the new syntax.

snowleopard:
Hmm, Travis OSX instance failed:
```
xcrun: error: unable to find utility "nm-classic", not a developer tool or in PATH
[...]
Non optional builder 'nm' is not specified in system.config file.
```
Can we make `--with-nm=$(xcrun --find nm-classic)` conditional, that is, if the lookup fails, we shouldn't pass the `--with-nm=` flag to `configure`.

snowleopard:
As discussed offline, another way to fix the Travis OSX failure is to somehow install `nm-classic` there.

hvr:
You shouldn't use `nm-classic` blindly, only when `nm` is known to be broken. Apple will fix `nm` sooner or later, so that `nm-classic` should rather be considered a transitional workaround. Otoh, unfortunately, we can't fallback to `nm-classic` via Autoconf automatically, as that's against the Autoconf philosophy.
The best we can do in line with Autoconf conventions is to detect that `nm` is broken, and ask the user to provide us a different one.

snowleopard:
> The best we can do in line with Autoconf conventions is to detect that nm is broken, and ask the user to provide us a different one.

@hvr This actually sounds good enough to me. Is this check already implemented? Does `configure` advise to user to rerun it with `"--with-nm=$(xcrun --find nm-classic)"` flag?

> You shouldn't use nm-classic blindly, only when nm is known to be broken. Apple will fix nm sooner or later, so that nm-classic should rather be considered a transitional workaround.

Yes, it is just a temporary workaround. It will be dropped as soon as `nm` is fixed.

snowleopard:
@wereHamster Do you have any ideas how to resolve the Travis failure?

wereHamster:
> Does configure advise to user to rerun it with "--with-nm=$(xcrun --find nm-classic)" flag?

No, this happens much later during build time, after `deriveConstants` is built and executed, only then we find out that `nm` is broken. I'll try to move the check to configure.ac.

As for the travis failure, it seems that travis is running a much older version of Mac OS X and/or XCode. This should all go away once we have a proper check in configure.ac.

On platforms which use `nm` (because not all do, some use `objdump` instead, such as AIX. Note to myself: find out why and document):

 1. Check if `nm` produces the correct result.
 2. If $build == darwin:
   1. Try to locate `nm-classic` and retry with it.
 3. If we don't have a working `nm`, fail hard.

snowleopard:
@wereHamster I've finally merged this -- Travis has got `nm-classic` now, so this works fine. Thank you!

PR #278, opened 2016-07-26
Make generatedDependencies an Expr [FilePath]

michalt:
This fixes a TODO to change the `generatedDependencies` to
use `Expr`.

### DIFF
--- a/src/Rules/Generate.hs
+++ b/src/Rules/Generate.hs
@@ -10,6 +10,8 @@ import Context
 import Expression
 import GHC
 import Oracles.ModuleFiles
+import Predicate ( (?) )
+import qualified Predicate as Predicate
 import Rules.Actions
 import Rules.Generators.ConfigHs
 import Rules.Generators.GhcAutoconfH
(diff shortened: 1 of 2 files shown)


src/Rules/Generate.hs
```diff
@@ -85,15 +85,18 @@ compilerDependencies stage =
...
+    pkg <- getPackage
+    stage <- getStage
+    let deps
+            | pkg   == compiler = compilerDependencies stage
+            | pkg   == ghcPrim  = ghcPrimDependencies stage
+            | pkg   == rts      = libffiDependencies ++ includesDependencies
+                               ++ derivedConstantsDependencies
+            | stage == Stage0   = includesDependencies
+            | otherwise         = []
+    return deps
```
snowleopard: I was actually thinking to use predicates, e.g.:
```haskell
generatedDependencies = mconcat
   [ package compiler ? compilerDependencies
   , ...
   , stage0           ? includeDependencies ]
```
And make `compilerDependencies` and `includeDependencies` expressions too.

snowleopard:
Thanks @michalt! See my comment above. I think if we turn all these lists into `Expr`'s the result will be more readable.

michalt:
@snowleopard Yeah, that looks much better! (I wasn't aware of `Predicate`) I haven't converted things like `platformH` since it's used in a few places not touched by this PR. Let me what you think!

snowleopard:
@michalt Great! Let me merge this, but I'll do a minor revision: `Predicate` is used without qualification in the rest of the build system, plus I'll refactor the code a little. I'll ping you when I'm done.

snowleopard:
@michalt Done! By the way, if I understand correctly, most of these `Expr [FilePath]` will become obsolete once you implement your clever fix-point search of generated dependencies -- is that right?

PR #279, opened 2016-08-02
Use GCC's `-MM`/`-MG` to find missing dependencies

michalt:
So this is probably not ready to be merged, but I wanted to discuss a few things :-)

- I'm not sure what we could clean up - I'm planning to have a look at the places that use the current lists of generated files and see whether we can remove some of the explicit calls to `need` in favor of this approach. Ideas are welcome!

- It seems that applying the same trick to `.hs` files might be a bit more complicated - we can try to tell GCC to treat it just as a `.c` or `.h` file and stop at pre-processing, but things like `MIN_VERSION_*` still cause it to die with an error (since they're not defined).

- Does GHC actually have a case where a generated file `#include`s another generated file? (if not we could just go with a bit more brute-force approach of grepping the source?).

- Finally, I'm wondering if we should create a cache/oracle for this information - it's pretty easy to have the same file being analyzed multiple times just by the virtue of being a dependency for many other `.c` files.

### DIFF
--- a/src/Builder.hs
+++ b/src/Builder.hs
@@ -21,7 +21,8 @@ import Stage
 -- 3) Linking object files & static libraries into an executable.
 -- We have CcMode for CC and GhcMode for GHC.
 
-data CcMode = CompileC | FindCDependencies
+-- TODO: Consider merging FindCDependencies and FindMissingInclude
+data CcMode = CompileC | FindCDependencies | FindMissingInclude
     deriving (Eq, Generic, Show)
 
 data GhcMode = CompileHs | FindHsDependencies | LinkHs
--- a/src/Rules/Compile.hs
+++ b/src/Rules/Compile.hs
@@ -5,8 +5,15 @@ import Context
 import Expression
 import Oracles.Dependencies
 import Rules.Actions
+import Rules.Generate
 import Settings.Paths
 import Target
+
+import Development.Shake.Util
+
+import Data.Maybe
+import Data.List
+import qualified Data.Set as Set
 
 compilePackage :: [(Resource, Int)] -> Context -> Rules ()
 compilePackage rs context@Context {..} = do
@@ -22,6 +29,9 @@ compilePackage rs context@Context {..} = do
         if ("//*.c" ?== src)
         then do
             need $ src : deps
+            -- TODO: Improve parallelism by collecting all dependencies and
+            -- need'ing them all at once
+            mapM_  (needGenerated context) . filter ("//*.c" ?==) $ src : deps
             build $ Target context (Cc CompileC stage) [src] [obj]
         else do
             need $ src : deps
@@ -39,3 +49,33 @@ needCompileDependencies :: Context -> Action ()
 needCompileDependencies context@Context {..} = do
     when (isLibrary package) $ need =<< return <$> pkgConfFile context
     needContext =<< contextDependencies context
+
+needGenerated :: Context -> FilePath -> Action ()
+needGenerated context origFile = go Set.empty
+  where
+    go :: Set.Set String -> Action ()
+    go done = withTempFile $ \outFile -> do
+        let builder = Cc FindMissingInclude $ stage context
+            target = Target context builder [origFile] [outFile]
+        build target
+        deps <- parseFile outFile
+
+        -- Get the full path if the include refers to a generated file and call
+        -- `need` on it.
+        needed <- liftM catMaybes $
+            interpretInContext context (mapM getPathIfGenerated deps)
+        need needed
+
+        let newdone = Set.fromList needed `Set.union` done
+        -- If we added a new file to the set of needed files, let's try one more
+        -- time, since the new file might include a genreated header of itself
+        -- (which we'll `need`).
+        when (Set.size newdone > Set.size done) (go newdone)
+
+    parseFile :: FilePath -> Action [String]
+    parseFile file = do
+        input <- liftIO $ readFile file
+        case parseMakefile input of
+            [(_file, deps)] -> return deps
+            _               -> return []
+
--- a/src/Rules/Generate.hs
+++ b/src/Rules/Generate.hs
@@ -1,6 +1,7 @@
 module Rules.Generate (
     generatePackageCode, generateRules, installTargets, copyRules,
-    includesDependencies, derivedConstantsPath, generatedDependencies
+    includesDependencies, derivedConstantsPath, generatedDependencies,
+    getPathIfGenerated
     ) where
 
 import qualified System.Directory as IO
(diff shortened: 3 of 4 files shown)


src/Rules/Compile.hs
```diff
@@ -39,3 +47,35 @@ needCompileDependencies :: Context -> Action ()
...
+        build target
+        deps <- parseFile outFile
+
+        -- Get the full path if the include refers to a generated file and call
+        -- `need` on it.
+        needed <- liftM catMaybes $
+            interpretInContext context (mapM getPathIfGenerated deps)
+        need needed
+
+        let newdone = foldl' (flip Set.insert) done needed
```
ndmitchell: I tend to find it to be clearer to use: Set.fromList needed `Set.union` done - it's probably more efficient too, but that's very much a side benefit.

src/Rules/Compile.hs
```diff
@@ -39,3 +47,35 @@ needCompileDependencies :: Context -> Action ()
...
+        -- If we added a new file to the set of needed files, let's try one more
+        -- time, since the new file might include a genreated header of itself
+        -- (which we'll `need`).
+        if Set.size newdone > Set.size done
+           then go newdone
```
ndmitchell: I'd use a when here, when (Set.size newdone > Set.size done) $ go newdone

ndmitchell:
I don't see any way to avoid running this once per .c file, so an oracle probably won't help - different header files might include different things depending on their originating .c file.

Any idea of the performance of this pass? Does it mean we now need to run each C file through this pass, then subsequently through normal compilation?

simonmar:
> Does GHC actually have a case where a generated file #includes another generated file? (if not we could just go with a bit more brute-force approach of grepping the source?).

I wouldn't be at all surprised if we #included things that transitively include `ghcautoconf.h` or `DerivedConstants.h`, which are generated.

src/Builder.hs
```diff
@@ -21,7 +21,7 @@ import Stage
 -- 3) Linking object files & static libraries into an executable.
 -- We have CcMode for CC and GhcMode for GHC.
 
-data CcMode = CompileC | FindCDependencies
+data CcMode = CompileC | FindCDependencies | FindMissingInclude
     deriving (Eq, Generic, Show)
```
snowleopard: By the way, I think we don't need both `FindCDependencies` and `FindMissingInclude`, so at some point we'll probably drop one of them. Maybe add a TODO: `TODO: Consider merging FindCDependencies and FindMissingInclude`.

src/Rules/Compile.hs
```diff
@@ -22,6 +29,7 @@ compilePackage rs context@Context {..} = do
         if ("//*.c" ?== src)
         then do
             need $ src : deps
+            mapM_  (needGenerated context) . filter ("//*.c" ?==) $ src : deps
```
snowleopard: Another TODO: `TODO: Improve parallelism by collecting all dependencies and need'ing them all at once.`

src/Rules/Compile.hs
```diff
@@ -39,3 +47,35 @@ needCompileDependencies :: Context -> Action ()
 needCompileDependencies context@Context {..} = do
     when (isLibrary package) $ need =<< return <$> pkgConfFile context
     needContext =<< contextDependencies context
+
+needGenerated :: Context -> FilePath -> Action ()
+needGenerated context origFile = go Set.empty
+  where
+    go :: Set.Set String -> Action ()
+    go done = withTempFile $ \outFile -> do
+        let mode = Cc FindMissingInclude $ stage context
```
snowleopard: This should be called `builder`, not `mode`?

src/Rules/Generate.hs
```diff
@@ -196,3 +197,19 @@ generateRules = do
...
+getPathIfGenerated include = do
+  generated <- generatedFiles
+  -- For includes of generated files, we cannot get the full path of the file
+  -- (since it might be included due to some include dir, i.e., through `-I`).
+  -- So here we try both the name and the path.
+  let nameOrPath (name, path) = include == name || include == path
+  case find nameOrPath generated of
+    Just (_, filePath) -> return $! Just filePath
+    Nothing            -> return Nothing
+
```
snowleopard: Is this whole thing just `return $ fmap snd $ find nameOrPath generated`?

snowleopard:
@michalt Thanks for the working on this!

Here are some responses to your questions:

> I'm not sure what we could clean up [...]

I would suggest not to worry about this for now. We can separate these two issues: once we have a reliable way to detect & `need` generated files, we'll see what we can simplify in the build system. 

> It seems that applying the same trick to .hs files might be a bit more complicated [...]

Yes, I suspected this. But again, let's merge the first step which only deals with C files. Then we can think about generalising the approach to work with Haskell files too (e.g. we could set missing defines via `-D`).

The other two questions seem to be handled by @ndmitchell and @simonmar. 

There are a few comments by @ndmitchell and myself -- please have a look and I think we can merge.

michalt:
Thanks for all the comments! :-)

I'll try to fix everything this week (I've been slow to respond due to
traveling).

michalt:
I meant the upcoming week :-)

src/Rules/Compile.hs
michalt: Done.

src/Rules/Compile.hs
michalt: Done.

src/Builder.hs
michalt: Ok, added a TODO.

src/Rules/Compile.hs
michalt: Done.

src/Rules/Compile.hs
michalt: Yes, thanks for noticing! (I refactored the code a bit but forgot to change the name)

src/Rules/Generate.hs
michalt: Nice! Done.

snowleopard:
@michalt Thanks for fixing the comments! Is this ready to be merged?

michalt:
@snowleopard Yes, I think it's ok to merge (AppVeyor failed, but it seems that it's unrelated). After merging, what do you think our next step should be?

snowleopard:
@michalt Thanks, merged! 

> After merging, what do you think our next step should be?

I think we should now try to make the best use of this feature and avoid explicit `need`s on some generated headers. I guess you are the best person to try to tackle this -- would you like to try?

michalt:
@snowleopard Sure! I was just wondering if you already have something in mind. In any case, I'll probably just try a few things over the weekend. :-)

snowleopard:
@michalt Awesome! :-)

Get in touch if you'd like to discuss anything. I can do some experiments/coding too. Perhaps we could set up a separate issue, so we have everything in one place and others can comment.

michalt:
@snowleopard Opened #285

Issue #283, opened 2016-08-20
build.stack.sh mixes up stackage snapshots


cocreature:
When i try to compile hadrian (after `clean` and removing hadrian/.stack-work just to be sure) by build fails with one of the following two messages:

```
| Run Happy (stage = Stage0, package = genprimopcode): utils/genprimopcode/Parser.y => _build/stage0/utils/genprimopcode/Parser.hs
happy: /home/moritz/.stack/snapshots/x86_64-linux/nightly-2016-01-06/7.10.3/share/x86_64-linux-ghc-7.10.3/happy-1.19.5/HappyTemplate-arrays-coerce: openFile: does not exist (No such file or directory)
shakeArgsWith   0.000s    0%                           
Function shake  0.247s    8%  ==                       
Database read   0.061s    2%                           
With database   0.002s    0%                           
Running rules   2.484s   88%  =========================
Total           2.794s  100%                           
Error when running Shake build system:
* inplace/bin/genprimopcode
* _build/stage0/utils/genprimopcode/Parser.o
* OracleQ (ObjDepsKey ("_build/stage0/utils/genprimopcode/.dependencies","_build/stage0/utils/genprimopcode/Parser.o"))
* _build/stage0/utils/genprimopcode/.dependencies
* _build/stage0/utils/genprimopcode/.hs-dependencies
* _build/stage0/utils/genprimopcode/Parser.hs
user error (Development.Shake.cmd, system command failed
Command: /home/moritz/.local/bin/happy -agc --strict utils/genprimopcode/Parser.y -o _build/stage0/utils/genprimopcode/Parser.hs
Exit code: 1
Stderr:
happy: /home/moritz/.stack/snapshots/x86_64-linux/nightly-2016-01-06/7.10.3/share/x86_64-linux-ghc-7.10.3/happy-1.19.5/HappyTemplate-arrays-coerce: openFile: does not exist (No such file or directory)
)
```
```
alex: /home/moritz/.stack/snapshots/x86_64-linux/nightly-2016-02-15/7.10.3/share/x86_64-linux-ghc-7.10.3/alex-3.1.7/AlexTemplate-ghc-nopred: openFile: does not exist (No such file or directory)
shakeArgsWith   0.000s    0%                           
Function shake  0.225s    8%  ==                       
Database read   0.067s    2%                           
With database   0.002s    0%                           
Running rules   2.490s   89%  =========================
Total           2.784s   99%                           
Error when running Shake build system:
[...]

PR #284, opened 2016-08-21
Bump stack to lts-6.12 and remove extra pkg from extra-deps

michalt:
Apart from simplifying `stack.yaml`, removing explicit `extra` will also make the upgrade to GHC 8.0 easier (once stackage LTS moves to it).
`extra` is no longer necessary since it's included on stackage:  https://www.stackage.org/lts-6.12/package/extra-1.4.10

I've tested this by building GHC locally.


snowleopard:
@michalt Thanks! I was about to do this :-)

/cc #283

PR #286, opened 2016-09-03
Lowercase flavour names in --help

cocreature:
This corresponds to the actual argument names accepted by --flavour which is what most people will be looking for if they run --help.


snowleopard:
@cocreature I agree, this is better. Thanks for the fix!

PR #290, opened 2016-10-08
Build mkUserGuidePart with stage-0

bgamari:
This addresses GHC #12619, allowing the users guide to be built with only the stage 0 compiler.

Also eliminates some `fromJust`s that I encountered when putting this together.

### DIFF
--- a/src/Builder.hs
+++ b/src/Builder.hs
@@ -98,11 +98,14 @@ isOptional = \case
     Objdump  -> True
     _        -> False
 
--- TODO: Get rid of fromJust.
 -- | Determine the location of a 'Builder'.
 builderPath :: Builder -> Action FilePath
 builderPath builder = case builderProvenance builder of
-    Just context -> return . fromJust $ programPath context
+    Just context
+      | Just path <- programPath context -> return path
+      | otherwise                        ->
+        error $ "Cannot determine builderPath for " ++ show builder
+             ++ " in context " ++ show context
     Nothing -> case builder of
         Alex          -> fromKey "alex"
         Ar            -> fromKey "ar"
--- a/src/Builder.hs
+++ b/src/Builder.hs
@@ -104,6 +104,7 @@ builderPath builder = case builderProvenance builder of
     Just context
       | Just path <- programPath context -> return path
       | otherwise                        ->
+        -- TODO: Make builderPath total.
         error $ "Cannot determine builderPath for " ++ show builder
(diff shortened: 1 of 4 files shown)


bgamari:
I have clearly missed some subtlety in this change. @snowleopard, do you have any idea what might be going on here?

src/Settings/Default.hs
```diff
@@ -105,7 +105,8 @@ defaultPackages = mconcat
 packagesStage0 :: Packages
 packagesStage0 = mconcat
     [ append [ binary, cabal, compiler, ghc, ghcBoot, ghcBootTh, ghcCabal
-             , ghcPkg, hsc2hs, hoopl, hpc, templateHaskell, transformers ]
+             , ghcPkg, hsc2hs, hoopl, hpc, templateHaskell, transformers
+             , mkUserGuidePart ]
     -- the stage0 predicate makes sure these packages are built only in Stage0
     , stage0 ? append [ deriveConstants, dllSplit, genapply, genprimopcode
                       , hp2ps, unlit ]
```
snowleopard: You need to add `mkUserGuidePart` here. The above list actually gets built both in `Stage0` and in `Stage1`.

snowleopard:
@bgamari Thanks! See my comment above to fix the breakage.

I've long wanted to refactor the way I define what gets built when -- the current approach is quite confusing, sorry! (Note the `TODO: Simplify`).

src/Builder.hs
```diff
@@ -98,11 +98,14 @@ isOptional = \case
     Objdump  -> True
     _        -> False
 
--- TODO: Get rid of fromJust.
```
snowleopard: By the way, I actually wanted something more in this TODO: I think eventually we'll be able to get rid of partiality here altogether, by keeping the path to the executable in the `Package` datatype, e.g. `data Package = Library | Package FilePath`. So, perhaps, let's change the TODO to `Get rid of partiality`?

src/Settings/Default.hs
bgamari: Bah, of course! Thanks for catching that.

snowleopard:
@bgamari I wonder what's going on with the CI... AppVeyor seems to have failed while downloading, so I restarted it, but Linux & Mac OS seem to be broken in a way unrelated to your changes. Any ideas? Shall I merge?

snowleopard:
OK, AppVeyor build succeeded, so I'm merging.

PR #292, opened 2016-11-06
stack.yaml: bump to lts-7.6

michalt:


snowleopard:
I think Hadrian cannot require GHC 8 for bootstrapping according to `configure.ac`:
```
FP_COMPARE_VERSIONS([$GhcVersion],[-lt],[7.10],
    [AC_MSG_ERROR([GHC version 7.10 or later is required to compile GHC.])])
```
So, I guess we can't use `lts-7.6`. Perhaps, we can bump to `lts-6.24`, which is still GHC 7.10.3?

michalt:
@snowleopard I might be missing something but the snippet from `configure.ac` seems to indicate that you need *at least* 7.10. GHC for sure can be bootstrapped with GHC 8 (I've been using that for a while and I know other people have as well).
Whether we should default to GHC 7.10 or GHC 8 is a different story and either answer is equally problematic:
- If `stack.yaml` uses `lts-6.x` then people who use GHC 8 need to get GHC 7.10.
- If `stack.yaml` uses `lts-7.x` then people who use GHC 7.10 need to get GHC 8.

I'd probably lean towards the `lts-7.6`/GHC 8 since most people hacking on GHC probably do have GHC 8 around (to try out new features, etc.)

snowleopard:
@michalt I meant requiring GHC 8.0 is too harsh -- as far as I understand the build system should support two previous releases, so Hardian should be buildable with GHC 7.10.

> Whether we should default to GHC 7.10 or GHC 8 is a different story and either answer is equally problematic

I agree that both choices are problematic, but not equally so. If we default to GHC 8 we may accidentally start relying on some GHC 8 features and that won't be picked up by CI bots.

> most people hacking on GHC probably do have GHC 8 around

Agreed. Perhaps, the right way forward would be to have CI bots checking both GHC 7.10 and GHC 8 somehow. That would ensure Hadrian supports both, and will allow us to safely bump to `lts-7.x`.

Not sure what would be the right way to achieve that though: our Travis & AppVeyor bots at present are already very close to the maximum allowed CI time (and occasionally fail to finish). Adding anything else into a single CI job would make this even more fragile. Can we run two CI jobs for a single commit? That would be great.

michalt:
Oh, I didn't know the CI bots were using stack.

> Agreed. Perhaps, the right way forward would be to have CI bots checking both GHC 7.10 and GHC 8 somehow. That would ensure Hadrian supports both, and will allow us to safely bump to lts-7.x.
>
> Not sure what would be the right way to achieve that though: our Travis & AppVeyor bots at present are already very close to the maximum allowed CI time (and occasionally fail to finish). Adding anything else into a single CI job would make this even more fragile. Can we run two CI jobs for a single commit? That would be great.

Testing against both would be ideal, unfortunately I know very little about configuring Travis CI or AppVeyor... :-/

Maybe you're right and we should simply hold off with this.. (I'll probably still build with GHC 8 locally and 7.10 will be tested through CI)

snowleopard:
> Testing against both would be ideal, unfortunately I know very little about configuring Travis CI or AppVeyor... :-/
Maybe you're right and we should simply hold off with this.. (I'll probably still build with GHC 8 locally and 7.10 will be tested through CI)

Let's keep this PR open. I'll try to experiment with running CI jobs for both GHC 7.10 and GHC 8.0 and if this works out we'll default to GHC 8.0.

mboes:
My thoughts: I agree with @snowleopard that we should check that there are no regressions with 7.10. At the same time, there are advantages to using newer versions of dependencies as found in newer LTS releases for developers day-to-day (as such this newer LTS should probably be the default). So building both sounds like the right way forward to me.

snowleopard:
Thanks @mboes! 

> So building both sounds like the right way forward to me.

Yes, it's been on my todo list for a while. Perhaps, we could address it as part of #322. (/cc @izgzhen)

@michalt Sorry for taking so long to look into this.

izgzhen:
Ping back: our newly added CircleCI OS X build uses ghc-8.0.2. It works fine.

snowleopard:
@izgzhen So, ideally we need to have CI that tests both GHC 7.10 and GHC 8.0. Then we can safely bump the default GHC to 8.0, knowing that if we introduce a regression w.r.t. GHC 7.10 that will be detected.

How do we achieve that?

izgzhen:
> How do we achieve that?

Travis has build matrix: https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix.

snowleopard:
Yes, but on Travis some builds time out. Is there something like this for CircleCI too?

izgzhen:
> Yes, but on Travis some builds time out. Is there something like this for CircleCI too?

Do you mean building with a older GHC on CircleCI?

snowleopard:
> Do you mean building with a older GHC on CircleCI?

Yes, can we set up a build matrix on CircleCI instead of Travis which is slow?

izgzhen:
> Yes, can we set up a build matrix on CircleCI instead of Travis which is slow?

I think I can try it

izgzhen:
> > Yes, can we set up a build matrix on CircleCI instead of Travis which is slow?
> 
> I think I can try it

I tried [here](https://circleci.com/gh/izgzhen/hadrian/74), guided by https://circleci.com/docs/2.0/configuration-reference/#jobs. However, this experimental config failed, and I am not very confident that matrix is well-supported on Circle CI neither. So, maybe we can only rely on Linux ones for a while.

PR #300, opened 2017-02-12
Try --depth 1 for Appveyor build

gracjan:


gracjan:
Well, this needs to be tested after ghc.haskell.org is upgraded.

snowleopard:
Thanks @gracjan! I'll retry this PR once `ghc.haskell.org` is upgraded

izgzhen:
Why not using the GHC on GitHub instead like Travis? I trie a bit but this https://ci.appveyor.com/project/izgzhen/hadrian/build/1.0.1 failed out of some weird reason

snowleopard:
@izgzhen Yes, we can use the GihHub mirror. In fact we did use it the past, including the `--depth 1` flag proposed in this PR, e.g. see this old snapshot:

https://github.com/snowleopard/hadrian/blob/3a04d3425a06ff9ad62eb71d20ddd7cce26ae0a0/appveyor.yml#L20-L25

However, at some point I decided to simplify the build script and have rewritten, trying to make it closer to the recommended way of building GHC on Windows ([doc](https://github.com/snowleopard/hadrian/blob/master/doc/windows.md)).

I don't really mind to revert back to cloning from GitHub as long as it brings us some time saving compared to the current simpler solution.

snowleopard:
Ideally, I'd like us to test the build script in the [doc](https://github.com/snowleopard/hadrian/blob/master/doc/windows.md) via AppVeyor, so it doesn't get out of date.

izgzhen:
I tried what this PR tried again, and failed again. What about closing this PR and restate the problem in an issue instead?

snowleopard:
@izgzhen Good idea. Also, we need to quantify how much it actually helps. If this only shaves off 1% of the build time, then it may be not worth the effort -- the usual variability of build times is around 10% anyway.

snowleopard:
@gracjan Thank you for the PR, but it looks like it may require a bit more digging to implement it.

PR #303, opened 2017-03-31
Added a note about git line-ending settings

mechkg:
Added a note about git's default line-ending settings on Windows that can break the build and require starting over with the correct setting.


snowleopard:
@mechkg Thanks for reporting and fixing!

Was the issue with GHC files or with Hadrian files? If the latter, we could add `.gitattributes` to the repository to override user's `core.autocrlf` setting, as described [here](https://help.github.com/articles/dealing-with-line-endings/), which would be more convenient.

If the problems were with GHC files, let me know and I'll merge this.

snowleopard:
Reading your email, it looks like the issue was with GHC files, so I'm merging. Thanks!

PR #305, opened 2017-04-11
Add wrapper for Runhaskell, Fix #304

### LINKED ISSUE #304: runhaskell can't find ghc
Go to `inplace/bin` and

```
> ./runhaskell
runhaskell: cannot find ghc
```

This will happen for GHC 65b185d4886b4efa3efe3cc5ecc8dd6e07d89afe and Hadrian 295c7812ab8fe8e34e6812127e039220a850a344 on mu Ubuntu 15.04 machine

izgzhen:


izgzhen:
It works for me, but note the HACK part. I guess there should be a more disciplined way to fetch the path of `ghc-stage2`

snowleopard:
@izgzhen Thank you! I think I can fix the HACK bit.

We have a strange failure of Travis though: I've no idea how your change led to recursive build rules, but somehow that did happen... Let me try to understand what's going on.

snowleopard:
@izgzhen Does Hadrian still work for you from a clean build after this change?

izgzhen:
> Does Hadrian still work for you from a clean build after this change?
Yes, it works for me.

snowleopard:
@izgzhen Thanks! I've restarted the CI to see if the error persists.

izgzhen:
Looks like Travis is still not happy, and by the way:

1. My Linux one is GHC 7.10.3. and seemly works fine
2. I also tried it on my macOS Sierra, GHC 8.0.1. And **it break just like Travis**:

```
ghci-8.3: Warning: haddock-interfaces: /Users/zz/Repos/ghc/_build/stage1/libraries/ghci/doc/html/ghci/ghci.haddock doesn't exist or isn't a file
| Remove file _build/stage1/libraries/Cabal/Cabal/libHSCabal-2.0.0.0.a
/----------------------------------------------------\
| Run GhcCabal (stage = Stage1, package = ghc)       |
|      input: compiler/ghc.cabal                     |
|  => output: _build/stage1/compiler/package-data.mk |
\----------------------------------------------------/
shakeArgsWith    0.000s    0%                           
Function shake   0.369s    0%                           
Database read    0.000s    0%                           
With database    0.000s    0%                           
Running rules  711.839s   99%  ======================== 
Total          712.209s  100%                           
* inplace/bin/iserv-bin
* OracleQ (PackageDataKey ("_build/stage1/iserv/package-data.mk","COMPONENT_ID"))
* _build/stage1/iserv/package-data.mk
* OracleQ (PackageDataKey ("_build/stage1/iserv/package-data.mk","COMPONENT_ID"))
Build system error - recursion detected:
  Key type:   OracleQ PackageDataKey
  Key value:  OracleQ (PackageDataKey ("_build/stage1/iserv/package-data.mk","COMPONENT_ID"))
Rules may not be recursive
```

izgzhen:
A rough guess is that I should put `runhaskell` in a later build stage since it depends on `ghc-stage2` -- but I am still struggling in finding how to do that

izgzhen:
BTW, I wonder why it is `runhaskell` rather than `runghc`. I found such weird co-existing names in my prior Haskell experience as well, but never wrap my head around why it is do.

snowleopard:
@izgzhen Many thanks for investigating this! 

> A rough guess is that I should put runhaskell in a later build stage since it depends on ghc-stage2 -- but I am still struggling in finding how to do that

This can be done by moving `runGhc` to `stage2Packages`, see here:

https://github.com/snowleopard/hadrian/blob/master/src/Settings/Default.hs#L114-L146

However, I'm not sure we should be doing that, because the Make build system build it in Stage1.

> And how would that have anything to do with iserv?

I have no idea, this looks like a Hadrian bug to me, but I can't figure out the reason.

> BTW, I wonder why it is runhaskell rather than runghc. I found such weird co-existing names in my prior Haskell experience as well, but never wrap my head around why it is so.

Indeed, there are many such instances in the GHC source tree. I guess most of these are just historical artefacts that people learned to live with, but which could be easily fixed to make the codebase easier to understand for newcomers.

snowleopard:
Ahh! Mystery resolved.

Your PR is not the cause of failure. I've rebuilt one of the previous commits and it failed too -- this must be some new changes in GHC that caused a loop in the build system.

I'm merging your PR, thank you!

snowleopard:
> BTW, I wonder why it is runhaskell rather than runghc. I found such weird co-existing names in my prior Haskell experience as well, but never wrap my head around why it is so.

Just thought: I'll probably rename things in your commit to `runGhc` to avoid confusion within Hadrian.

izgzhen:
Cool, thank you!

> Just thought: I'll probably rename things in your commit to runGhc to avoid confusion within Hadrian.

I think it is a good idea :)

snowleopard:
@izgzhen I've renamed everything to `runGhc` for consistency (see above commit) and also fixed the build. Unfortunately OSX instance of Travis still fails due to timeout -- I'll see what I can do.

Issue #296, opened 2017-01-06
Migrate warnings settings


snowleopard:
Warnings haven't been translated properly from the old build system to Hadrian, so we can sometimes generate lots and lots of them. I believe this is the reason behind [timing out on our OSX Travis instance](https://travis-ci.org/snowleopard/hadrian/jobs/189395066).

- [x] `-Wall` currently comes after package-specific warning settings coming from `cabal` files in the command line, overriding them.
- [x] General warnings settings: https://github.com/ghc/ghc/blob/master/mk/warnings.mk
- [x] RTS-related warnings: https://github.com/ghc/ghc/blob/master/rts/ghc.mk#L262
- [x] `ghc-cabal`-related warnings: https://github.com/ghc/ghc/blob/master/utils/ghc-cabal/ghc.mk#L55

izgzhen:
What is the difference between "boot" and "install" in `XXX_dist-boot_EXTRA_HC_OPTS` and `XXX_dist-install_EXTRA_HC_OPTS` (https://github.com/ghc/ghc/blob/master/mk/warnings.mk)?

izgzhen:
Oh never mind, I found that probably "boot" means stage1 and "install" means stage2

izgzhen:
But why we don't add `Werror` anywhere in Hadrian now?

izgzhen:
Some attempt:

```diff
diff --git a/src/Settings/Default.hs b/src/Settings/Default.hs
index d242502..f5bf97f 100644
--- a/src/Settings/Default.hs
+++ b/src/Settings/Default.hs
@@ -215,6 +215,12 @@ defaultBuilderArgs = mconcat
     , makeBuilderArgs
     , tarBuilderArgs ]
 
+-- | Disable some warnings in packages we use
+-- | https://github.com/ghc/ghc/blob/master/mk/warnings.mk#L46
+disableWarningArgs :: Args
+disableWarningArgs = stage Stage0 ? builder Ghc ? mconcat
+    [ package transformers ? mconcat [ arg "-fno-warn-unused-matches", arg "-fno-warn-unused-imports" ] ]
+
 -- | All 'Package'-dependent command line arguments.
 defaultPackageArgs :: Args
 defaultPackageArgs = mconcat
@@ -227,4 +233,5 @@ defaultPackageArgs = mconcat
     , haddockPackageArgs
     , integerGmpPackageArgs
     , rtsPackageArgs
-    , runGhcPackageArgs ]
[...]

PR #307, opened 2017-04-28
Disable some warnings

izgzhen:
Part of https://github.com/snowleopard/hadrian/issues/296

### DIFF
--- a/src/Settings/Default.hs
+++ b/src/Settings/Default.hs
@@ -215,6 +215,43 @@ defaultBuilderArgs = mconcat
     , makeBuilderArgs
     , tarBuilderArgs ]
 
+-- | Disable some warnings in packages we use
+-- | https://github.com/ghc/ghc/blob/master/mk/warnings.mk#L46
+disableWarningArgsStage0 :: Args
+disableWarningArgsStage0 = stage Stage0 ? builder Ghc ? mconcat
+    [ package transformers ? append [ "-fno-warn-unused-matches", "-fno-warn-unused-imports" ]
+    , package terminfo ? append [ "-fno-warn-unused-imports" ] ]
+
+disableWarningArgsStage1 :: Args
+disableWarningArgsStage1 = notStage0 ? builder Ghc ? mconcat
+    [ package bytestring ? append [ "-Wno-inline-rule-shadowing" ]
+    , package haddock ? append [ "-Wno-unused-imports", "-Wno-deprecations" ]
+    , package directory ? append [ "-Wno-unused-imports" ]
+    , package binary ? append [ "-Wno-deprecations" ]
+    , package haskeline ? append [ "-Wno-deprecations", "-Wno-unused-imports",
+                                 "-Wno-redundant-constraints",
+                                 "-Wno-simplifiable-class-constraints" ]
+    , package pretty ? append [ "-Wno-unused-imports" ]
+    , package primitive ? append [ "-Wno-unused-imports", "-Wno-deprecations" ]
+    , package terminfo ? append [ "-Wno-unused-imports" ]
+    , package xhtml ? append [ "-Wno-unused-imports", "-Wno-tabs" ]
+    , package transformers ? append [ "-Wno-unused-matches", "-Wno-unused-imports",
+                                    "-Wno-redundant-constraints", "-Wno-orphans" ]
+    , package base ? append [ "-Wno-trustworthy-safe" ]
+    , package ghcPrim ? append [ "-Wno-trustworthy-safe" ]
+    , package win32 ? append [ "-Wno-trustworthy-safe" ] ]
+
+-- GhcLibExtraHcOpts += -Wno-deprecated-flags
+-- GhcBootLibExtraHcOpts += -fno-warn-deprecated-flags
+disableWarningArgsLibs :: Args
+disableWarningArgsLibs = do
+  pkg <- getPackage
+  isLibrary pkg ? builder Ghc ? mconcat
+    [ notStage0 ? arg "-Wno-deprecated-flags"
+    , stage Stage0 ? arg "-fno-warn-deprecated-flags"]
+
+-- TODO: Disable warnings for Windows specifics
+
 -- | All 'Package'-dependent command line arguments.
 defaultPackageArgs :: Args
 defaultPackageArgs = mconcat
@@ -227,4 +264,7 @@ defaultPackageArgs = mconcat
     , haddockPackageArgs
     , integerGmpPackageArgs
     , rtsPackageArgs
-    , runGhcPackageArgs ]
+    , runGhcPackageArgs
+    , disableWarningArgsStage0
+    , disableWarningArgsStage1
+    , disableWarningArgsLibs ]


izgzhen:
Work in progress up to now. A bunch of XXX and TODOs are left in the source. @snowleopard a few comments on them maybe?

src/Settings/Default.hs
```diff
@@ -215,6 +215,44 @@ defaultBuilderArgs = mconcat
...
     , tarBuilderArgs ]
 
+-- | Disable some warnings in packages we use
+-- | https://github.com/ghc/ghc/blob/master/mk/warnings.mk#L46
+disableWarningArgsStage0 :: Args
+disableWarningArgsStage0 = stage Stage0 ? builder Ghc ? mconcat
+    [ package transformers ? mconcat [ arg "-fno-warn-unused-matches", arg "-fno-warn-unused-imports" ]
+    , package terminfo ? mconcat [ arg "-fno-warn-unused-imports" ] ]
+
+-- XXX: How to express Stage1 and upper?
```
snowleopard: There is `notStage0` predicate which I typically use for this.

src/Settings/Default.hs
```diff
@@ -215,6 +215,44 @@ defaultBuilderArgs = mconcat
...
+-- XXX: What about the commented-out packages?
```
snowleopard: There are a few packages which are not built by the Make build system by default and which I therefore never tried to support. I think I would prefer to leave them out for now for two reasons:
* Our priority is to build GHC and the default set of packages as fast as we can.
* Once Hadrian is in the tree, adding a new package (e.g. `vector`) should be a relatively straightforward and well-defined task, which could be handled in a separate GHC ticket if/when needed.

src/Settings/Default.hs
```diff
@@ -215,6 +215,44 @@ defaultBuilderArgs = mconcat
...
+    , package primitive ? mconcat [ arg "-Wno-unused-imports", arg "-Wno-deprecations" ]
+--    , package random ? mconcat [ arg "-Wno-redundant-constraints" ]
+    , package terminfo ? mconcat [ arg "-Wno-unused-imports" ]
+    , package xhtml ? mconcat [ arg "-Wno-unused-imports", arg "-Wno-tabs" ]
+--    , package dphBase ? mconcat [ arg "-Wwarn" ]
+--    , package dphPrimInterface ? mconcat [ arg "-Wwarn"]
+--    , package dphPrimSeq ? mconcat [ arg "-Wwarn"]
+--    , package dphPrimPar ? mconcat [ arg "-Wwarn"]
+--    , package dphLiftedCommon ? mconcat [ arg "-Wwarn"]
+    , package transformers ? mconcat [ arg "-Wno-unused-matches", arg "-Wno-unused-imports", arg "-Wno-redundant-constraints", arg "-Wno-orphans" ]
```
snowleopard: I usually prefer to fit lines into 80 (or near) characters, so it's easier to work with in different situations. Right, for example, I don't see the end of this line.

src/Settings/Default.hs
```diff
@@ -215,6 +215,44 @@ defaultBuilderArgs = mconcat
     , makeBuilderArgs
     , tarBuilderArgs ]
 
+-- | Disable some warnings in packages we use
+-- | https://github.com/ghc/ghc/blob/master/mk/warnings.mk#L46
+disableWarningArgsStage0 :: Args
+disableWarningArgsStage0 = stage Stage0 ? builder Ghc ? mconcat
+    [ package transformers ? mconcat [ arg "-fno-warn-unused-matches", arg "-fno-warn-unused-imports" ]
```
snowleopard: `append` adds a list of arguments, so you don't need to do `mconcat`.

snowleopard: (Maybe not the best name and could be renamed to `args`.)

src/Settings/Default.hs
```diff
@@ -215,6 +215,44 @@ defaultBuilderArgs = mconcat
...
+--    , package dphPrimSeq ? mconcat [ arg "-Wwarn"]
+--    , package dphPrimPar ? mconcat [ arg "-Wwarn"]
+--    , package dphLiftedCommon ? mconcat [ arg "-Wwarn"]
+    , package transformers ? mconcat [ arg "-Wno-unused-matches", arg "-Wno-unused-imports", arg "-Wno-redundant-constraints", arg "-Wno-orphans" ]
+    , package base ? mconcat [ arg "-Wno-trustworthy-safe" ]
+    , package ghcPrim ? mconcat [ arg "-Wno-trustworthy-safe" ]
+    , package win32 ? mconcat [ arg "-Wno-trustworthy-safe" ] ]
+
+-- TODO: Disable warnings for the following
+-- GhcLibExtraHcOpts += -Wno-deprecated-flags
```
snowleopard: I usually handle these as follows:
```haskell
    pkg <- getPackage
    mconcat [ isLibrary pkg ? ... ]
```
But I could create a combinator to make this terser, e.g. `libraryPackage ?`. Would you like one?

src/Settings/Default.hs
```diff
@@ -215,6 +215,44 @@ defaultBuilderArgs = mconcat
...
+-- GhcBootLibExtraHcOpts += -fno-warn-deprecated-flags
```
snowleopard: With the above, this would be `stage0 ? libraryPackage ? arg "-fno-warn-deprecated-flags"`.

snowleopard:
@izgzhen Many thanks! See my comments above.

src/Settings/Default.hs
```diff
@@ -215,6 +215,52 @@ defaultBuilderArgs = mconcat
...
     , tarBuilderArgs ]
 
+-- | Disable some warnings in packages we use
+-- | https://github.com/ghc/ghc/blob/master/mk/warnings.mk#L46
+disableWarningArgsStage0 :: Args
+disableWarningArgsStage0 = stage Stage0 ? builder Ghc ? mconcat
+    [ package transformers ? append [ "-fno-warn-unused-matches", "-fno-warn-unused-imports" ]
+    , package terminfo ? append [ "-fno-warn-unused-imports" ] ]
+
+-- NOTE(izgzhen): The commented-out packages are NOT built by default with make either,
```
izgzhen: @snowleopard I just leave them commented here, is that okay?

src/Settings/Default.hs
```diff
@@ -215,6 +215,52 @@ defaultBuilderArgs = mconcat
...
+-- so they are only here for consistency
+disableWarningArgsStage1 :: Args
+disableWarningArgsStage1 = notStage0 ? builder Ghc ? mconcat
+    [ package bytestring ? append [ "-Wno-inline-rule-shadowing" ]
```
izgzhen: I found the name `args` is already used at another place ... `append` looks not too bad to me

src/Settings/Default.hs
```diff
@@ -215,6 +215,52 @@ defaultBuilderArgs = mconcat
...
+--    , package dphLiftedCommon ? append [ "-Wwarn"]
+    , package transformers ? append [ "-Wno-unused-matches", "-Wno-unused-imports",
+                                    "-Wno-redundant-constraints", "-Wno-orphans" ]
+    , package base ? append [ "-Wno-trustworthy-safe" ]
+    , package ghcPrim ? append [ "-Wno-trustworthy-safe" ]
+    , package win32 ? append [ "-Wno-trustworthy-safe" ] ]
+
+-- GhcLibExtraHcOpts += -Wno-deprecated-flags
+-- GhcBootLibExtraHcOpts += -fno-warn-deprecated-flags
+disableWarningArgsLibs :: Args
```
izgzhen: @snowleopard implemented this part

izgzhen: not sure whether it would be profitable to have another combinator, since I found `isLibrary`'s implementation quite simple

src/Settings/Default.hs
snowleopard: > So does it mean that GhcLibExtraHcOpts will be used in the compilation of every package?

I think `GhcLibExtraHcOpts` is added to every library package in Make. Isn't this the case?

> A combinator would be nice -- though I think I can probably try to help making one :)

Sure, go ahead and send a PR for it separately :-) If you do this, it would be nice if you also find other spots in Hadrian which could be simplified thanks to it.

izgzhen:
BTW, this is my first time using GitHub's builtin review system and ... it is still bad at handling `rebase` and forced push ... before that I have been using https://reviewable.io in Servo, which is better at handling this, though a bit heavyweight.

src/Settings/Default.hs
snowleopard: I would prefer not to have them at all for now, because I'm not sure how up-to-date these settings are in Make. These packages are not built by default and the settings could be outdated.

snowleopard:
I've never used any external code review tools on GitHub, but we could give it a try if you like.

src/Settings/Default.hs
snowleopard: Don't worry about the extra combinator then.

src/Settings/Default.hs
snowleopard: These packages are currently not considered part of GHC, but are still left in the tree (because at some point they shipped with GHC I believe). They may be gone in future. Or may be added back into packages built by default, in which case we'll add them to Hadrian.

This is not related to the build flavour.

izgzhen:
Deleted the commented packages.

Regarding the review tool, I can still stay with the built-in one for now. Reviewable may be useful if we have too many changes to track in future.

snowleopard:
Shall I merge this? As far as I'm concerned the current version is fine. (But I might do a minor revision later).

Let's see whether the CI bots are happy.

I guess the next step is enable `-Werror` and see if the build still works?

izgzhen:
> Shall I merge this?

Yes, I'm ready.

> I guess the next step is enable -Werror and see if the build still works?

Sure, but maybe in another issue?

snowleopard:
Merged now.

> Sure, but maybe in another issue?

I think this should be the same issue, but a different PR.

izgzhen:
Linux CI failed instantly ...

```
/home/travis/build/snowleopard/hadrian/ghc/hadrian/src/Rules/Cabal.hs:7:8:
    Could not find module ‘Distribution.Types.CondTree’
    Use -v to see a list of the files searched for.
/home/travis/build/snowleopard/hadrian/ghc/hadrian/src/Rules/Cabal.hs:8:8:
    Could not find module ‘Distribution.Types.Dependency’
    Use -v to see a list of the files searched for.
```

snowleopard:
Ah, looks like they changed the in-tree Cabal library again, which Hadrian depends on.

This happens quite regularly unfortunately.

snowleopard:
Usually, the fix is pretty simple (e.g. add a new import).

I will have time to look at this in 4-5 hours. Feel free to give it a try too.

izgzhen:
Weird enough, I updated the GHC etc. and compiled fine.

snowleopard:
Indeed! I restarted the bots and they are happy now.

PR #313, opened 2017-05-15
Add copyFileUntracked

izgzhen:


izgzhen:
Should I fix the bit in `Libffi`? `copyFile` is somewhat unreasonable, but not really affecting the last issue.

snowleopard:
Thanks!

Yes, let's fix the missing build result in the `libffi` rule and switch to `copyFileUntracked`.

snowleopard:
I mean, we could probably do it within this PR.

src/Util.hs
```diff
@@ -98,6 +98,14 @@ copyFile source target = do
     putProgressInfo $ renderAction "Copy file" source target
     copyFileChanged source target
 
+-- Same as copyFile, but not tracking the source as a build dependency
+copyFileUntracked :: FilePath -> FilePath -> Action ()
+copyFileUntracked source target = do
+    let dir = takeDirectory target
+    unlessM (liftIO $ IO.doesDirectoryExist dir) $ createDirectory dir
```
ndmitchell: Why not just createDirectoryIfMissing? That also works recursively, which is usually what you want.

snowleopard: Aha, looks like it fits indeed.

Are you suggesting to replace the line with `createDirectoryIfMissing True dir`?

ndmitchell: Yes. That's what Shake itself uses.

snowleopard: @izgzhen Let's use `createDirectoryIfMissing True dir` indeed. Please also amend the `copyFile` function.

izgzhen:
I am ready on this, any more comments?

snowleopard:
Many thanks @izgzhen! Merged now.

PR #319, opened 2017-06-05
Fix CABAL_VERSION argument in building ghc-cabal

izgzhen:
Part of https://github.com/snowleopard/hadrian/pull/312, which can fix the version number in `setup-config`'s header section

### DIFF
--- a/src/Settings/Packages/GhcCabal.hs
+++ b/src/Settings/Packages/GhcCabal.hs
@@ -1,9 +1,17 @@
 module Settings.Packages.GhcCabal (ghcCabalPackageArgs) where
 
+import Distribution.PackageDescription.Parse
+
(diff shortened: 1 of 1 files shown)


snowleopard:
Thanks! Looks good to me.

Just one note: we could probably also get rid of the hard-coded list `cabalDeps`, since we are now parsing the `.cabal` file anyway. Do you think you could do that?

Anyway, this should probably be done in a separate PR. Shall I merge this?

izgzhen:
> Just one note: we could probably also get rid of the hard-coded list cabalDeps, since we are now parsing the .cabal file anyway. Do you think you could do that?

Sure, I will do it another time.

> Shall I merge this?

Yes, thanks.

PR #320, opened 2017-06-06
Compute cabalDeps in GhcCabal build

izgzhen:


src/Oracles/Dependencies.hs
```diff
@@ -47,6 +48,17 @@ contextDependencies context@Context {..} = do
     pkgs <- sort <$> interpretInContext (pkgContext package) getPackages
     return . map pkgContext $ intersectOrd (compare . pkgNameString) pkgs deps
 
+-- | Given a `Package`, this `Action` looks up its package dependencies
+-- 'Settings.Paths.packageDependencies' using 'packageDependenciesOracle'
+-- The context will be the vanilla context with stage equal to 1
+pkgDependencies :: Package -> Action [Package]
+pkgDependencies pkg = do
```
snowleopard: It looks like `pkgDependencies = fmap Context.package . contextDependencies . vanillaContext Stage1`.

Is it so? At least the rest of the code seems to be duplicated.

izgzhen: You are right on this point -- since my initiation is to export a more handy function, so I will rewrite it based on `contextDependencies` instead. thanks

snowleopard:
Nice PR! Is it ready for merge?

izgzhen:
Yes

PR #321, opened 2017-06-07
Add binary wrappers for hp2ps, hpc, hsc2hs

izgzhen:


snowleopard:
Looks good to me. There is are a few spots with code duplication, but not sure it is worth abstracting over.

Is this ready for merge?

izgzhen:
Yes, I am ready.

> There is are a few spots with code duplication, but not sure it is worth abstracting over.

I don't think we need to further abstract this, which might only make it more confusing.

snowleopard:
Agreed. And merged.

Issue #323, opened 2017-06-11
Lint error when building documentation


izgzhen:
Output is like below when running `./build.stack.sh --flavour=quickest --integer-simple --haddock` on OS X:

```
Lint checking error - 11 values have changed since being depended upon:
  Key:  /Users/zz/Repos/ghc2/_build/stage1/libraries/ghc-prim/doc/html/ghc-prim/src/hscolour.css
  Old:  File {mod=0xED41D0A0,size=0xF3,digest=0x8E3F54C4}
  New:  File {mod=0xED8DCDC0,size=0xF3,digest=0x8E3F54C4}
       
  Key:  /Users/zz/Repos/ghc2/_build/stage1/libraries/ghc-prim/doc/html/ghc-prim/src/GHC-Types.html
  Old:  File {mod=0xED41F7B0,size=0xCA58,digest=0xE04D4498}
  New:  File {mod=0xED8DF4D0,size=0xCA58,digest=0xE04D4498}
       
  Key:  /Users/zz/Repos/ghc2/_build/stage1/libraries/ghc-prim/doc/html/ghc-prim/src/GHC-Tuple.html
  Old:  File {mod=0xED41F7B0,size=0x46F90,digest=0x4BD977EF}
  New:  File {mod=0xED8DF4D0,size=0x46F90,digest=0x4BD977EF}
       
  Key:  /Users/zz/Repos/ghc2/_build/stage1/libraries/ghc-prim/doc/html/ghc-prim/src/GHC-PrimopWrappers.html
  Old:  File {mod=0xED41F7B0,size=0x847E0,digest=0xBB92D9EA}
  New:  File {mod=0xED8DCDC0,size=0x847E0,digest=0xBB92D9EA}
       
  Key:  /Users/zz/Repos/ghc2/_build/stage1/libraries/ghc-prim/doc/html/ghc-prim/src/GHC-Prim.html
  Old:  File {mod=0xED41F7B0,size=0x144BC5,digest=0xDB2E8E4E}
  New:  File {mod=0xED8DF4D0,size=0x144BC5,digest=0xDB2E8E4E}
       
  Key:  /Users/zz/Repos/ghc2/_build/stage1/libraries/ghc-prim/doc/html/ghc-prim/src/GHC-Magic.html
  Old:  File {mod=0xED41D0A0,size=0x3AB6,digest=0xE190FDBE}
  New:  File {mod=0xED8DCDC0,size=0x3AB6,digest=0xE190FDBE}
       
[...]

Issue #318, opened 2017-06-04
ghc-cabal copy assumes library artifacts in build subpath


izgzhen:
Related `ghc-cabal` code: https://github.com/ghc/packages-Cabal/blob/master/Cabal/Distribution/Simple/Configure.hs#L366

This creates problems for install rules.

We discussed a bit in https://github.com/snowleopard/hadrian/pull/312, but I don't think we have consensus on it yet. Here are several possible ways to fix:

1. Revert the change and build stuff in `build` subpath, as what the old system does.
2. Patch the Cabal source in our build system
3. Add rules for copying or symbolically linking built stuff to `build` (the later one will probably not work on Windows).

The first one is difficult. The second one is a bit hack-ish and will be influenced by upstream code changes. The third one is hash-ish as well, more robust against upstream changes, but a bit wasteful if we have to copy over stuff .....

Overall, I personally favor 2 and propose to do it in a separate PR.

snowleopard:
I think we also discussed another option:

4. Implement the `ghc-cabal copy` functionality in Hadrian.

Presumably, we just need to copy all package objects and the library? It's easy to get a list of all build artefacts that Hadrian produces for a given package. For example, you can get all Haskell objects using this: https://github.com/snowleopard/hadrian/blob/master/src/Oracles/ModuleFiles.hs#L84-L87.

I think this may be the best way to go in the long term. In this way we also have the ability to track files that are being copied, so we'll rebuild outdated files if need be. 

@izgzhen Can you list all the files that need to be copied for a given package, say for `libraries/array`?

izgzhen:
For `libraries/array`:

```
├── Data
│   ├── Array
│   │   ├── Base.dyn_hi
│   │   ├── Base.hi
│   │   ├── IArray.dyn_hi
│   │   ├── IArray.hi
│   │   ├── IO
│   │   │   ├── Internals.dyn_hi
│   │   │   ├── Internals.hi
│   │   │   ├── Safe.dyn_hi
│   │   │   └── Safe.hi
│   │   ├── IO.dyn_hi
│   │   ├── IO.hi
│   │   ├── MArray
│   │   │   ├── Safe.dyn_hi
│   │   │   └── Safe.hi
│   │   ├── MArray.dyn_hi
│   │   ├── MArray.hi
│   │   ├── ST
│   │   │   ├── Safe.dyn_hi
│   │   │   └── Safe.hi
│   │   ├── ST.dyn_hi
[...]

PR #324, opened 2017-06-11
Fix documentation rules

izgzhen:


izgzhen:
And quite a lot of warning like these:

```
    lintCoreExpr
Warning: CorePrep: could not find link destinations for:
    CpeBody wrapBinds Floats cpeBodyNF
Warning: ByteCodeInstr: could not find link destinations for:
    LocalLabel
Warning: Linker: could not find link destinations for:
    LinkerUnitId linkSomeBCOs
Warning: TcIface: could not find link destinations for:
    mergeIfaceDecl
Warning: TcHsType: could not find link destinations for:
    TcTyMode tc_infer_hs_type
Warning: TcTypeable: could not find link destinations for:
    KindRepM
Warning: TcBackpack: could not find link destinations for:
    requirementMerges
Warning: DsMonad: could not find link destinations for:
    P Put VM SetM State checkLoadDAP
Warning: DsForeign: could not find link destinations for:
    Binding
Warning: Check: could not find link destinations for:
    pmcheck pmcheckGuards pmcheckHd
Warning: Match: could not find link destinations for:
    MatchId
Warning: TcTyClsDecls: could not find link destinations for:
    FamTyConShape
Warning: TcRnDriver: could not find link destinations for:
    requirementMerges
Warning: Vectorise.Builtins.Initialise: could not find link destinations for:
    externalVar
Warning: Vectorise.Generic.PAMethods: could not find link destinati
```

But generally speaking it works on my Mac now.

ndmitchell:
Don't worry about the warnings. Haddock spits out a lot of garbage which has zero value.

izgzhen:
@snowleopard Can you look into that when available?

snowleopard:
@izgzhen I thought this was blocked by #323. Have you found a fix for the lint error?

izgzhen:
@ndmitchell I can't turn off lint checking of those html files by letting `shakeLintIgnore = ["_build//*.html"]`. Can you see why it is so?

izgzhen:
I can confirm that `haddock` might modify the built files after creating it in the first place, thus causing the lint error. However, trying `shakeLintIgnore` or `trackAllow` doesn't quite work.

Since this PR doesn't really break anything, I think we can merge this fix first, then concentrating in the either the problem of `haddock` or `shake` afterwards. What do you think? @snowleopard

izgzhen:
Here is a gist https://gist.github.com/izgzhen/74dd1672727ce19e2e3ca58b9299b982 that demos the problem.

src/Rules/Documentation.hs
```diff
@@ -34,6 +37,12 @@ buildPackageDocumentation context@Context {..} =
             -- HsColour sources
             -- TODO: what is the output of GhcCabalHsColour?
             whenM (isSpecified HsColour) $ do
+                top <- interpretInContext context getTopDirectory
+                let installDistDir = top -/- buildPath context
```
snowleopard: What about doing install-related copying in install rules? The current solution seems to always do the copying, which is suboptimal.

snowleopard:
@izgzhen See a new comment above. I think if you move the copying part into the install rule, then not only this will be faster for the common case (building docs without installation), it may also mitigate the current lint error, because it will only happen when docs are rebuilt during installation. 

What do you think?

izgzhen:
> it may also mitigate the current lint error, because it will only happen when docs are rebuilt during installation.

I don't get why copying will be relevant to the lint error, can you elaborate on it? (I will try mitigating this line at the same time as well)

snowleopard:
> I don't get why copying will be relevant to the lint error, can you elaborate on it?

Sure. As far as I can see, our build rules never list `*.html` files as build rule targets. We only explicitly track `*.haddock` files, partly for efficiency, partly because listing all `*.html` files would be a pain.

On the other hand, the lint error talks about `*.css` and `*.html` files! The reason, I think, is that they are tracked during copying. If this step is moved to the installation rule, we shouldn't see the lint error during the usual build. 

Does this make sense?

izgzhen:
> The reason, I think, is that they are tracked during copying

Oh I got it. Ideally, I am thinking of the dependency graph as a directed graph where all zero-indeg nodes are source code, and all actions are the edge that produces output _functionally_. But, well, it looks like there are some black-box action (external builders) that produces more than specified, and also some black-box, like `copyFile`, works somewhat non-deterministically.

izgzhen:
Updated the code, but the `build --haddock` is still not perfect: after one build _finished_, if I repeated the `build --haddock`, then it will start to rebuild (not always all) documentations again. It seems diverging.

snowleopard:
> and also some black-box, like `copyFile`, works somewhat non-deterministically.

I'm pretty sure `copyFile` is fine. It just adds a dependency on `*.html` files that are rebuilt multiple times by Haddock, which is the core of the issue, as spotted by Shake's lint. 

The fact that `shakeLintIgnore` doesn't work should probably be recorded in a Shake issue.

snowleopard:
And while we are reporting bugs, we might as well raise a GHC ticket about Haddock being so naughty :)

src/Rules/Install.hs
```diff
@@ -193,6 +193,20 @@ installPackages = do
...
                 need [ ghcCabalInplace ]
 
+                let cabalFile = pkgCabalFile pkg
+                -- HsColour sources
+                -- QUESTION: what is the output of GhcCabalHsColour?
+                whenM (isSpecified HsColour) $ do
+                    top <- interpretInContext context getTopDirectory
+                    let installDistDir = top -/- buildPath context
+                    -- HACK: copy stuff back to the place favored by ghc-cabal
+                    quietly $ copyDirectoryContents (Not excluded)
```
snowleopard: I don't think copying should be conditional on `HsColour`. The latter seems to be optional for documentation.

snowleopard:
> Updated the code, but the build --haddock is still not perfect: after one build finished, if I repeated the build --haddock, then it will start to rebuild (not always all) documentations again. It seems diverging.

But you are still doing some copying during the normal build -- was it left accidentally?

src/Rules/Install.hs
izgzhen: copying is to make sure that build $ Target context GhcCabalHsColour [cabalFile] [] can run well. I don't know if coloring would be optional for docs or not, but if we want color, we need copying as a temp hack I guess.

izgzhen:
> The fact that shakeLintIgnore doesn't work should probably be recorded in a Shake issue.

https://github.com/ndmitchell/shake/issues/531

> But you are still doing some copying during the normal build -- was it left accidentally?

Indeed! I thought my Emacs has already saved that but (shrugging). I've updated the code.

src/Rules/Install.hs
snowleopard: Wait, I'm confused now.

We need to run `build $ Target context GhcCabalHsColour [cabalFile] []` during normal builds. Why are you moving this to the install rule? 

I would expect to see `GhcCabalHsColour` to be left in the documentation rule, while the install rule should simply copy all resulting files. Isn't this sufficient?

snowleopard:
@izgzhen Thanks! I'm still not sure the current version is what we are after -- I've left one more comment.

src/Rules/Install.hs
izgzhen: `GhcCabalHsColour` will invoke `ghc-cabal`, which assumes the path (`build`). So this is basically the same problem as #318.

So, copying has to be done before `GhcCabalHsColour`, or `installPackages` (where we use `ghc-cabal copy`.

snowleopard: Oh, right. I didn't realise `GhcCabalHsColour` was broken in the normal build. It may be worth checking what it does and whether we can just run `hscolour` on resulting files ourselves.

snowleopard:
@izgzhen Right, with the last bit clarified, I've merged this. Thanks for your patience!

PR #325, opened 2017-06-14
Build dynamic libs

izgzhen:
Solve #4

### DIFF
--- a/src/Rules.hs
+++ b/src/Rules.hs
@@ -64,6 +64,10 @@ packageRules = do
         [ Rules.Compile.compilePackage readPackageDb
         , Rules.Library.buildPackageLibrary ]
 
+    let dynamicContexts = liftM3 Context [Stage1 ..] knownPackages [dynamic]
+
+    forM_ dynamicContexts Rules.Library.buildDynamicLib
+
     forM_ programContexts $ Rules.Program.buildProgram readPackageDb
 
     forM_ vanillaContexts $ mconcat
--- a/src/Rules/Library.hs
+++ b/src/Rules/Library.hs
@@ -1,4 +1,7 @@
-module Rules.Library (buildPackageLibrary, buildPackageGhciLibrary) where
+module Rules.Library (
+    buildPackageLibrary, buildPackageGhciLibrary,
+    buildDynamicLib
+) where
 
 import Data.Char
 import qualified System.Directory as IO
@@ -10,33 +13,62 @@ import Flavour
 import GHC
 import Oracles.ModuleFiles
 import Oracles.PackageData
+import Oracles.Dependencies (contextDependencies)
 import Settings
 import Settings.Path
 import Target
 import UserSettings
 import Util
 
+getLibraryObjs :: Context -> Action [FilePath]
+getLibraryObjs context@Context{..} = do
+    hsObjs   <- hsObjects    context
+    noHsObjs <- nonHsObjects context
+
+    -- This will create split objects if required (we don't track them
+    -- explicitly as this would needlessly bloat the Shake database).
+    need $ noHsObjs ++ hsObjs
+
+    split <- interpretInContext context $ splitObjects flavour
+    let getSplitObjs = concatForM hsObjs $ \obj -> do
+             let dir = dropExtension obj ++ "_" ++ osuf way ++ "_split"
+             contents <- liftIO $ IO.getDirectoryContents dir
+             return . map (dir -/-) $ filter (not . all (== '.')) contents
+
+    (noHsObjs ++) <$> if split then getSplitObjs else return hsObjs
+
+buildDynamicLib :: Context -> Rules ()
+buildDynamicLib context@Context{..} = do
+    -- macOS
+    matchGhcVersionedFilePath libPrefix "dylib" ?> buildDynamicLibUNIX
+    -- Linux
+    matchGhcVersionedFilePath libPrefix "so" ?> buildDynamicLibUNIX
+    -- TODO: Windows
+  where
+    path       = buildPath context
+    libPrefix  = path -/- "libHS" ++ pkgNameString package
+
+    buildDynamicLibUNIX so = do
+        deps <- contextDependencies context
+
+        forM_ deps $ \dep -> do
+            lib <- pkgLibraryFile dep
+            need [lib]
+
+        removeFile so
+
+        objs <- getLibraryObjs context
+
+        build $ Target context (Ghc LinkHs stage) objs [so]
+
 buildPackageLibrary :: Context -> Rules ()
 buildPackageLibrary context@Context {..} = do
     let path       = buildPath context
         libPrefix  = path -/- "libHS" ++ pkgNameString package
-    -- TODO: handle dynamic libraries
     matchVersionedFilePath libPrefix (waySuffix way <.> "a") ?> \a -> do
         removeFile a
-        hsObjs   <- hsObjects    context
-        noHsObjs <- nonHsObjects context
 
-        -- This will create split objects if required (we don't track them
-        -- explicitly as this would needlessly bloat the Shake database).
-        need $ noHsObjs ++ hsObjs
-
-        split <- interpretInContext context $ splitObjects flavour
-        let getSplitObjs = concatForM hsObjs $ \obj -> do
-                let dir = dropExtension obj ++ "_" ++ osuf way ++ "_split"
-                contents <- liftIO $ IO.getDirectoryContents dir
-                return . map (dir -/-) $ filter (not . all (== '.')) contents
-
-        objs <- (noHsObjs ++) <$> if split then getSplitObjs else return hsObjs
+        objs <- getLibraryObjs context
 
         asuf <- libsuf way
         let isLib0 = ("//*-0" ++ asuf) ?== a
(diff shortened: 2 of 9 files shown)


izgzhen:
It can successful compile a hello-world dynamically now:

```
inplace/bin/ghc-stage2 -dynamic --make ~/Desktop/Playground/Main.hs
```

snowleopard:
@izgzhen Whoa, hang on, you're moving too fast! You were supposed to start yesterday, but you already resolved two main issues :D

I'll do my best to review both PRs tomorrow -- today's is a bit busy.

izgzhen:
> I'll do my best to review both PRs tomorrow -- today's is a bit busy.

:-) Take your time, thanks

src/Settings/Flavours/Quickest.hs
```diff
@@ -2,13 +2,16 @@ module Settings.Flavours.Quickest (quickestFlavour) where
...
 import {-# SOURCE #-} Settings.Default
 
 quickestFlavour :: Flavour
 quickestFlavour = defaultFlavour
     { name        = "quickest"
     , args        = defaultBuilderArgs <> quickestArgs <> defaultPackageArgs
-    , libraryWays = append [vanilla]
+    , libraryWays = mconcat
+                    [ append [vanilla]
+                    , notStage0 ? platformSupportsSharedLibs ? append [dynamic] ]
```
snowleopard: The `quickest` flavour is supposed to be the quickest one that produces a working GHC, to be used by CI bots and avoid time limits, so I suggest we stick to pure `vanilla`.

src/Rules.hs
```diff
@@ -61,6 +61,10 @@ packageRules = do
         [ Rules.Compile.compilePackage readPackageDb
         , Rules.Library.buildPackageLibrary ]
 
+    let dynContexts = liftM3 Context [Stage1 ..] knownPackages [dynamic]
```
snowleopard: I am not a fan of using contractions. Can we use `dynamic` throughout the code instead of `dyn`? This will be more consistent with the rest of Hadrian.

src/Settings/Builders/Ghc.hs
```diff
@@ -28,7 +29,9 @@ ghcLinkArgs = builder (Ghc LinkHs) ? do
                    buildInfo <- lift $ readFileLines gmpBuildInfoPath
                    return $ concatMap (words . strip) buildInfo
                else return []
-    mconcat [ arg "-no-auto-link-packages"
+    mconcat [ (Dynamic `wayUnit` way) ?
+              append [ "-shared", "-dynamic", "-dynload", "deploy" ]
+            , arg "-no-auto-link-packages"
```
snowleopard: I'm confused why `-no-auto-link-packages` is not included in the above `append`.

snowleopard:
Nice, looks a lot simpler than I imagined :-) 

I've added a few minor comments.

Does this fully solve #4? I remember varous mysterious settings like `DynamicGhcPrograms`, `DYNAMIC_TOO`, etc. Also, what about Windows? Wasn't there a recent GHC patch that implemented dynamic way on Windows?

snowleopard:
By the way, Linux CI fails.

src/Settings/Builders/Ghc.hs
izgzhen: Isn't it still on the line after the `append`? Both static and dynamic needs that flag I think.

src/Settings/Flavours/Quickest.hs
izgzhen: But this is not consistent with the behavior of `quickest` flavor specified in `mk/build.mk`. I remember that dynamic libs will be produced in that one. Will it be a surprise if we redefine this?

src/Settings/Builders/Ghc.hs
snowleopard: Ah, my bad! Please disregard this comment.

src/Settings/Flavours/Quickest.hs
snowleopard: The difference between `quick` and `quickest` in the Make build system seemed too marginal to me, so I thought it would be better to deviate from the existing definition. The `quickest` flavour should be the one that can't be beaten in terms of speed or else why name it this way :)

At the moment, `quickest` in Hadrian is already different, so I suggest not to try to change this in this PR. This discussion deserves a separate issue.

izgzhen: That makes sense.

src/Rules.hs
izgzhen: Sounds good

izgzhen:
> By the way, Linux CI fails.

Ah, my bad, `dylib` is a mac OS thing, the Linux uses `so`

izgzhen:
> Also, what about Windows? Wasn't there a recent GHC patch that implemented dynamic way on Windows?

Including Windows, I will work on the platform thing soon

izgzhen:
> Including Windows, I will work on the platform thing soon

Well, I found that it is hard to test Windows-version on my machine for now. Can I leave it for future?

snowleopard:
> Well, I found that it is hard to test Windows-version on my machine for now. Can I leave it for future?

Yes, sure.

izgzhen:
> I remember varous mysterious settings like DynamicGhcPrograms, DYNAMIC_TOO

Relevant wiki:

- https://ghc.haskell.org/trac/ghc/wiki/DynamicGhcPrograms
- https://ghc.haskell.org/trac/ghc/wiki/DynamicLinkingDebate
- https://ghc.haskell.org/trac/ghc/wiki/DynamicByDefault

Looks like this issue is around `GHCi`, a more specialized problem. I believe it is one of the dynamic way milestone tasks, but I prefer to submit a separate issue and PR for it.

Then how should I define this PR? "Build dynamic libs for GHC executable compilation"?

izgzhen:
BTW, I should have fixed the Linux dynamic lib. The weird bit is, for object like `rts/Adjust.dyn_o`, the old system use `ghc-stage2` to build it (the source is C and it works!), while our system will use `Cc` builder. So the ultimate `.dyn_o` is built by different commands, comparing my patch and the reference.

snowleopard:
> Looks like this issue is around GHCi, a more specialized problem. I believe it is one of the dynamic way milestone tasks, but I prefer to submit a separate issue and PR for it.

Yes, sounds good.

> Then how should I define this PR? "Build dynamic libs for GHC executable compilation"?

Works for me!

izgzhen:
I pushed the changes to follow the upstream change of dropping `hoopl` dependency in this PR as well. Should I open another PR for it?

Also, it should really *work* now, but both Linux and OS X CI timeout due to the lengthy process of building dynamic libs.

snowleopard:
@izgzhen I'd prefer a separate PR for dropping `hoopl`.

> but both Linux and OS X CI timeout due to the lengthy process of building dynamic libs

So, let's remove `dynamic` from `quickest`. We have a separate issue where we'll keep discussing this.

izgzhen:
> So, let's remove dynamic from quickest. We have a separate issue where we'll keep discussing this.

Done (`quick` still keeps it).

snowleopard:
@izgzhen Thanks! Shall I merge this one too?

src/Settings/Builders/Cc.hs
```diff
@@ -3,12 +3,17 @@ module Settings.Builders.Cc (ccBuilderArgs) where
...
+ccBuilderArgs = do
+  way <- getWay
+  builder Cc ? mconcat
     [ append =<< getPkgDataList CcArgs
     , argSettingList . ConfCcArgs =<< getStage
     , cIncludeArgs
 
     , builder (Cc CompileC) ? mconcat [ arg "-Werror"
+                                      , (Dynamic `wayUnit` way) ?
+				        mconcat [ arg "-fPIC"
```
snowleopard: Should this be `append ["-fPIC", "-DDYNAMIC"]`?

src/Rules/Library.hs
```diff
@@ -10,17 +13,53 @@ import Flavour
...
+            lib <- pkgLibraryFile dep
+            need [lib]
+
+        removeFile so
+        hsObjs   <- hsObjects    context
+        noHsObjs <- nonHsObjects context
+
+        -- This will create split objects if required (we don't track them
+        -- explicitly as this would needlessly bloat the Shake database).
+        need $ noHsObjs ++ hsObjs
```
snowleopard: Can we get rid of this code duplication?

snowleopard: See `buildPackageLibrary` below, which seems to have a lot of overlap with this build rule.

snowleopard:
I've added a couple of more comments after looking through the code again.

src/Rules/Library.hs
izgzhen: Yes there is code duplication since `buildDynamicLib` is adapted from `buildPackageLibrary`. Before that I am not sure if I should refactor or not, but let's refactor it.

src/Settings/Builders/Cc.hs
izgzhen: Fixed

izgzhen:
> Shall I merge this one too?

I am ready now, you may merge if as you see fit.

snowleopard:
Thanks! Merged.

PR #328, opened 2017-06-25
Drop dependency on hoopl

izgzhen:


snowleopard:
Thanks, looks good! And the OSX CI finished pretty comfortably.

PR #329, opened 2017-06-27
Don't pass gcc warning options to ghc

bgamari:
We would previously pass -Werror=unused-but-set-variable and -Wno-error=inline
to ghc, despite the fact that they are gcc flags.


snowleopard:
Thanks @bgamari! Merged.

PR #331, opened 2017-06-27
Various portability fixes

### LINKED ISSUE #330: Hadrian assumes path of bash
Trying to build with Hadrian on, for instance, FreeBSD falls over at the first gate with,
```
Linking /usr/home/ben/ghc/hadrian/dist-newstyle/build/hadrian-0.1.0.0/build/hadrian/hadrian ...
/-------------------------------------------------------\
| Run Configure "." (stage = Stage0, package = ghc)     |
|      input: hadrian/cfg/system.config.in (and 2 more) |
|  => output: hadrian/cfg/system.config (and 2 more)    |
\-------------------------------------------------------/
exec: /bin/bash: not found
shakeArgsWith   0.000s    0%                           
Function shake  0.320s   59%  =========================
Database read   0.000s    0%                           
With database   0.000s    0%                           
Running rules   0.220s   40%  =================        
Total           0.542s  100%                           
* OracleQ (ConfigKey "host-os")
* hadrian/cfg/system.config
* hadrian/cfg/system.config settings mk/config.h
Command: /bin/sh -c 'sh configure'
Current directory: /usr/home/ben/ghc
Exit code: 127
Stderr:
exec: /bin/bash: not found
)
```
It's not clear to me where this assumption is coming from.

bgamari:
These are the result of an attempt to build GHC with Hadrian on FreeBSD on amd64. This resolves #330.

### DIFF
--- a/src/Oracles/Path.hs
+++ b/src/Oracles/Path.hs
@@ -1,6 +1,6 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 module Oracles.Path (
-    topDirectory, getTopDirectory, systemBuilderPath, pathOracle
+    topDirectory, getTopDirectory, systemBuilderPath, pathOracle, bashPath
     ) where
 
 import Control.Monad.Trans.Reader
@@ -28,7 +28,7 @@ systemBuilderPath builder = case builder of
     Cc  _  Stage0   -> fromKey "system-cc"
     Cc  _  _        -> fromKey "cc"
     -- We can't ask configure for the path to configure!
-    Configure _     -> return "bash configure"
+    Configure _     -> return "sh configure"
     Ghc _  Stage0   -> fromKey "system-ghc"
     GhcPkg _ Stage0 -> fromKey "system-ghc-pkg"
     Happy           -> fromKey "happy"
@@ -54,6 +54,10 @@ systemBuilderPath builder = case builder of
                 ++ quote key ++ " is not specified in system.config file."
             return "" -- TODO: Use a safe interface.
         else fixAbsolutePathOnWindows =<< lookupInPath path
+
+-- | Lookup the path to the @bash@ interpreter.
+bashPath :: Action FilePath
+bashPath = lookupInPath "bash"
 
 -- | Lookup an executable in @PATH@.
 lookupInPath :: FilePath -> Action FilePath
@@ -95,4 +99,3 @@ pathOracle = do
         path <- unifyPath <$> unpack <$> liftIO (findExecutable name)
         putLoud $ "Executable found: " ++ name ++ " => " ++ path
         return path
-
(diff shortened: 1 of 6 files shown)


src/Oracles/Path.hs
```diff
@@ -28,7 +28,7 @@ systemBuilderPath builder = case builder of
     Cc  _  Stage0   -> fromKey "system-cc"
     Cc  _  _        -> fromKey "cc"
     -- We can't ask configure for the path to configure!
-    Configure _     -> return "bash configure"
+    Configure _     -> return "sh configure"
```
snowleopard: Should this be `bash ++ " configure"` too?

bgamari: Nope, as I mention in the commit message I don't think there's any reason why we need to use `bash` at all here. `sh` should be perfectly sufficient unless I've missed something.

snowleopard: 👍

src/Util.hs
```diff
@@ -62,10 +62,7 @@ customBuild rs opts target@Target {..} = do
                     top   <- topDirectory
                     cmd [Cwd output] [path] "x" (top -/- input)
 
-            Configure dir -> do
-                -- Inject /bin/bash into `libtool`, instead of /bin/sh
-                let env = AddEnv "CONFIG_SHELL" "/bin/bash"
```
snowleopard: Removing this breaks the Windows build (the CI failed).

I agree that this bit is mysterious and needs to be better documented and perhaps slightly refactored. Could you inject `bashPath` here and add a TODO for me?

bgamari:
Done.

bgamari:
Uh oh, it looks like this introduces a cycle,
```
/-------------------------------------------------------\
| Run Configure "." (stage = Stage0, package = ghc)     |
|      input: hadrian/cfg/system.config.in (and 2 more) |
|  => output: hadrian/cfg/system.config (and 2 more)    |
\-------------------------------------------------------/
shakeArgsWith   0.000s    0%                           
Function shake  0.270s   64%  =========================
Database read   0.000s    0%                           
With database   0.000s    0%                           
Running rules   0.149s   35%  =============            
Total           0.420s  100%                           
* OracleQ (ConfigKey "host-os")
* hadrian/cfg/system.config
* hadrian/cfg/system.config settings mk/config.h
* OracleQ (ConfigKey "host-os")
Build system error - recursion detected:
  Key type:   OracleQ ConfigKey
  Key value:  OracleQ (ConfigKey "host-os")
Rules may not be recursive
```

snowleopard:
@bgamari Yes, it looks like we'll have to hard code `"/bin/bash"` for now. Let's at least make it conditional for running on Windows, i.e. `when (System.os == "mingw32") ...`?

bgamari:
@snowleopard I just dropped the `fixAbsolutePathOnWindows` and it appears to pass on Windows. Any objections to this?

snowleopard:
@bgamari Aha, I see. It's probably fine, so I've merged this. Thank you!

PR #336, opened 2017-07-02
Various improvements to the Stack build.

mboes:
Have a look at the list of commits for details.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -25,6 +25,9 @@ install:
     - appveyor-retry stack exec -- pacman -S autoconf automake-wrapper make patch tar --noconfirm
 
 build_script:
+    # Workaround https://github.com/commercialhaskell/stack/issues/595
+    stack build alex happy
+
     # Build Hadrian
     - stack build
 
--- a/build.stack.sh
+++ b/build.stack.sh
@@ -31,6 +31,9 @@ function rl {
 absoluteRoot="$(dirname "$(rl "$0")")"
 cd "$absoluteRoot"
 
+# XXX workaround https://github.com/commercialhaskell/stack/issues/595
+stack build alex happy
+
 stack build --no-library-profiling
 
 stack exec hadrian --              \
--- a/hadrian.cabal
+++ b/hadrian.cabal
@@ -120,10 +120,10 @@ executable hadrian
                        , ansi-terminal        == 0.6.*
                        , Cabal                == 2.0.0.0
                        , containers           == 0.5.*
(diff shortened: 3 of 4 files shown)


snowleopard:
Many thanks @mboes!

Looks like this breaks the Windows CI: 

https://ci.appveyor.com/project/snowleopard/hadrian/build/1.0.432

Perhaps, we need to modify `appveyor.yml` to adapt to your `alex`/`happy` changes? If yes, does this mean we need to change this doc too: https://github.com/snowleopard/hadrian/blob/master/doc/windows.md?

Regarding the switch to LTS-8.21: please have a look at PR #292. Could you share your thoughts there?

mboes:
Could it be that earlier builds on Windows happened a while ago and therefore picked up an earlier version of Stack? I can't see anything in my PR that would cause the `alex` / `happy` problem to appear if it hadn't already been there. Maybe switching to a more recent LTS triggers it, but I'm skeptical. Either that or previous builds used a build cache primed with the right build tools and that cache expired maybe.

> Perhaps, we need to modify appveyor.yml to adapt to your alex/happy changes?

Yes, and I've just pushed a tentative change to that. But then we also need to change the docs, and then also `./build.stack.bat`. Could be perhaps have `appveyor.yaml` just call out to the latter to avoid triplicating the fix?

snowleopard:
> Could it be that earlier builds on Windows happened a while ago and therefore picked up an earlier version of Stack? 

I don't think so: look at #334, which has just built successfully after the failed build here.

> Could be perhaps have `appveyor.yaml` just call out to the latter to avoid triplicating the fix?

Good idea! It would be nice to make it work and reduce code duplication.

snowleopard:
Btw, I think you missed a `-` in front of `stack build alex happy`.

snowleopard:
@mboes Hey, we are ready to bump the LTS version, but it looks like the Stack's issue has been resolved meanwhile, so the workaround (installing Alex and Happy directly) is no longer required.

Would you like to remove the workaround from this PR, or shall I simply close this, and:
* Bump the LTS version
* Incorporate your changes to `stack.yaml` that remove the boilerplate and add `nix` stuff?

izgzhen:
@snowleopard I think we can bump the version safely now. since @mboes is not responsive, I guess you can fix it directly.

@mboes still a lot of thanks though

snowleopard:
@izgzhen Agreed, I'll do this today and will close the PR.

snowleopard:
Done.

PR #338, opened 2017-07-03
Update flavours doc

izgzhen:


snowleopard:
What about RTS ways? Have a look here:

https://github.com/snowleopard/hadrian/blob/master/src/Settings/Default.hs#L178-L180

Shouldn't we also add `debugDynamic`, `threadedDynamic`, `threadedDebugDynamic`, `loggingDynamic`, `threadedLoggingDynamic` into the RTS column?

izgzhen:
Thanks of reminding me these. I've updated a bit including the format, but wondering about why there is no a `threadedProfilingDynamic`?

bgamari:
> but wondering about why there isn't a `threadedProfilingDynamic`?

That looks like a bug to me. For instance,
```
$ make show! VALUE=GhcRTSWays
make --no-print-directory -f ghc.mk show NO_INCLUDE_PKGDATA=YES
GhcRTSWays="l debug thr thr_debug thr_l thr_p dyn debug_dyn thr_dyn thr_debug_dyn l_dyn thr_l_dyn"
```

snowleopard:
@bgamari I'm confused: your list doesn't have `thr_p_dyn` either.

bgamari:
Whoops! Indeed I misread `thr_p dyn` as `thr_p_dyn`. The eye will see what it wants to see.

That being said, `thr_p_dyn` *is* listed in `mk/ways.mk`. It would take some digging to determine why it's not included in `GhcRTSWays`.

snowleopard:
@bgamari Here is how RTS ways are computer:
```bash
# We always have the threaded versions, but note that SMP support may be disabled
# (see GhcWithSMP).
GhcRTSWays += thr thr_debug thr_l
GhcRTSWays += $(if $(findstring p, $(GhcLibWays)),thr_p,)
GhcRTSWays += $(if $(findstring dyn, $(GhcLibWays)),dyn debug_dyn thr_dyn thr_debug_dyn l_dyn thr_l_dyn,)
```
As you can see, `p` causes `thr_p` to be added, and `dyn` causes `thr_dyn` to be added. But `thr_p_dyn` is never added, because it requires both `p` and `dyn` preconditions.

Looks like a bug to me, but noone complained about it, so perhaps `thr_p_dyn` is never actually needed?

izgzhen:
So ... I guess we can open another issue for the discussion of `threadedProfilingDynamic`, and let this PR at least reflect what we currently have for now?

snowleopard:
@izgzhen Let me merge this. I don't think we need to create an issue about `threadedProfilingDynamic`, because the current build system is the closest artefact we have to a specification :)

Issue #248, opened 2016-05-11
Fix dynamicGhcPrograms (or not!)


thomie:
`git grep dynamicGhcPrograms` in the `hadrian` repository returns suspiciously few results. I don't think this is working yet.

All `utils` that are shipped with GHC (like `hsc2hs`, `haddock` etc), and also `ghc` itself, should be built `-dynamic`.

Depends on #4 (Fix dynamic way).

**BUT MAYBE IT IS NOT NECESSARY TO IMPLEMENT DYNAMIC_GHC_PROGRAMS**

Because we can make `ghci -fexternal-interpreter` the default. From: https://ghc.haskell.org/trac/ghc/wiki/RemoteGHCi#Motivation:

> It separates the decision about whether to use the dynamic linker for interpreted code from whether GHC itself must be dynamically linked. We could go back to a statically-linked GHC, which should be faster."

Ask @simonmar to be sure.

See also https://ghc.haskell.org/trac/ghc/wiki/DynamicLinkingDebate and https://ghc.haskell.org/trac/ghc/wiki/DynamicGhcPrograms.

DemiMarie:
Problem: currently the debugger does not work with `-fexternal-interpreter`.

bgamari:
Can this be closed now?

snowleopard:
@bgamari Not sure. I think @izgzhen fixed the `dynamic` way (on Linux and Mac) but we have not yet implemented full support for `dynamicGhcPrograms`. Do we need it?

bgamari:
Well, the current default build configuration of the make-based build system is `DYNAMIC_GHC_PROGRAMS=YES` where possible. We could (and perhaps should) reconsider this, but I'd be weary of changing just because the build system doesn't support it. In particular I'm a bit concerned about platforms where the RTS linker isn't terribly robust (e.g. ARM). In this case I fear that we may regress in terms of stability if we change. Consequently I think we should tread carefully.

izgzhen:
then at least it should be a post-merge goal?

snowleopard:
@bgamari @izgzhen Is this issue still relevant?

izgzhen:
not sure, checked the GHC Wiki and it seemed that both pages might be outdated (latest change 3 years ago).

snowleopard:
Let me close this. If this is acually still an issue, I suggest that we open a new one with an up-to-date description of what needs to be done.

PR #340, opened 2017-07-04
Add Stage1Only rule

### LINKED ISSUE #337: Implement Stage1Only
I think we probably don't have this feature and we need it to support #177.

Except for the control switch, we also need to know what can be built at stage0 and what can't (will `latestBuildStage` cover this? @snowleopard ?)

izgzhen:
Fix #337

### DIFF
--- a/src/Rules.hs
+++ b/src/Rules.hs
@@ -1,4 +1,4 @@
-module Rules (topLevelTargets, buildLib, buildRules) where
+module Rules (topLevelTargets, buildPackage, buildRules) where
 
 import Base
 import Context
@@ -18,22 +18,35 @@ import qualified Rules.Library
 import qualified Rules.Perl
 import qualified Rules.Program
 import qualified Rules.Register
+import Oracles.Dependencies (needContext)
+import Util (needBuilder)
 import Settings
 import Settings.Path
 
 allStages :: [Stage]
 allStages = [minBound ..]
 
--- | This rule 'need' all top-level build targets.
+-- | This rule 'need' all top-level build targets
+-- or Stage1Only targets
 topLevelTargets :: Rules ()
-topLevelTargets = do
-    want $ Rules.Generate.inplaceLibCopyTargets
+topLevelTargets = action $ do
+    need $ Rules.Generate.inplaceLibCopyTargets
 
-    forM_ allStages $ \stage ->
-        forM_ (knownPackages \\ [rts, libffi]) $ \pkg -> action (buildLib stage pkg)
+    if stage1Only
+        then do
+             forAllPkgs $ \stg pkg ->
+                 when (isLibrary pkg) $
+                     buildPackage stg pkg
+             forM_ programsStage1Only $ buildPackage Stage0
+        else
+             forAllPkgs buildPackage
+  where
+    forAllPkgs f =
+      forM_ allStages $ \stage ->
+          forM_ (knownPackages \\ [rts, libffi]) $ \pkg -> f stage pkg
 
-buildLib :: Stage -> Package -> Action ()
-buildLib stage pkg = do
+buildPackage :: Stage -> Package -> Action ()
+buildPackage stage pkg = do
     let context = vanillaContext stage pkg
     activePackages <- interpretInContext context getPackages
     when (pkg `elem` activePackages) $
@@ -44,6 +57,7 @@ buildLib stage pkg = do
             ways <- interpretInContext context getLibraryWays
             libs <- mapM (pkgLibraryFile . Context stage pkg) ways
             docs <- interpretInContext context $ buildHaddock flavour
+            needContext [context]
             need $ libs ++ [ pkgHaddockFile context | docs && stage == Stage1 ]
         else -- otherwise build a program
             need =<< maybeToList <$> programPath (programContext stage pkg)
@@ -90,3 +104,9 @@ buildRules = do
     Rules.Libffi.libffiRules
     packageRules
     Rules.Perl.perlScriptRules
+
+programsStage1Only :: [Package]
+programsStage1Only =
+  [ deriveConstants, genprimopcode, hp2ps, runGhc
+  , ghcCabal, hpc, dllSplit, ghcPkg, hsc2hs
+  , genapply, ghc ]
--- a/src/UserSettings.hs
+++ b/src/UserSettings.hs
@@ -4,7 +4,7 @@
 -- accidentally commit them.
 module UserSettings (
     buildRootPath, userFlavours, userKnownPackages, verboseCommands,
-    putBuild, putSuccess, destDir
+    putBuild, putSuccess, defaultDestDir, defaultStage1Only
     ) where
 
 import System.Console.ANSI
@@ -47,5 +47,20 @@ putSuccess = putColoured Dull Green
 -- It is by default empty, representing the root of file system,
 -- or it might be a directory.
 -- It is usually used with @prefix@, like @/usr/local@
-destDir :: FilePath
-destDir = ""
+defaultDestDir :: FilePath
+defaultDestDir = ""
+
+{-
+  Stage1Only=YES means:
+   - don't build ghc-stage2 (the executable)
+   - don't build utils that rely on ghc-stage2
+     See Note [No stage2 packages when CrossCompiling or Stage1Only] in
+     ./ghc.mk.
+   - install ghc-stage1 instead of ghc-stage2
+   - install the ghc-pkg that was built with the stage0 compiler
+   - (*do* still build compiler/stage2 (i.e. the ghc library))
+   - (*do* still build all other libraries)
+-}
+-- | Stage1Only flag, default off
+defaultStage1Only :: Bool
+defaultStage1Only = False
--- a/src/Oracles/Dependencies.hs
+++ b/src/Oracles/Dependencies.hs
@@ -63,7 +63,7 @@ needContext cs = do
         lib0     <- buildDll0          context
         ghciLib  <- pkgGhciLibraryFile context
         ghciFlag <- interpretInContext context $ getPkgData BuildGhciLib
-        let ghci = ghciFlag == "YES" && stage context == Stage1
+        let ghci = ghciFlag == "YES" && (stage context == Stage1 || stage1Only)
         return $ [ libFile ] ++ [ lib0File | lib0 ] ++ [ ghciLib | ghci ]
     confs <- mapM pkgConfFile cs
     need $ libs ++ confs
--- a/src/Rules/Install.hs
+++ b/src/Rules/Install.hs
@@ -191,7 +191,7 @@ installPackages = do
                 let context = vanillaContext stg pkg
                 top <- interpretInContext context getTopDirectory
                 let installDistDir = top -/- buildPath context
-                buildLib stg pkg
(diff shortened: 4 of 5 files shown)


src/Rules.hs
```diff
@@ -44,7 +45,14 @@ buildLib stage pkg = do
             ways <- interpretInContext context getLibraryWays
             libs <- mapM (pkgLibraryFile . Context stage pkg) ways
             docs <- interpretInContext context $ buildHaddock flavour
-            need $ libs ++ [ pkgHaddockFile context | docs && stage == Stage1 ]
+
+            version <- interpretInContext context (getPkgData Version)
+            let topObj = buildPath context -/- "HS" ++
+                         pkgNameString pkg ++ "-" ++ version ++
+                         (waySuffix (way context) <.> "o")
```
izgzhen: @snowleopard BTW, this `topObj` is a bugfix not quite related to the topic of this PR though. We forget to build, for example, equivalence of `ghc-standard/libraries/ghc-compact/dist-install/build/HSghc-compact-0.1.0.0.o`.

snowleopard: Hmm, isn't this the same path as computed by this function?

https://github.com/snowleopard/hadrian/blob/master/src/Settings/Path.hs#L104-L107

snowleopard:
How does `stage1only` target interact with these requirements from your comment?
```
   - install ghc-stage1 instead of ghc-stage2
   - install the ghc-pkg that was built with the stage0 compiler
```
I mean, what does "install" mean in this context? Creating a binary in `inplace/bin`?

src/Rules.hs
izgzhen: Exactly! Then why isn't it built for packages beyond `ghci`?

izgzhen:
>  what does "install" mean in this context? Creating a binary in inplace/bin?

I guess it means `make install`. But for `inplace/bin`, here all I can find:

```
count_lines     genapply        ghc-pkg         hpc             runghc
deriveConstants genprimopcode   ghc-stage1      hsc2hs
dll-split       ghc-cabal       hp2ps           mkdirhier
```

src/Rules.hs
snowleopard: I think these objects are built for all libraries (I call them "ghci libraries", because as far as I understand they are required for `ghci`). Here is the build rule: 

https://github.com/snowleopard/hadrian/blob/master/src/Rules/Library.hs#L83-L89

Perhaps, the actual bug is here: 

https://github.com/snowleopard/hadrian/blob/master/src/Oracles/Dependencies.hs#L64-L67

snowleopard:
Right. So if I understand your response correctly, the current implementation of `stage1only` target already does the 'install' part.

src/Rules.hs
izgzhen: Intersting ... `ghc-compact`'s `HS<pkg>.o` is not built by default IIRC, and `haskeline`'s `HS<pkg>.o`, after deleted, can't be rebuilt by `build`.

izgzhen: This should be relevant to #248 as well IIRC?

izgzhen:
> Right. So if I understand your response correctly, the current implementation of stage1only target already does the 'install' part.

Not quite. In the old system, `Stage1Only` is a static flag in `build.mk`, but here if we are really going to make it a phony target, then it will conflict with the duty of `install` phony target in that aspect.

izgzhen:
So maybe it should be made a setting instead?

src/Rules.hs
snowleopard: Could this be because of `stage context == Stage1` condition in the latest link? Perhaps, we are missing `|| isStage1Only`?

snowleopard:
> So maybe it should be made a setting instead?

Yes, this will be closer to Make's behaviour. To add a new setting, you'll probably need to modify the `Flavour` datatype and all relevant bits:

https://github.com/snowleopard/hadrian/blob/master/src/Flavour.hs

Then the user will be able to override the settings. We could also provide a corresponding command line flag, i.e. see how `splitObjects` or `buildHaddock` are implemented.

src/Rules.hs
izgzhen: Why `|| isStage1Only`? I thought it should be stage-irrelavant ... let's me check out the Makefiles sometime

snowleopard: Well, my intuition says that since everything worked fine with `stage context == Stage1` condition but started to produce incorrect results with Stage1Only flag, this has something to do with the stage when we build these objects.

izgzhen:
As a corrected note, all following `HS<pkg>.o` that will be built by default:

```
_build/stage1/libraries/ghc-prim/HSghc-prim-0.5.0.0.o
_build/stage1/libraries/integer-simple/HSinteger-simple-0.1.1.1.o
_build/stage1/libraries/base/HSbase-4.10.0.0.o
_build/stage1/libraries/array/HSarray-0.5.1.2.o
_build/stage1/libraries/deepseq/HSdeepseq-1.4.3.0.o
_build/stage1/libraries/bytestring/HSbytestring-0.10.8.2.o
_build/stage1/libraries/time/HStime-1.8.0.1.o
_build/stage1/libraries/pretty/HSpretty-1.1.3.3.o
_build/stage1/libraries/ghc-boot-th/HSghc-boot-th-8.3.o
_build/stage1/libraries/unix/HSunix-2.7.2.2.o
_build/stage1/libraries/filepath/HSfilepath-1.4.1.2.o
_build/stage1/libraries/directory/HSdirectory-1.3.0.2.o
_build/stage1/libraries/process/HSprocess-1.6.0.0.o
_build/stage1/libraries/transformers/HStransformers-0.5.2.0.o
_build/stage1/libraries/template-haskell/HStemplate-haskell-2.12.0.0.o
_build/stage1/libraries/containers/HScontainers-0.5.10.2.o
_build/stage1/libraries/binary/HSbinary-0.8.4.1.o
_build/stage1/libraries/ghc-boot/HSghc-boot-8.3.o
_build/stage1/libraries/ghci/HSghci-8.3.o
_build/stage1/utils/hsc2hs/HSCParser.o
_build/stage1/libraries/hpc/HShpc-0.6.0.3.o
_build/stage1/libraries/terminfo/HSterminfo-0.4.1.0.o
_build/stage1/libraries/Cabal/Cabal/HSCabal-2.0.0.0.o
_build/stage1/libraries/haskeline/HShaskeline-0.7.4.0.o
```

There should be no problem rebuilding them after deletion.

izgzhen:
Maybe I understand the problem like this:

For our existing implementation, a `HS<pkg>.o` has two conditions to be actually needed: 

1. Satisfy the condition as in `needContext`
2. The context is actually needed (being depended on)

For `ghc-compact` and `haskeline`, maybe the second condition isn't satisfied (not sure, and doesn't explain the difference of some packages not being tracked).

But we actually *need* them since ghci might load them when user uses it.

izgzhen:
I've updated my change according the the above comment

src/Rules.hs
```diff
@@ -29,8 +30,11 @@ topLevelTargets :: Rules ()
...
 
-    forM_ allStages $ \stage ->
-        forM_ (knownPackages \\ [rts, libffi]) $ \pkg -> action (buildLib stage pkg)
+    if stage1Only
+       then action buildStage1Only
+       else forM_ allStages $ \stage ->
+                 forM_ (knownPackages \\ [rts, libffi]) $ \pkg ->
+                      action (buildLib stage pkg)
 
 buildLib :: Stage -> Package -> Action ()
```
snowleopard: This also builds programs (not only libraries), so I think the name is confusing.

src/Rules.hs
```diff
@@ -90,3 +95,29 @@ buildRules = do
...
+   - don't build utils that rely on ghc-stage2
+     See Note [No stage2 packages when CrossCompiling or Stage1Only] in
+     ./ghc.mk.
+   - install ghc-stage1 instead of ghc-stage2
+   - install the ghc-pkg that was built with the stage0 compiler
+   - (*do* still build compiler/stage2 (i.e. the ghc library))
+   - (*do* still build all other libraries)
+-}
+
+buildStage1Only :: Action ()
```
snowleopard: It appears there is quite a bit of code duplication here. Can we get rid of it?

src/UserSettings.hs
```diff
@@ -49,3 +49,7 @@ putSuccess = putColoured Dull Green
 -- It is usually used with @prefix@, like @/usr/local@
 destDir :: FilePath
 destDir = ""
+
+-- | Stage1Only flag, default off
+defaultStage1Only :: Bool
+defaultStage1Only = True
```
snowleopard: Wait, if the default is 'off', shouldn't it be `False`?

src/UserSettings.hs
```diff
@@ -49,3 +49,7 @@ putSuccess = putColoured Dull Green
 -- It is usually used with @prefix@, like @/usr/local@
 destDir :: FilePath
```
snowleopard: Just spotted: should this be also called `defaultDestDir`, for consistency?

src/Rules.hs
```diff
@@ -90,3 +95,29 @@ buildRules = do
     Rules.Libffi.libffiRules
     packageRules
     Rules.Perl.perlScriptRules
+
+programsStage1Only :: [Package]
+programsStage1Only =
+  [ deriveConstants, genprimopcode, hp2ps, runGhc
```
snowleopard: My guess is that most of these will be transitively required anyway. Couldn't we simply ask to build `ghc`?

snowleopard:
@izgzhen I've added a bunch of comments, please have a look.

snowleopard:
By the way, I think the CI failed because `defaultStage1Only` is currently set to `True`, so `ghc-stage2` is missing.

src/Rules.hs
izgzhen: Agreed

src/UserSettings.hs
izgzhen: Ahh, sorry that I pushed my local change ...

src/UserSettings.hs
izgzhen: Sure

src/Rules.hs
izgzhen: No problem

src/Rules.hs
izgzhen: No problem

src/Rules.hs
izgzhen: Not really, `needBuilder (Ghc CompileHs Stage1)` or `buildPackage Stage0 ghc` will only build:

```
deriveConstants genapply        genprimopcode   ghc-pkg         ghc-stage1
```

which is less then my previous comment about what the legacy system would build.

izgzhen:
Updated

src/Rules.hs
snowleopard: Aha, I see. Let's keep the list as is then.

snowleopard:
@izgzhen Looks good now. Shall I merge?

snowleopard:
Ah, wait. Looks like a compile error now?

izgzhen:
Sorry that I try to avoid adding local change in `UserSettings.hs` but also avoiding a necessary patch at the same time. I've updated the code and if it can pass CI I think we can merge this now.

snowleopard:
Thanks, merged!

PR #342, opened 2017-07-04
Automate dependency analysis of installed packages

izgzhen:


izgzhen:
Note that this has conflict with #340 because I also need the `HS-<pkg>.o` here.

snowleopard:
I'm lost here. Is this addressing a specific issue? What problem is being solved?

snowleopard:
Ah, I think it's this todo: `Use automatic dependency analysis, rather than hardcoding the ordering`.

src/Oracles/Dependencies.hs
```diff
@@ -81,3 +81,16 @@ dependenciesOracles = do
...
+
+sortPkgsByDep :: [Package] -> Action [Package]
+sortPkgsByDep pkgs = do
+    elems <- mapM (\p -> (p,) <$> pkgDependencies p) pkgs
+    return $ map fst $ topSort elems
+  where
+    annotateInDeg es e =
+     (foldr (\e' s -> if fst e' `elem` snd e then s + 1 else s) 0 es, e)
+    topSort [] = []
+    topSort es =
```
snowleopard: This looks very suspicious. Why do we need topological sort? Isn't this what Shake is for?

izgzhen: In this case, I am using it to install packages (`Rules.Install.installPackages`). But since we don't track installed artifacts and simply execute the installation repetitively, so there is no dependency information maintained.

I think, to use Shake's functionality, we need to write installPackages as a set of `Rules`, instead of `Action`.

snowleopard: I see, thanks for clarifying.

> I think, to use Shake's functionality, we need to write installPackages as a set of Rules, instead of Action.

Perhaps, we should indeed do this? Reinventing `topSort` on the spot seems like a bad idea to me... Let's at least add a `TODO` here to think how we can get rid of this in future.

izgzhen: Good

izgzhen: Add an issue, and I will link it back to https://github.com/snowleopard/hadrian/issues/344

izgzhen:
This passes CI fine.

snowleopard:
Some conflicts here now.

src/Rules/Install.hs
```diff
@@ -177,13 +179,10 @@ installPackages = do
     copyFile ghcBootPlatformHeader (pkgPath compiler -/- "ghc_boot_platform.h")
```
snowleopard: Can this TODO be dropped now?

izgzhen: Sure

izgzhen:
Updated

snowleopard:
Right, this looks good to me, so I'm merging.

PR #351, opened 2017-07-07
Parallelize stage0 build

bgamari:
This eliminates a rather significant duration of serial compilation from the
build.

I'm still not 100% certain we want to do this by default, but I thought it would be good to put it up for discussion.

### DIFF
--- a/src/Settings/Default.hs
+++ b/src/Settings/Default.hs
@@ -262,6 +262,13 @@ disableWarningArgsLibs = do
     [ notStage0 ? arg "-Wno-deprecated-flags"
     , stage Stage0 ? arg "-fno-warn-deprecated-flags"]
 
+-- | Perform stage0 build (which uses a monolithic @ghc --make@ invocation)
+-- with @-j@.
+parallelStage0BuildArgs :: Args
+parallelStage0BuildArgs = do
+    threads <- shakeThreads <$> lift getShakeOptions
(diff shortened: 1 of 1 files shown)


izgzhen:
Will `build.sh -j` have the same effect?

bgamari:
> Will `build.sh -j` have the same effect?

It won't; this only applies to the initial bootstrap build of `Cabal` and `ghc-cabal`, which is currently serial.

> From the issue content, I guess your answer would be "not" ... but at least we should only invoke this when user add -j to build.sh

Certainly a good point.

snowleopard:
@bgamari We could also add `-j` conditionally, e.g. see 

https://github.com/snowleopard/hadrian/blob/master/src/Settings/Builders/Make.hs#L8

That is, if Hadrian is run in single-threaded mode, we don't parallelise this bit.

bgamari:
Thanks for the hint, @snowleopard! How does this look?

snowleopard:
@bgamari Wait, if I understand correctly, this is supposed to only apply to building `ghc-cabal`, right? But then I just found that we already do `-j` when building it. See here:

https://github.com/snowleopard/hadrian/blob/master/src/Settings/Packages/GhcCabal.hs#L29

Are you sure the change you are doing in this PR actually speeds up your build?

bgamari:
Ahh, indeed we do! I honestly didn't check the Hadrian side of things so consider this closed.

PR #352, opened 2017-07-07
Update README.md

izgzhen:


README.md
```diff
@@ -106,6 +106,14 @@ are still up-to-date.
 
 To build a GHC source distribution tarball, run Hadrian with the `sdist-ghc` target.
 
+#### Installation
+
+To install GHC artifacts, first build everything, then run the `install` target.
```
snowleopard: Why is the first step needed? All required artefacts should be (re)built automatically if need be, before copying, via Shake's tracking. Users shouldn't have to remember these things.

README.md
```diff
@@ -122,12 +130,11 @@ zero (see [#197][test-issue]).
 Current limitations
 -------------------
 The new build system still lacks many important features:
-* There is currently no support for the `dynamic` build way: [#4][dynamic-issue].
```
snowleopard: There is currently no support for dynamic linking on Windows.

README.md
izgzhen: That's a great point -- but this is only partially supported now. I should fix it later

izgzhen: Fixing should be simple, I will do that first.

snowleopard: Thanks! Ping me when ready.

izgzhen:
This one is also updated.

snowleopard:
Thanks, merged!

PR #353, opened 2017-07-08
Better tracking of dependence in installation

izgzhen:


src/Rules/Install.hs
```diff
@@ -131,18 +128,18 @@ withLatestBuildStage pkg m = do
...
 installPackageConf :: Action ()
-installPackageConf = do
+installPackageConf = do    
     let context = vanillaContext Stage0 rts
     liftIO $ IO.createDirectoryIfMissing True (takeDirectory pkgConfInstallPath)
     build $ Target context HsCpp [ pkgPath rts -/- "package.conf.in" ]
                                  [ pkgConfInstallPath <.> "raw" ]
-    Stdout out <- cmd ("grep" :: String) [ "-v", "^#pragma GCC"
-                                         , pkgConfInstallPath <.> "raw" ]
+    Stdout content <- cmd ("grep" :: String) [ "-v", "^#pragma GCC"
```
snowleopard: Do you need `:: String` because of `OverloadedStrings`? If yes, how do you use the extension? (I couldn't find.)

snowleopard:
Looks good! Could you clarify what this improves, specifically?

src/Rules/Install.hs
izgzhen: Thanks for points this out! It is not useful now, and I'll drop it. IIRC, it was initially for doing binary patch, so I need to write a literal `ByteString`.

izgzhen:
> Could you clarify what this improves, specifically?

Sorry I didn't mention that it was for one point we discussed in #352: with this, we can do `build install` safely without doing `build` beforehand.

snowleopard:
> we can do `build install` safely without doing `build` beforehand.

Great! I was hoping for this.

I'll merge when CI bots complete.

izgzhen:
CI Completed

PR #357, opened 2017-07-10
Don't optimize cabal stage0 build

bgamari:
The performance of the stage0 `Cabal` build is not critical. I am guessing that this will help bring down the stage0 compilation time, although I have yet to quantify this.

### DIFF
--- a/hadrian.cabal
+++ b/hadrian.cabal
@@ -91,6 +91,7 @@ executable hadrian
                        , Settings.Flavours.Quick
                        , Settings.Flavours.Quickest
                        , Settings.Packages.Base
+                       , Settings.Packages.Cabal
                        , Settings.Packages.Compiler
                        , Settings.Packages.Ghc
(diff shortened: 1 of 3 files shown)


snowleopard:
Makes sense! It's sad that this won't help the CI bots, because in the `quickest` flavour we already add `-O0`.

In the current architecture, we put all package-specific settings into `src/Settings/Packages`. Shall we move this tweak there? The downside is that this will require adding a new file `src/Settigns/Packages/Cabal.hs`, which sounds like an overkill. Still this may be a convenient convention in the long-term.

snowleopard:
The current version breaks CI, because `package cabal ? stage Stage0 ? append [ "-O0" ]` adds `-O0` to all bulders. There should be another predicate `builder Ghc`.

bgamari:
I think this should be all set now.

snowleopard:
Thanks @bgamari! Merged.

PR #362, opened 2017-07-14
Fix validate's dependency on generated files

izgzhen:


snowleopard:
@izgzhen Thank you, this looks fine.

PR #365, opened 2017-07-16
Fix double copy

izgzhen:
Verified that with this change, https://github.com/snowleopard/hadrian/issues/323 can be resolved. But this is still a HACK because of https://github.com/snowleopard/hadrian/issues/318.

### DIFF
--- a/src/Rules/Install.hs
+++ b/src/Rules/Install.hs
@@ -194,22 +194,17 @@ installPackages = do
                 need [ ghcCabalInplace ]
 
                 let cabalFile = pkgCabalFile pkg
-                -- HsColour sources
-                -- QUESTION: what is the output of GhcCabalHsColour?
-                whenM (isSpecified HsColour) $ do
-                    top <- interpretInContext context getTopDirectory
-                    let installDistDir = top -/- buildPath context
-                    -- HACK: copy stuff back to the place favored by ghc-cabal
-                    quietly $ copyDirectoryContents (Not excluded)
-                                  installDistDir (installDistDir -/- "build")
 
                 pkgConf <- pkgConfFile context
                 need [ cabalFile, pkgConf ] -- TODO: check if need pkgConf
-                build $ Target context GhcCabalHsColour [cabalFile] []
 
                 -- HACK (#318): copy stuff back to the place favored by ghc-cabal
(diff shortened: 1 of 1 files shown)


src/Rules/Install.hs
```diff
@@ -200,15 +200,15 @@ installPackages = do
...
+                    quietly $ copyDirectoryContentsUntracked (Not excluded)
                                   installDistDir (installDistDir -/- "build")
 
                 pkgConf <- pkgConfFile context
                 need [ cabalFile, pkgConf ] -- TODO: check if need pkgConf
                 build $ Target context GhcCabalHsColour [cabalFile] []
 
                 -- HACK (#318): copy stuff back to the place favored by ghc-cabal
-                quietly $ copyDirectoryContents (Not excluded)
+                quietly $ copyDirectoryContentsUntracked (Not excluded)
```
snowleopard: It looks like when `isSpecified HsColour` holds, you'll do the copying twice?

izgzhen: Indeed!

src/Rules/Install.hs
```diff
@@ -194,22 +194,16 @@ installPackages = do
...
                 pkgConf <- pkgConfFile context
                 need [ cabalFile, pkgConf ] -- TODO: check if need pkgConf
-                build $ Target context GhcCabalHsColour [cabalFile] []
 
                 -- HACK (#318): copy stuff back to the place favored by ghc-cabal
-                quietly $ copyDirectoryContents (Not excluded)
+                quietly $ copyDirectoryContentsUntracked (Not excluded)
                             installDistDir (installDistDir -/- "build")
+
+                build $ Target context GhcCabalHsColour [cabalFile] []
```
snowleopard: This line should be conditional on `whenM (isSpecified HsColour)`, right?

izgzhen: Right!

izgzhen:
I've just verified that tracked copy is not really the source of lint error, while double copy probably is, which is fixed now.

snowleopard:
Ah, great! I'll merge this once CI passes.

snowleopard:
Not sure what happened with AppVeyor and CircleCI. I've restarted AppVeyor.

Issue #105, opened 2016-01-05
iserv-bin requires a wrapper


angerman:
See #102. This is here, so we don't forget to implement the wrapper.

angerman:
And it also requires to build `p` and `dyn` versions.

snowleopard:
Just for the record: I think `profiling` might works already (it's currently disabled not to slow down test builds), and `dynamic` is handled in this unbelievably old issue: #4.

angerman:
Ohh, what I meant was that the `ghc.mk` for `iserv` does some really weird stuff. It basically add's `iserv` in dynamic and profiling to the targets. The `ghc.mk` looks like this:

```make
# -----------------------------------------------------------------------------
#
# (c) 2009-2012 The University of Glasgow
#
# This file is part of the GHC build system.
#
# To understand how the build system works and how to modify it, see
#      http://ghc.haskell.org/trac/ghc/wiki/Building/Architecture
#      http://ghc.haskell.org/trac/ghc/wiki/Building/Modifying
#
# -----------------------------------------------------------------------------

iserv_USES_CABAL = YES
iserv_PACKAGE = iserv-bin

ifeq "$(GhcDebugged)" "YES"
iserv_stage2_MORE_HC_OPTS += -debug
iserv_stage2_p_MORE_HC_OPTS += -debug
iserv_stage2_dyn_MORE_HC_OPTS += -debug
endif

iserv_stage2_MORE_HC_OPTS += -threaded
iserv_stage2_p_MORE_HC_OPTS += -threaded
iserv_stage2_dyn_MORE_HC_OPTS += -threaded

# Override the default way, because we want a specific version of this
# program for each way.  Note that it's important to do this even for
# the vanilla version, otherwise we get a dynamic executable when
# DYNAMIC_GHC_PROGRAMS=YES.
iserv_stage2_PROGRAM_WAY = v
iserv_stage2_p_PROGRAM_WAY = p
iserv_stage2_dyn_PROGRAM_WAY = dyn

iserv_stage2_PROGNAME = ghc-iserv
iserv_stage2_p_PROGNAME = ghc-iserv-prof
iserv_stage2_dyn_PROGNAME = ghc-iserv-dyn

iserv_stage2_MORE_HC_OPTS += -no-hs-main
iserv_stage2_p_MORE_HC_OPTS += -no-hs-main
iserv_stage2_dyn_MORE_HC_OPTS += -no-hs-main

iserv_stage2_INSTALL = YES
[...]

PR #367, opened 2017-07-17
Add ghc-iserv wrapper

izgzhen:
Fixes https://github.com/snowleopard/hadrian/issues/105.

verified with:

```
inplace/bin/ghc-stage2 -fexternal-interpreter --interactive
```

### DIFF
--- a/src/Rules/Program.hs
+++ b/src/Rules/Program.hs
@@ -34,10 +34,34 @@ buildProgram rs context@Context {..} = when (isProgram package) $ do
         inplaceBinPath -/- programName context <.> exe %> \bin -> do
             binStage <- installStage
             buildBinaryAndWrapper rs (context { stage = binStage }) bin
-        -- We build only unwrapped binaries in inplace/lib/bin
+
         inplaceLibBinPath -/- programName context <.> exe %> \bin -> do
             binStage <- installStage
+            if package /= iservBin then
+                -- We *normally* build only unwrapped binaries in inplace/lib/bin,
+                buildBinary rs (context { stage = binStage }) bin
+            else
+                -- build both binary and wrapper in inplace/lib/bin
+                -- for ghc-iserv on *nix platform now
+                buildBinaryAndWrapperLib rs (context { stage = binStage }) bin
+
+        inplaceLibBinPath -/- programName context <.> "bin" %> \bin -> do
+            binStage <- installStage
             buildBinary rs (context { stage = binStage }) bin
+
+buildBinaryAndWrapperLib :: [(Resource, Int)] -> Context -> FilePath -> Action ()
+buildBinaryAndWrapperLib rs context bin = do
+    windows <- windowsHost
+    if windows
+    then buildBinary rs context bin -- We don't build wrappers on Windows
+    else case lookup context inplaceWrappers of
+        Nothing      -> buildBinary rs context bin -- No wrapper found
+        Just wrapper -> do
+            top <- topDirectory
+            let libdir = top -/- inplaceLibPath
+            let wrappedBin = inplaceLibBinPath -/- programName context <.> "bin"
+            need [wrappedBin]
+            buildWrapper context wrapper bin (WrappedBinary libdir (takeFileName wrappedBin))
 
 buildBinaryAndWrapper :: [(Resource, Int)] -> Context -> FilePath -> Action ()
 buildBinaryAndWrapper rs context bin = do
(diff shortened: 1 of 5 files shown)


izgzhen:
BTW, the weird name and layout are designed to conform with Make-system. It looks like `ghc-stage2` hardcoded the name of iserv it will call as external interpreter (i.e. `lib/bin/ghc-iserv`)

src/Rules/Program.hs
```diff
@@ -37,8 +37,26 @@ buildProgram rs context@Context {..} = when (isProgram package) $ do
         -- We build only unwrapped binaries in inplace/lib/bin
         inplaceLibBinPath -/- programName context <.> exe %> \bin -> do
             binStage <- installStage
+            buildBinaryAndWrapperLib rs (context { stage = binStage }) bin
+
+        inplaceLibBinPath -/- programName context <.> "bin" <.> exe %> \bin -> do
```
snowleopard: Do we need `<.> exe` here? Presumably this rule is not relevant for Windows, and that't the only case when `exe` is not empty.

src/Rules/Program.hs
```diff
@@ -37,8 +37,26 @@ buildProgram rs context@Context {..} = when (isProgram package) $ do
...
+buildBinaryAndWrapperLib rs context bin = do
+    windows <- windowsHost
+    if windows
+    then buildBinary rs context bin -- We don't build wrappers on Windows
+    else case lookup context inplaceWrappers of
+        Nothing      -> buildBinary rs context bin -- No wrapper found
+        Just wrapper -> do
+            top <- topDirectory
+            let libdir = top -/- inplaceLibPath
+            let wrappedBin = inplaceLibBinPath -/- programName context <.> "bin" <.> exe
```
snowleopard: Same here: `<.> exe` is a no-op.

src/Rules/Program.hs
izgzhen: I actually don't know how it looks like on Windows, at least on Mac, the binary is `ghc-iserv.bin` and the wrapper is `ghc-iserv`. So I guess on Windows, the binary might be `ghc-iserv.bin.exe` and the wrapper is `ghc-iserv.exe`

izgzhen: So, you mean that on Windows the intended ".exe" becomes a no-op and won't be appended?

snowleopard: No, I mean `iserv` is not built on Windows at all. So, this code will never run on Windows, hence no need for `exe`.

izgzhen: Oh, I got it.

src/Rules/Wrappers.hs
```diff
@@ -117,13 +118,35 @@ haddockWrapper WrappedBinary{..} = do
...
 
+iservBinWrapper :: WrappedBinary -> Expr String
+iservBinWrapper WrappedBinary{..} = do
+    lift $ need [sourcePath -/- "Rules/Wrappers.hs"]
+    activePackages <- filter isLibrary <$> getPackages
+    -- TODO: Figure our the reason of this hardcoded exclusion
+    let pkgs = activePackages \\ [ cabal, process, haskeline
+                                 , terminfo, ghcCompact, hpc, compiler ]
+    contexts <- catMaybes <$> mapM (\p -> do
+                                        m <- lift $ latestBuildStage p
```
snowleopard: Can we just `return $ fmap (\s -> vanillaContext s p) m` here?

src/Rules/Program.hs
izgzhen: But wait, since it is not built at all, so that doesn't harm right? What if Windows support gets added someday in future? I still think that leaving the code as general as we can now would be beneficial.

src/Rules/Wrappers.hs
izgzhen: Good idea

src/Rules/Program.hs
snowleopard: It sounds a bit far-fetched, because there are no wrappers at all on Windows at the moment. It's unclear whether they will ever be needed, even if `iserv` is added someday. If this does happen, then adding a few `exe` in the right spots will be easy, but right now it looks like a potential source of confusion.

izgzhen:
Fixed.

snowleopard:
Great, thanks! Will merge after CI.

snowleopard:
CI fails with:
```
* inplace/lib/bin/touchy.exe
  inplace/lib/bin/touchy.exe
```

izgzhen:
I think it is ready now.

snowleopard:
Thanks, merged!

PR #369, opened 2017-07-18
Fix typo in comment

duog:


snowleopard:
Thanks @duog! Good catch :-)

Issue #322, opened 2017-06-11
OS X CI timeout


izgzhen:
This is not a new issue: we had frequent OS X CI timeout on Travis, because the build took too long. Very sparsely, it can finish within the time limit.

It can be annoying since we might never know if the build will actually succeed or not (lucky that my dev machine is OS X), not to say spending more time running tests.

[Circle CI](http://circleci.com) provides better OS X support, but it is not free. Travis also supports a beta-version [cache mechanism](https://docs.travis-ci.com/user/caching/), which seems useful (as in https://github.com/travis-ci/travis-ci/issues/6410), but I am not sure about this.

snowleopard:
One possible solution is to use a different (faster) bootstrapping compiler. I hear GHC 8.2 is fast -- perhaps it could bring build times down?

izgzhen:
> I hear GHC 8.2 is fast -- perhaps it could bring build times down?

I can try this one later on.

izgzhen:
> > I hear GHC 8.2 is fast -- perhaps it could bring build times down?
> 
> I can try this one later on.

Well, `brew` only supports the stable release, i.e. `8.0.2` (which is supposed to be used currently).

izgzhen:
As mentioned in #325, this time even Linux build timeouts.

Another thought on infrastructure is to see if the larger community can help on this, like providing a self-hosted Linux build bot.

@jaspervdj Do you have any suggestion on this?

izgzhen:
Building with `dyn` way on my MacBook Pro needs 112 minutes

jaspervdj:
The caching mechanism for Travis works fairly well.

I'm not sure about a build bot, but since this is for a GHC-related project, @hvr might know more.

izgzhen:
https://circleci.com/gh/izgzhen/hadrian/43 (not sure it is publicly viewable...)

I've successfully built hadrian with CircleCI OS X. I applied for their OSS seed plan two weeks ago, and it looks like their service is less likely to time out than Travis's.

@snowleopard What's your idea?

snowleopard:
@izgzhen Nice! Looks like it's pretty fast. But as far as I see it's not free (beyond the two-week trial).

izgzhen:
I applied a special program for Open Source project, we can use the Seed
Plan for free.

snowleopard:
If we can use it for free then sure, let's add another CI integration to Hadrian!

izgzhen:
[...]

PR #370, opened 2017-07-19
Make OS X build faster and Add GHC 8.0.2 build on Travis CI

izgzhen:
Related: #322, #292

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,36 @@
 sudo: true
-
+language: haskell
 matrix:
     include:
         - os: linux
           env: MODE="--flavour=quickest"
+          compiler: "GHC 8.0.2"
+          addons:
+              apt:
+                  packages:
+                      - ghc-8.0.2
+                      - cabal-install-2.0
+                      - zlib1g-dev
+                  sources: hvr-ghc
+
+          before_install:
+              - PATH="/opt/ghc/8.0.2/bin:$PATH"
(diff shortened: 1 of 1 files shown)


izgzhen:
Looks like Travis CI OS X can't afford a full build, even if we use ghc 8.0.2 https://travis-ci.org/izgzhen/hadrian/jobs/255095240

izgzhen:
Updated, now it is a 7.10.3 Linux + 8.0.2 Linux + 7.10.3 OS X (stage1) combo

snowleopard:
Great, thanks! Will merge it once it goes through.

snowleopard:
OS X finished pretty quickly! 

https://travis-ci.org/snowleopard/hadrian/builds/255238071?utm_source=github_status&utm_medium=notification

@izgzhen Do you know if we could somehow make it visible on the above page which GHC is used for the two Linux instances? Then if one of them fails, one could easily which version is broken. Perhaps, we could do this by adding a dummy variable "GHC=7.10.3" and "GHC=8.0.2"?

izgzhen:
I can let this page display the version info (https://travis-ci.org/snowleopard/hadrian/builds/255282296) together, but not on badge. I checked several other projects, it looks like there is no way to make a fancier badge.

snowleopard:
> I can let this page display the version info together, by adding a compiler:

That's perfect, thanks! I've merged this.

PR #372, opened 2017-07-19
Drop language in Travis CI config

izgzhen:
The latest build failure https://travis-ci.org/snowleopard/hadrian/jobs/255299317 is probably caused by this. Adding this line makes the build status list clearer since there is not a non-relavent "Ruby", but I should anticipate that it might introduce some unwanted effects.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 sudo: true
-language: haskell
 matrix:
     include:
         - os: linux


snowleopard:
So, we're stuck with this "Ruby" thing? Oh well.

I'll wait for the green light from Travis.

izgzhen:
> So, we're stuck with this "Ruby" thing? Oh well.

Shrug ... I think there is insufficient doc on how the text is composed.

snowleopard:
OK, merged now. Thanks @izgzhen!

Issue #366, opened 2017-07-16
build validate on Linux failed in lack of `test   spaces/hsc2hs`


izgzhen:
```
Cannot find hsc2hs: /tmp/tmpRClrGx/ghc/inplace/test   spaces/hsc2hs.  Stop.
```

stderr log: http://hadrian-status-dev.zhenzhang.me:3000/logs/2017-07-16T12%3A04%3A49.765572%25validate%252139.31077409%25failure.err.log

The weird thing is that the validation works fine on my Mac.

snowleopard:
So, presumably this has something to do with paths containing spaces. One possible test is install GHC into a directory whose name contains spaces. I'll try to do on Windows, as part of #345.

izgzhen:
this is fixed I think

PR #375, opened 2017-07-23
Fix validate's executable dependency

izgzhen:
Addresses https://github.com/snowleopard/hadrian/issues/366

### DIFF
--- a/src/Rules/Test.hs
+++ b/src/Rules/Test.hs
@@ -22,6 +22,8 @@ testRules = do
         needBuilder $ Ghc CompileHs Stage2
         needBuilder $ GhcPkg Update Stage1
         needBuilder Hpc
+        needBuilder Hsc2Hs
+        need ["inplace/bin/hp2ps"] -- TODO: Eliminate explicit filepaths in "need" (#376)
         build $ Target (vanillaContext Stage2 compiler) (Make "testsuite/tests") [] []
 
     "test" ~> do


izgzhen:
The problem can only be reproduced by invoking `validate` only without building top-level target first.

src/Rules/Test.hs
```diff
@@ -22,6 +22,8 @@ testRules = do
         needBuilder $ Ghc CompileHs Stage2
         needBuilder $ GhcPkg Update Stage1
         needBuilder Hpc
+        needBuilder Hsc2Hs
+        need ["inplace/bin/hp2ps"]
```
snowleopard: Can we make this less fragile and compute the path? Alternatively, add a TODO to address this later.

snowleopard:
By the way, there are a few other instances where we `need` filenames directly, so I'm planning to go through the code and solve this issue in general. So leaving a `TODO` is fine.

izgzhen:
> By the way, there are a few other instances where we need filenames directly, so I'm planning to go through the code and solve this issue in general. So leaving a TODO is fine.

I will open an issue and leave a TODO. I think this type of task is relatively easy, well-defined, and not so urgent, maybe you don't have to do it yourself. Leave it for the first-time contributors, if any, would be nice I think.

snowleopard:
@izgzhen Yes, I agree. Thank you!

PR #386, opened 2017-07-31
cabal-install-2.x can new-build

phadej:


snowleopard:
@phadej Thank you! Looks like `PKGVER` is now unused and can be dropped too?

phadej:
I see, yeah, we can drop it.

Btw, cabal-install-2.1 (i.e. the one you get from hvr-ppa) has `new-run`. If we can use it, things will be less ad-hoc (and don't want to support cabal-install-1.24 `new-build`).

snowleopard:
I'd be in favour of switching to `new-run`, but I'm not sure if that's preferable for everyone. Perhaps, we could ask @hvr and @bgamari?

snowleopard:
Thanks, I've merged this and will drop the redundant line.

phadej:
Thanks!

PR #388, opened 2017-08-01
Move need of inplaceLibCopyTargets to top-level

izgzhen:


izgzhen:
BTW, I suspect that we might have run out of CircleCI's plan quota (shrugging ...) I'll ask since we didn't receive any notification about it

src/Rules.hs
```diff
@@ -29,6 +30,7 @@ allStages = [minBound ..]
 -- or Stage1Only targets
 topLevelTargets :: Rules ()
 topLevelTargets = action $ do
+    need inplaceLibCopyTargets
```
snowleopard: `need` is a blocking operation, so it's better to put all targets together and do a single top-level `need` -- see the code below.

snowleopard:
> BTW, I suspect that we might have run out of CircleCI's plan quota (shrugging ...) I'll ask since we didn't receive any notification about it

Yes, please. I haven't received anything from them.

izgzhen:
Updated!

snowleopard:
@izgzhen Thank you, merged!

PR #390, opened 2017-08-17
delete cfg/config.h.in and move generated file to _build

izgzhen:


snowleopard:
@izgzhen Thank you! But I am tempted to either (i) simply delete `cfg/config.h.in` without doing anything else or, better (ii) do The Right Thing. The current PR doesn't look like an improvement: it introduces additional complexity, which at some point will be removed.

So, here is what The Right Thing is:

* Add `hadrian/cfg/config.h` to this section in GHC's `configure.ac`:

    ```bash
    # Create the configuration for the Hadrian build system if it is present
    if test -e hadrian/cfg/system.config.in; then
        AC_CONFIG_FILES([hadrian/cfg/system.config])
    fi
    ```

* Switch to using `hadrian/cfg/config.h` instead of `mk/config.h`.

That's it. In this way we do not depend on any old `mk` file in the Make build system.

@izgzhen Would you like to try this, or shall I attempt this myself? I think I have GHC commit rights, so it may be easier for me.

snowleopard:
Then there is still the question of whether `hadrian/cfg/*` should be moved to the `buildRoot/hadrian/*`. I'm still uncertain of this. One advantage of keeping `configure` output in `hadrian/cfg/*` is that it may potentially allow us not to rerun it on every clean build. After all, the system configuration almost never changes, and the `configure` script is painfully slow (at least on Windows).

izgzhen:
> Would you like to try this, or shall I attempt this myself? I think I have GHC commit rights, so it may be easier for me

Would that mean that we will commit hadrian related code to GHC  before being merged?

I would certainly like to try this since I want to work on this project in the long term

snowleopard:
> Would that mean that we will commit hadrian related code to GHC before being merged?

Yes, it already has Hadrian-related code: 

https://github.com/ghc/ghc/blob/master/configure.ac#L1224-L1227

We'll just need to add one more file into the `AC_CONFIG_FILES` list.

> I would certainly like to try this since I want to work on this project in the long term

Great! Then please proceed by patching `configure.ac` in GHC, so that file `hadrian/cfg/config.h.in` is processed into `hadrian/cfg/config.h`, and then we'll come back to this (or a different) PR.

izgzhen:
I added a phad diff ... https://phabricator.haskell.org/D3869

This is actually the first time I used Phab.

snowleopard:
@izgzhen Thanks! See my Phab comment -- apparently, `mk/config.h.in` is generated, so we definitely need to delete our Hadrian copy, which is probably only relevant to my machine.

Perhaps, we should just follow the approach (i) above: delete `cfg/config.h.in` and keep everything else as is until The Merge.

izgzhen:
According to my judgement in https://phabricator.haskell.org/D3869, it seems impossible to generate `config.h.in` conditionally. So my proposal would be drop the phab diff above, merge this simple deletion instead.

snowleopard:
Thanks @izgzhen! I think this is the right decision given what was discovered in D3869. Merged!

PR #392, opened 2017-08-17
.bat file tidy up plus shake-0.16 compatibility

ndmitchell:
Two things in the patch:

* I cleaned up the `.bat` files - more readable, don't change the directory in the calling shell.
* I added `RuleResult` instances for every oracle, which is required to compile with shake-0.16. I added a compat wrapper for now, but once the new version is released that can be removed.

This patch means the shake release is getting a lot closer.

### DIFF
--- a/build.bat
+++ b/build.bat
@@ -1,4 +1,5 @@
 @echo off
+setlocal
 cd %~dp0
 mkdir bin 2> nul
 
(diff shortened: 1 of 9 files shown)


snowleopard:
@ndmitchell Many thanks! Looks good, so I'll merge after CI passes.

PR #397, opened 2017-08-20
Add DESTDIR to command line arguments

izgzhen:


izgzhen:
After that, I can conveniently setup an `install` test in http://hadrian-status-dev.zhenzhang.me

src/Settings/Packages/Rts.hs
```diff
@@ -62,6 +62,7 @@ rtsPackageArgs = package rts ? do
     ffiIncludeDir  <- getSetting FfiIncludeDir
     ffiLibraryDir  <- getSetting FfiLibDir
     ghclibDir      <- expr installGhcLibDir
+    destDir        <- expr getDestDir
```
snowleopard: Looks like this `destDir` is unused?

snowleopard: Ah, it is used! Disregard this comment.

snowleopard:
@izgzhen Thanks, looks good! I might do some further minor refactoring later, but I'll merge this now.

PR #398, opened 2017-08-22
Install rule: copy gmp header

izgzhen:
Fix:

```
* install
Command: inplace/bin/ghc-cabal copy libraries/integer-gmp /tmp/tmpME8BvT/ghc/_build/stage1/libraries/integer-gmp strip /tmp/tmpME8BvT /usr/local /usr/local/lib/ghc-8.3.20170821 /usr/local/share/doc/ghc-8.3.20170821/html/libraries 'v p dyn'
Exit code: 1
Stderr:
ghc-cabal: can't find include file ghc-gmp.h
)
```

http://hadrian-status-dev.zhenzhang.me/logs/2017-08-22T06%3A48%3A13.572819%25default%255631.28768206%25failure.err.log

### DIFF
--- a/src/Rules/Install.hs
+++ b/src/Rules/Install.hs
@@ -185,6 +185,9 @@ installPackages = do
 
     installLibPkgs <- topsortPackages (filter isLibrary activePackages)
 
+    -- TODO (izgzhen): figure out what is the root cause of the missing ghc-gmp.h error
+    copyFile (pkgPath integerGmp -/- "gmp/ghc-gmp.h") (pkgPath integerGmp -/- "ghc-gmp.h")
+
     forM_ installLibPkgs $ \pkg -> do
         when (isLibrary pkg) $
             withLatestBuildStage pkg $ \stage -> do


snowleopard:
@izgzhen Thanks!

Can you clarify: is this a regression because of some of my paths refactoring, or was it always broken?

snowleopard:
I'm just thinking that maybe this is not the right fix -- perhaps we should actually fix my bug elsewhere?

izgzhen:
> Can you clarify: is this a regression because of some of my paths refactoring, or was it always broken?

I am not sure actually. When I tested `install` rule before on my machine, it is always `integer-simple` plus `quickest`, just to speed things up. But this one runs in `default` with `gmp` on.

> I'm just thinking that maybe this is not the right fix -- perhaps we should actually fix my bug elsewhere?

I try to find in `package-data.mk` that somewhere pointing to `integer-gmp/ghc-gmp.h` rather than the source file, but I can't find it. You suggestion sounds better, though.

snowleopard:
> When I tested install rule before on my machine, it is always `integer-simple` plus `quickest`,

Aha, I see!

Well, what about at least adding a `TODO` to figure out if this hack can be removed later?

izgzhen:
> Well, what about at least adding a TODO to figure out if this hack can be removed later?

Sure

izgzhen:
Updated

snowleopard:
@izgzhen Thanks!

PR #400, opened 2017-08-23
Drop defaultDestDir and use Shake's verbosity to control verbose commands

izgzhen:


snowleopard:
@izgzhen I agree that using Shake's `verbosity` setting to control Hadrian's output is a good idea, but your current proposal is too crude: it removes a very useful feature (at least for me), which allows the user to print only problem-related command lines, e.g. by setting `verboseCommands = package X ? builder Y`.

Also, there was a useful comment on `defaultDestDir`, which is now gone. It should perhaps be moved to the installation section of the `README`, or even better into a installation-specific `doc/install.md`.

izgzhen:
Oh, I see your point. So your point is to **separately** control the verbosity of shake output, and builder output?

snowleopard:
Note that Shake's verbosity setting is already used in some places, e.g. I use `putLoud` in oracles:

https://github.com/snowleopard/hadrian/blob/master/src/Hadrian/Oracles/TextFile.hs#L102

To see this diagnostic info, you need to call Hadrian with `--verbose` option.

However, we also have `--progress-info` command line flag, that controls build output:

https://github.com/snowleopard/hadrian/blob/master/src/Hadrian/Utilities.hs#L284-L290

This is probably a good candidate to be merged with Shake's verbosity setting, but it's not clear how. I think there are no 100% clear solutions here, and since it's not a high-priority issue, I think we shouldn't spend too much effort on this right now.

snowleopard:
But do remove `defaultDestDir`! We just need to preserve the comment somewhere, which is particularly useful for Windows users.

izgzhen:
Okay ... so `verboseCommands` should be be a `Predicate`? If so, then we probably can't put any command line argument in that I guess.

snowleopard:
> so `verboseCommands` should be be a `Predicate`?

Yes, to preserve its current features it needs to be able to access the current build context etc. 

> If so, then we probably can't put any command line argument in that I guess.

What do you mean? We can read command line arguments from within a `Predicate`. For example:

```haskell
verboseCommands = do
    buildHaddock <- expr cmdBuildHaddock
    package <- getPackage
    return $ buildHaddock && (package == haddock)
```

izgzhen:
I see, I forgot the useful `expr`. Thanks!

izgzhen:
Updated

src/UserSettings.hs
```diff
@@ -33,7 +33,9 @@ userKnownPackages = []
 -- this is a 'Predicate', hence you can enable verbose output only for certain
 -- targets, e.g.: @verboseCommands = package ghcPrim@.
 verboseCommands :: Predicate
-verboseCommands = return False
+verboseCommands = do
+    verbosity <- expr getVerbosity
+    return $ verbosity >= Loud
```
snowleopard: Nice! I like this solution.

snowleopard:
@izgzhen Great, thank you!

PR #401, opened 2017-08-26
Add cross compilation

izgzhen:
Tested with arm-linux-gnueabihf using https://gist.github.com/izgzhen/26e49e7bf2e79a9a8f0a3ec37e4837af

### DIFF
--- a/src/Settings.hs
+++ b/src/Settings.hs
@@ -1,7 +1,7 @@
 module Settings (
     getArgs, getLibraryWays, getRtsWays, flavour, knownPackages,
     findPackageByName, getPkgData, getPkgDataList, isLibrary, stagePackages,
-    latestBuildStage, programContext, integerLibraryName, getDestDir, stage1Only
+    latestBuildStage, programContext, integerLibraryName, getDestDir
     ) where
 
 import Context
@@ -69,11 +69,6 @@ latestBuildStage :: Package -> Action (Maybe Stage)
 latestBuildStage pkg = do
     stages <- filterM (fmap (pkg `elem`) . stagePackages) [Stage0 ..]
     return $ if null stages then Nothing else Just $ maximum stages
-
--- TODO: Set this from command line
--- | Stage1Only flag.
-stage1Only :: Bool
-stage1Only = defaultStage1Only
 
 -- | Install's DESTDIR setting.
 getDestDir :: Action FilePath
--- a/src/Settings/Default.hs
+++ b/src/Settings/Default.hs
@@ -34,8 +34,10 @@ import Settings.Packages.Compiler
 import Settings.Packages.Ghc
 import Settings.Packages.GhcCabal
 import Settings.Packages.Ghci
+import Settings.Packages.GhcPkg
 import Settings.Packages.GhcPrim
 import Settings.Packages.Haddock
+import Settings.Packages.Haskeline (haskelinePackageArgs)
 import Settings.Packages.IntegerGmp
 import Settings.Packages.Rts
 import Settings.Packages.RunGhc
@@ -207,4 +209,6 @@ defaultPackageArgs = mconcat
     , integerGmpPackageArgs
     , rtsPackageArgs
     , runGhcPackageArgs
-    , disableWarningArgs ]
+    , disableWarningArgs
+    , ghcPkgPackageArgs
+    , haskelinePackageArgs ]
--- a/doc/cross-compile.md
+++ b/doc/cross-compile.md
@@ -1,0 +1,57 @@
+## Build a cross-compiling GHC
+
+Our host machine is "Ubuntu 16.04.2 LTS, Linux ubuntu 4.4.0-79-generic 86_64".
+
+We need to download necessary tools, including:
+
+- [LLVM-4.0 source](http://releases.llvm.org/4.0.0/llvm-4.0.0.src.tar.xz), you need to build it yourself. Remember to choose release channel and use gold linker (`cmake -DCMAKE_BUILD_TYPE=Release -DLLVM_USE_LINKER=gold ..`)
+- `sudo apt-get install gcc-arm-linux-gnueabihf` to install the GCC cross-compiler
+- Download and install [Haskell Platform 8.0.2](https://haskell.org/platform/download/8.0.2/haskell-platform-8.0.2-unknown-posix--full-x86_64.tar.gz). Install it according to [instructions here](https://www.haskell.org/platform/linux.html#linux-generic)
+
+After all the dependencies are in place:
+
+- `git clone https://github.com/ghc/ghc`
+- `cd ghc`
+- `git clone https://github.com/snowleopard/hadrian`
+- `git submodule update --init`
+- `./configure --target=arm-linux-gnueabihf`
+- `cd hadrian`
+- Modify `src/Settings.hs`, set `stage1Only` and `crossCompiling` to `True`.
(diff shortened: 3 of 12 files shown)


izgzhen:
Still needs refining, but the current design choices are clear by themselves. What is you idea? @snowleopard

README.md
```diff
@@ -90,7 +90,14 @@ a GHC [bug][ghc-split-objs-bug], you need a full clean rebuild when using this f
 * `--verbose`: run Hadrian in verbose mode. In particular this prints diagnostic messages
 by Shake oracles.
 
-#### User settings
+#### Cross Compiling
+
+1. Set up and test as described in https://gist.github.com/izgzhen/26e49e7bf2e79a9a8f0a3ec37e4837af
```
snowleopard: This doc should probably go into `doc/cross-compile.md`

src/GHC.hs
```diff
@@ -34,10 +34,10 @@ defaultKnownPackages =
     [ array, base, binary, bytestring, cabal, checkApiAnnotations, compareSizes
     , compiler, containers, deepseq, deriveConstants, directory, dllSplit
     , filepath, genapply, genprimopcode, ghc, ghcBoot, ghcBootTh, ghcCabal
-    , ghcCompact, ghci, ghcPkg, ghcPrim, ghcTags, haddock, haskeline, hsc2hs
+    , ghcCompact, ghci, ghcPkg, ghcPrim, ghcTags, haddock, hsc2hs
     , hp2ps, hpc, hpcBin, integerGmp, integerSimple, iservBin, libffi
     , mkUserGuidePart, mtl, parsec, parallel, pretty, primitive, process, rts
-    , runGhc, stm, templateHaskell, terminfo, text, time, touchy, transformers
+    , runGhc, stm, templateHaskell, text, time, touchy, transformers
```
snowleopard: What happened with `haskeline` and `terminfo`?

src/Settings.hs
```diff
@@ -98,11 +98,24 @@ programPath context@Context {..} = do
         let installPath = if install then inplaceInstallPath package else path
         return $ installPath -/- programName context <.> exe
 
--- TODO: Set this from command line
```
snowleopard: Do we no longer want to longer want to set this flag from command line?

src/Settings.hs
```diff
@@ -98,11 +98,24 @@ programPath context@Context {..} = do
...
--- | Stage1Only flag.
-stage1Only :: Bool
-stage1Only = defaultStage1Only
-
 -- | Install's DESTDIR setting.
 getDestDir :: Action FilePath
 getDestDir = fromMaybe "" <$> cmdInstallDestDir
+
+crossCompiling :: Bool
+crossCompiling = False
```
snowleopard: We only allow GHC users to edit `UserSettings.hs`, so `crossCompiling` and `stage1Only` flags should be moved there.

src/Settings/Default.hs
```diff
@@ -267,6 +267,9 @@ disableWarningArgs = builder Ghc ? mconcat
                                     , "-Wno-tabs" ]
       , libraryPackage       ? pure [ "-Wno-deprecated-flags" ] ] ]
 
+defaultGhcPkgArgs :: Args
```
snowleopard: According to current conventions, `defaultGhcPkgArgs` should live in `Settings/Packages/GhcPkg.hs`.

src/Utilities.hs
```diff
@@ -252,3 +254,6 @@ putInfo t = putProgressInfo =<< renderAction
     digest [] = "none"
     digest [x] = x
     digest (x:xs) = x ++ " (and " ++ show (length xs) ++ " more)"
+
+contextDependencies' :: Context -> Action [Context]
```
snowleopard: I don't think you need this function. What you need to do is to modify `stage0Packages` in `Settings/Default.hs` removing `terminfo` in case of cross compilation. Then it will automatically be removed from `contextDependencies`.

snowleopard:
@izgzhen Awesome! I've left a few questions and suggestions in the code.

README.md
izgzhen: Ah, good idea

src/GHC.hs
izgzhen: `terminfo` can't be compiled because lack of dependencies on the target platform IIRC. And `haskeline` depends on it. In GHC's tree, you can find `WITH_TERMINFO=NO` in `mk/flavours/quick-cross.mk`, which disables it.

For `haskeline`, it is more or less a quick hack:

```
libraries/haskeline_CONFIGURE_OPTS += --flags=-terminfo
```

I should fix `haskeline` like above rather than turning it completely off. Thanks!

src/Settings.hs
izgzhen: We've already done this in #397, do you mean that?

src/Settings.hs
izgzhen: Emm...sounds good.

src/Settings/Default.hs
izgzhen: right

src/Utilities.hs
izgzhen: Oh, I was trying to find something like that. Thank!

src/GHC.hs
snowleopard: Thanks for the clarification!

src/Settings.hs
snowleopard: I thought we were planning to switch to `stage1Only` build using a command line flag, and this `TODO` was a reminder of that. Maybe I'm wrong and we never planned that, in which case it's fine to remove it.

izgzhen: This, plus `crossCompiling`, is actually quite tricky to handle. Make it compiler time only constants that is set through modifying `UserSettings`, we can use it anywhere. But if it must be passed from command line argument, then it must be wrapped, thus can't mixed with pure code freely.

snowleopard: Aha, interesting. Why is `crossCompiling` difficult to use if it is not a pure Boolean? At the moment I don't see any places where you couldn't use an `Action Bool` instead.

izgzhen: For example, I might want to use crossCompiling to modify `defaultKnownPackages :: [Package]` directly.

You are right in that in the end the build is a monadic action, so ideally refactoring everything to respect `crossCompiling :: Action Bool` is possible. But that might clutter the code too much.

snowleopard: Do you really need to modify `defaultKnownPackages`? It's not supposed to be modified, since it captures all known GHC packages statically. Packages that are actually build by Hadrian are captured by expressions like `stage0Packages` which allow `Action Bool` as conditionals.

izgzhen: Oh, I see

izgzhen:
```
  Key value 3:  hadrian/cfg/system.config settings mk/config.h
  Key value 4:  OracleQ (KeyValue ("hadrian/cfg/system.config","cross-compiling"))
Rules may not be recursive
```

Another problem caused by using `Action Bool` instead of `Bool` for `crossCompiling`....I'll use the pure version for now.

snowleopard:
@izgzhen Sure, let's use `Bool` in the first implementation.

izgzhen:
updated the code

snowleopard:
@izgzhen Thank you, looks good. Shall I merge?

izgzhen:
I would like to merge it as a "just working" version.

I am reviewing the log diff of Make based build in normal and cross
compiling case. But maybe it would be better to discuss the patches
individually in future.

What do you think?

snowleopard:
@izgzhen Sounds good! I've merged this and we can continue via a few more patches.

PR #410, opened 2017-09-01
Refine cross-compilation implementation

izgzhen:
Tested with a more complex example https://gist.github.com/izgzhen/a2d03c333afdb90d3504e127b35f66da

### DIFF
--- a/src/Settings/Flavours/QuickCross.hs
+++ b/src/Settings/Flavours/QuickCross.hs
@@ -1,0 +1,23 @@
+module Settings.Flavours.QuickCross (quickCrossFlavour) where
+
+import Flavour
+import Expression
+import Oracles.Flag
+import {-# SOURCE #-} Settings.Default
+
+quickCrossFlavour :: Flavour
+quickCrossFlavour = defaultFlavour
+    { name        = "quick-cross"
+    , args        = defaultBuilderArgs <> quickCrossArgs <> defaultPackageArgs
+    , libraryWays = mconcat
+                    [ pure [vanilla]
+                    , notStage0 ? platformSupportsSharedLibs ? pure [dynamic] ] }
+
+quickCrossArgs :: Args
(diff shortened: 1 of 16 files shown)


izgzhen:
Still,

```
  Key value 3:  hadrian/cfg/system.config settings mk/config.h
  Key value 4:  OracleQ (KeyValue ("hadrian/cfg/system.config","cross-compiling"))
Rules may not be recursive
```

But I will try to fix this bug in this patch

izgzhen:
Fixed, it should be ready for review now, sorry for the delay!

src/Settings/Flavours/QuickCross.hs
```diff
@@ -0,0 +1,23 @@
+module Settings.Flavours.QuickCross (quickCrossFlavour) where
+
+import Flavour
+import Expression
+import Oracles.Flag
+import {-# SOURCE #-} Settings.Default
+
+quickCrossFlavour :: Flavour
+quickCrossFlavour = defaultFlavour
```
snowleopard: Both `quickCrossFlavour` and `quickCrossArgs` are mostly copied from `quickFlavour`. I wonder if we can/should reduce the duplication. Maybe it doesn't worth it.

@izgzhen What do you think?

snowleopard:
@izgzhen Thanks! It mostly looks fine, apart from a bit of code duplication -- see the above comment.

snowleopard:
I decided to merge this. We might fix the duplication later, although it's unclear if that would be a win or a loss in terms of ease of understanding.

src/Settings/Flavours/QuickCross.hs
izgzhen: I think we should keep this duplication, since it looks more clear what is implicated by `cross` or `quick-cross`, even if part of them overlaps.

izgzhen:
Thanks!

PR #413, opened 2017-09-14
Implement documentation building

patrickdoc:
Hello,

I am working on [Trac 14144](https://ghc.haskell.org/trac/ghc/ticket/14144). I thought it reasonable to revamp the doc building here rather than in the Make-based system. In short, this patch lets you run:

`$ ./build.cabal.sh --haddock docs`

Which will generate:

```
$ tree -L 2 _build/docs
_build/docs/
├── archives
│   ├── Haddock.html.tar.xz
│   ├── libraries.html.tar.xz
│   └── users_guide.html.tar.xz
├── html
│   ├── Haddock
│   ├── libraries
│   └── users_guide
└── pdfs
    ├── Haddock.pdf
    └── users_guide.pdf
```

Where the `libraries` folder holds haddocks and colored sources.

This is not ready quite yet, as there are a couple sticking points that I could use advice on:

1.) I would like to run the Sphinx or Xelatex builders on documentation "packages", like the users' guide at `ghc/docs/users_guide` or the Haddock guide at `ghc/utils/haddock/doc`. However, there is currently no way to represent these as `Package`'s. I tried adding some infrastructure for them into Hadrian, but it felt a little hack-y, in part because there are only two at the moment (ignoring Cabal which has some other issues). However, running the builders directly like I did with Xelatex also doesn't feel like the right way to do it. 

The current version just makes a bogus context for callling the Sphinx builder and manually calls the Xelatex builder, so you can see the difference.

2.) Being forced to use the `--haddock` flag as well as the target `docs` feels wrong to me. I looked, but couldn't tell if it was possible to imply a flag based on a target. Alternatively, I could build in the necessary rules, but then I don't know what would happen if someone did specify `--haddock docs` at the command line.

On a brighter note, some wins of this implementation:

- Haddock's `--hyperlinked-source` completely eliminates the need for HsColour and ghc-cabal's coloring. It also has nicer features.
- The `--read-interface` flag generation is more consistent, and so cross module links in the haddocks are working.
- This eliminates the need for the `libraries/gen_contents_index` script.
- This creates a centralized documentation directory, so distribution and installation will be very easy.
- The dependency on Cabal/ghc-cabal has been greatly reduced, so it should be easier to support building all documentation on any platform. (I also just realized `docs` might not build on Windows because I didn't filter `unix` from the haddocks. We need `package-data.mk` to run the haddocks, but Win32 can't be configured on Unix, and possible vice-versa. I'm going to look into this.)

Hopefully this helps!

### DIFF
--- a/src/Rules/Documentation.hs
+++ b/src/Rules/Documentation.hs
@@ -1,4 +1,4 @@
-module Rules.Documentation (buildPackageDocumentation, haddockDependencies) where
+module Rules.Documentation (buildPackageDocumentation, documentationRules, haddockDependencies) where
 
 import Base
 import Context
@@ -9,6 +9,81 @@ import Oracles.PackageData
 import Settings
 import Target
 import Utilities
+
+docPackage :: Package
+docPackage = hsLibrary "Documentation" "docs"
+
+-- | Build all documentation
+documentationRules :: Rules ()
+documentationRules = do
+    buildHtmlDocumentation
+    buildPdfDocumentation
+    buildDocumentationArchives
+    "docs" ~> do
+        root <- buildRoot
+        -- Archives `need` the associated html
+        let archives = map pathArchive docPaths
+            pdfs = map pathPdf $ drop 1 docPaths
+        need $ map (root -/-) $ archives ++ pdfs
+
+docPaths :: [FilePath]
+docPaths = [ "libraries", "users_guide", "Haddock" ]
+
+docRoot :: FilePath
+docRoot = "docs"
+
+htmlRoot :: FilePath
+htmlRoot = docRoot -/- "html"
+
+pdfRoot :: FilePath
+pdfRoot = docRoot -/- "pdfs"
+
+archiveRoot :: FilePath
+archiveRoot = docRoot -/- "archives"
+
+pathPdf :: FilePath -> FilePath
+pathPdf path = pdfRoot -/- path <.> ".pdf"
+
+pathIndex :: FilePath -> FilePath
+pathIndex path = htmlRoot -/- path -/- "index.html"
+
+pathLibIndex :: FilePath -> FilePath
+pathLibIndex path = htmlRoot -/- "libraries" -/- path -/- "index.html"
+
+pathArchive :: FilePath -> FilePath
+pathArchive path = archiveRoot -/- path <.> "html.tar.xz"
+
+pathPath :: FilePath -> FilePath
+pathPath "users_guide" = "docs/users_guide"
+pathPath "Haddock" = "utils/haddock/doc"
+pathPath _ = "lol"
+
+----------------------------------------------------------------------
+-- HTML
+
+-- | Build all HTML documentation
+buildHtmlDocumentation :: Rules ()
+buildHtmlDocumentation = do
+    mapM_ buildSphinxHtml $ drop 1 docPaths
+    buildLibraryDocumentation
+    "//html/index.html" %> \file -> do
+        root <- buildRoot
+        need $ map ((root -/-) . pathIndex) docPaths
+        copyFileUntracked "docs/index.html" file
+
+-----------------------------
+-- Sphinx
+
+-- | Compile a Sphinx ReStructured Text package to HTML
+buildSphinxHtml :: FilePath -> Rules ()
+buildSphinxHtml path = do
+    "//" ++ htmlRoot -/- path -/- "index.html" %> \file -> do
+        let dest = takeDirectory file
+            context = vanillaContext Stage0 docPackage
+        build $ target context (Sphinx Html) [pathPath path] [dest]
+
+-----------------------------
+-- Haddock
 
 haddockHtmlLib :: FilePath
 haddockHtmlLib = "inplace/lib/html/haddock-util.js"
@@ -25,9 +100,17 @@ haddockDependencies context = do
 -- files in the Shake database seems fragile and unnecessary.
 buildPackageDocumentation :: Context -> Rules ()
 buildPackageDocumentation context@Context {..} = when (stage == Stage1) $ do
-    "//" ++ contextDir context ++ "//*.haddock" %> \file -> do
-        srcs     <- hsSources context
+
+    -- Js and Css files for haddock output
+    when (package == haddock) $ haddockHtmlLib %> \_ -> do
+        let dir = takeDirectory haddockHtmlLib
+        liftIO $ removeFiles dir ["//*"]
+        copyDirectory "utils/haddock/haddock-api/resources/html" dir
+
+    -- Per-package haddocks
+    "//" ++ pkgName package <.> "haddock" %> \file -> do
         haddocks <- haddockDependencies context
+        srcs <- hsSources context
         need $ srcs ++ haddocks ++ [haddockHtmlLib]
 
         -- Build Haddock documentation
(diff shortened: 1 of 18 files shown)


patrickdoc:
Also, all of the pathX functions are meant to be placeholders for pkgX. So if you replace everywhere it says "path" with "pkg", it should look more familiar.

snowleopard:
@patrickdoc This is awesome, thank you! I'll go through the code and will leave a few comments, but my initial impression is that it's pretty much ready to be merged.

src/Rules/Documentation.hs
```diff
@@ -10,6 +10,81 @@ import Settings
...
+documentationRules :: Rules ()
+documentationRules = do
+    buildHtmlDocumentation
+    buildPdfDocumentation
+    buildDocumentationArchives
+    "docs" ~> do
+        root <- buildRoot
+        -- Archives `need` the associated html
+        let archives = map pathArchive docPaths
+            pdfs = map pathPdf $ drop 1 docPaths
```
snowleopard: Perhaps, `drop 1 docPaths` could be replaced with `docPaths \\ [ "libraries" ]` to avoid relying on the order and make it a bit more readable?

Alternatively, `pdfs = map pathPdf ["users_guide", "Haddock"]` doesn't look too bad either.

src/Rules/Documentation.hs
```diff
@@ -10,6 +10,81 @@ import Settings
...
+archiveRoot = docRoot -/- "archives"
+
+pathPdf :: FilePath -> FilePath
+pathPdf path = pdfRoot -/- path <.> ".pdf"
+
+pathIndex :: FilePath -> FilePath
+pathIndex path = htmlRoot -/- path -/- "index.html"
+
+pathLibIndex :: FilePath -> FilePath
+pathLibIndex path = htmlRoot -/- "libraries" -/- path -/- "index.html"
```
snowleopard: This seems to be unused?

src/Rules/Documentation.hs
```diff
@@ -10,6 +10,81 @@ import Settings
...
+
+pathArchive :: FilePath -> FilePath
+pathArchive path = archiveRoot -/- path <.> "html.tar.xz"
+
+pathPath :: FilePath -> FilePath
+pathPath "users_guide" = "docs/users_guide"
+pathPath "Haddock" = "utils/haddock/doc"
+pathPath _ = "lol"
```
snowleopard: Looks like we might want a data type here, e.g. `data Doc = UserGuide | Haddock` or something similar?

snowleopard: Actually, looking at how `pathPath` is used, my suggestion doesn't work. If we leave this as is, we better add a `TODO` to get rid of this hack later.

src/Rules/Documentation.hs
```diff
@@ -10,6 +10,81 @@ import Settings
...
+
+----------------------------------------------------------------------
+-- HTML
+
+-- | Build all HTML documentation
+buildHtmlDocumentation :: Rules ()
+buildHtmlDocumentation = do
+    mapM_ buildSphinxHtml $ drop 1 docPaths
```
snowleopard: See my above comment about `drop 1 docPaths`.

src/Rules/Documentation.hs
```diff
@@ -10,6 +10,81 @@ import Settings
...
+    buildLibraryDocumentation
+    "//html/index.html" %> \file -> do
+        root <- buildRoot
+        need $ map ((root -/-) . pathIndex) docPaths
+        copyFileUntracked "docs/index.html" file
```
snowleopard: Why do we need to keep two copies of the index in the tree?

snowleopard: And why is the copy untracked?

src/Rules/Documentation.hs
```diff
@@ -10,6 +10,81 @@ import Settings
 import Target
 import Utilities
 
+docPackage :: Package
+docPackage = hsLibrary "Documentation" "docs"
```
snowleopard: Not sure yet what to do with this fake package. We might add a `TODO` asking to fix this hack later.

Probably, the right solution is to add a new package type, but that can be done separately.

src/Rules/Documentation.hs
```diff
@@ -36,12 +119,72 @@ buildPackageDocumentation context@Context {..} = when (stage == Stage1) $ do
...
+                            , "-t", "Haskell Hierarchical Libraries"
+                            , "-p", "libraries/prologue.txt" ]
+                            ++ map iface libDocs)
+                           [] [file]
+  where
+    allHaddocks = do
+        pkgs <- stagePackages Stage1
+        sequence [ pkgHaddockFile $ vanillaContext Stage1 pkg
+                 | pkg <- pkgs, isLibrary pkg, isHsPackage pkg
+                                             , pkgName pkg /= "Win32" ]
```
snowleopard: You don't need to filter out the `win32` package. The `stagePackages` action returns the list of packages that are actually built, so it will contain either `win32` or `unix`, depending on the platform.

src/Rules/Documentation.hs
```diff
@@ -36,12 +119,72 @@ buildPackageDocumentation context@Context {..} = when (stage == Stage1) $ do
...
+buildLibraryDocumentation :: Rules ()
+buildLibraryDocumentation = do
+    "//" ++ htmlRoot -/- "libraries/index.html" %> \file -> do
+        haddocks <- allHaddocks
+        need haddocks
+        let libDocs = filter (\x -> takeFileName x /= "ghc.haddock") haddocks
+        let iface haddock = "--read-interface="
+                            ++ (takeFileName . takeDirectory) haddock ++ ","
+                            ++ haddock
+        runBuilder Haddock ([ "--gen-index"
```
snowleopard: Looks like this calls for something like `HaddockMode = BuildPackage | BuildIndex`?

src/Rules/Documentation.hs
```diff
@@ -36,12 +119,72 @@ buildPackageDocumentation context@Context {..} = when (stage == Stage1) $ do
...
+buildSphinxPdf :: FilePath -> Rules ()
+buildSphinxPdf path = do
+    "//" ++ path <.> "pdf" %> \file -> do
+        let context = vanillaContext Stage0 docPackage
+        withTempDir $ \dir -> do
+            -- TODO: Figure out a crutch
+            build $ target context (Sphinx Latex) [pathPath path] [dir]
+            --runBuilderWithCmdOptions [Cwd dir] Xelatex ["-halt-on-error"
+            --                                           , name <.> "tex" ] [] []
+            unit $ cmd Shell [Cwd dir] ["xelatex", "-halt-on-error"
```
snowleopard: I think it should be possible to add a custom build action for `Xelatex` here:

https://github.com/snowleopard/hadrian/blob/master/src/Builder.hs#L149

We can pass the temporary directory via the `outputs` field of the `Target`. For example:
```haskell
            case builder of
                Xelatex -> do
                    cmd Shell [Cwd output] [path] buildArgs
                    cmd Shell [Cwd output] [path] buildArgs
                    cmd Shell [Cwd output] [path] buildArgs
                    cmd Shell [Cwd output] ["makeindex"] (input -<.> "idx")
                    cmd Shell [Cwd output] [path] buildArgs
                    cmd Shell [Cwd output] [path] buildArgs
```
Then this rule becomes:
```haskell
        withTempDir $ \dir -> do
             build $ target context (Sphinx Latex) [pathPath path] [dir]
             build $ target context Xelatex [path <.> "tex"] [dir]
             copyFileUntracked (dir -/- path <.> "pdf") file
```

snowleopard:
@patrickdoc See my comments above. Answering some of your general questions:
* Let's keep the fake doc package for now. We can introduce documentation packages into Hadrian, but I don't think that this should be a blocker. Just leave a `TODO`.
* Regarding `--haddock docs`: we can probably drop the `--haddock` flag and make the `haddock` package lazy: that is, it will be in the list of active packages, but we will not be demanding it at the top-level for the default build rule. It will therefore only be built when we actually run Haddock from Hadrian, i.e. when Hadrian is run with the `docs` target.

src/Rules/Documentation.hs
patrickdoc: The file is auto-generated from docs/index.html.in, and so is set from configure time. As it happens, the ticket above asked for some upgrades for that page. After I get to those, this might be a static file that will be easier to handle.

src/Rules/Documentation.hs
patrickdoc: Ah, you are right. When I was checking if I could support `Win32`, I switched to `ghcPackages` instead of `stagePackages`. I seem to have missed that fix when I reverted that.

patrickdoc:
These all sound good to me. I'll clean it up and update.

snowleopard:
@patrickdoc Great, looking forward to it!

snowleopard:
@patrickdoc Thanks, the latest commit looks good. Is the PR ready to be merged?

patrickdoc:
@snowleopard Yep, should be good to go.

snowleopard:
@patrickdoc Thank you, merged!

PR #417, opened 2017-09-20
Fix Config.hs to import GhcPrelude

izgzhen:
simple fix reflecting upstream break change: https://github.com/ghc/ghc/commit/f63bc730c7ea42ca6882f8078eb86be8bf1cc6ad


snowleopard:
@izgzhen Thank you, merged!

Issue #406, opened 2017-08-31
Implement commented part in Rts


izgzhen:
See https://github.com/snowleopard/hadrian/blob/29046aa3b29a26e69db1bf38015f9376bfad2ff0/src/Settings/Packages/Rts.hs#L127-L258

snowleopard:
It's probably medium-to-high priority. I expect that these missing flags account for a lot of errors we see when we run `validate`.

izgzhen:
I'll close this now

PR #418, opened 2017-09-21
Complete RTS compiler args

izgzhen:
solve https://github.com/snowleopard/hadrian/issues/406

### DIFF
--- a/src/Settings/Packages/Rts.hs
+++ b/src/Settings/Packages/Rts.hs
@@ -20,6 +20,15 @@ rtsBuildPath = buildPath rtsContext
 rtsConfIn :: FilePath
 rtsConfIn = pkgPath rts -/- "package.conf.in"
 
+-- | Minimum supported Windows version.
+-- These numbers can be found at:
+-- https://msdn.microsoft.com/en-us/library/windows/desktop/aa383745(v=vs.85).aspx
+-- If we're compiling on windows, enforce that we only support Vista SP1+
+-- Adding this here means it doesn't have to be done in individual .c files
+-- and also centralizes the versioning.
+rtsWindowsVersion :: String
+rtsWindowsVersion = "0x06000100"
+
 rtsLibffiLibraryName :: Action FilePath
 rtsLibffiLibraryName = do
     useSystemFfi <- flag UseSystemFfi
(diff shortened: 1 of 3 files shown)


izgzhen:
@snowleopard you mentioned its relevance with validation errors -- so I will try to get some validate result at the same time when you review this code as well.

izgzhen:
current master:

```
SUMMARY for test run started at Thu Sep 21 20:15:33 2017 HKT
 0:08:11 spent to go through
    6078 total tests, which gave rise to
   14959 test cases, of which
    9003 were skipped

      40 had missing libraries
    5449 expected passes
     158 expected failures

       0 caused framework failures
       0 caused framework warnings
       3 unexpected passes
     223 unexpected failures
      83 unexpected stat failures
```

this patch:

```
SUMMARY for test run started at Thu Sep 21 19:37:30 2017 HKT
 0:06:10 spent to go through
    6078 total tests, which gave rise to
   14959 test cases, of which
    9003 were skipped

      40 had missing libraries
    5448 expected passes
     158 expected failures

       0 caused framework failures
       0 caused framework warnings
       3 unexpected passes
     224 unexpected failures
      83 unexpected stat failures
```

Not much difference though

izgzhen:
BTW, I was working around the ambiguous `RuleResult` error in CI at another branch. Weird that I can't reproduce this on my machine -- is it caused by different compiler version?

izgzhen:
Oh .. I guess it is because Shake has just updated new version 0.16 and the CI build includes it.

src/Settings/Packages/Rts.hs
```diff
@@ -29,13 +38,67 @@ rtsLibffiLibraryName = do
...
     suf     <- libsuf way
     rtsPath <- rtsBuildPath
     return $ rtsPath -/- "lib" ++ name ++ suf
 
+-- ref: mk/config.mk.in
+ghcRtsWithLibDw :: Action Bool
+ghcRtsWithLibDw = do
+    goodArch <- anyTargetArch ["i386", "x86_64"]
+    withLibDw <- setting WithLibDw
+    return $ goodArch && withLibDw == "YES"
```
snowleopard: It looks like `WithLibDw` should be a `Flag`, not a `Setting`, so the comparison with `"YES"` is taken care of automatically.

src/Settings/Packages/Rts.hs
```diff
@@ -122,137 +212,7 @@ rtsPackageArgs = package rts ? do
...
--- # for details
-
--- # Without this, thread_obj will not be inlined (at least on x86 with GCC 4.1.0)
--- ifneq "$(CC_CLANG_BACKEND)" "1"
--- rts/sm/Compact_CC_OPTS += -finline-limit=2500
--- endif
+               , "-DINCLUDE_DIR=\"" ++ destDir ++ ghclibDir -/- "include\"" ]
+        , builder HsCpp ? mconcat
+            [ ghcRtsWithLibDw ? arg "-DUSE_LIBDW"
+            , (== "YES") <$> setting HaveLibMingwEx ? arg "-DHAVE_LIBMINGWEX" ] ]
```
snowleopard: `HaveLibMingwEx` should also be a `Flag`, not a `Setting`.

src/Settings/Packages/Rts.hs
```diff
@@ -29,13 +38,67 @@ rtsLibffiLibraryName = do
...
 
+-- ref: mk/config.mk.in
+ghcRtsWithLibDw :: Action Bool
+ghcRtsWithLibDw = do
+    goodArch <- anyTargetArch ["i386", "x86_64"]
+    withLibDw <- setting WithLibDw
+    return $ goodArch && withLibDw == "YES"
+
+ccClangBackend :: Action Bool
+ccClangBackend = (== "1") <$> setting CcClangBackend
```
snowleopard: Ohh, is `"1"` here equivalent to `"YES"`? Or is this a version number?

snowleopard:
@izgzhen Great, thanks for clearing this up! I've left a couple of suggestions to keep the comparison of flag-like settings to `"YES"` in one place.

src/Settings/Packages/Rts.hs
snowleopard: If this is a version, I suggest we rename this to something like `ccClangBackendVersionEq1` to distinguish it from YES/NO settings.

src/Settings/Packages/Rts.hs
izgzhen: Oh I see, thanks!

src/Settings/Packages/Rts.hs
izgzhen: I think it is using "1" to represent "YES", and "0" for "NO".  Can we simply use `GccIsClang` here?

izgzhen: See https://github.com/ghc/ghc/blob/master/aclocal.m4#L1250

snowleopard: @izgzhen Yes, it looks like we can use `GccIsClang` instead. Perhaps we could even drop `CcClangBackend` altogether?

izgzhen: Indeed, I was thinking about doing that

izgzhen:
cleaned up the bits mentioned

snowleopard:
Thanks @izgzhen! Merged.

PR #419, opened 2017-09-21
Update shake and add stm to stage1 packages

izgzhen:


stack.yaml
```diff
@@ -8,6 +8,9 @@ packages:
 - '.'
 - '../libraries/Cabal/Cabal'
 
+extra-deps:
+- shake-0.16
+
```
snowleopard: Shall we switch to a later Stackage snapshot instead? I fear we might forget to remove this bit later when we switch to a new LTS.

snowleopard:
Thanks @izgzhen! Let's see if it works across CI bots.

P.S. I'm travelling, so it may take me some time to review the other PR.

hadrian.cabal
```diff
@@ -124,7 +124,7 @@ executable hadrian
                        , extra                >= 1.4.7
                        , mtl                  == 2.2.*
                        , QuickCheck           >= 2.6 && < 2.10
-                       , shake                >= 0.15.6
+                       , shake                == 0.16
```
ndmitchell: Suggest either >= 0.16 or 0.16.*. It should allow 0.16.1.

izgzhen: Sure, thanks

stack.yaml
izgzhen: I'll take a look

izgzhen: the latest LTS-9.5 doesn't support this yet -- I think we can open up an issue so we don't forget this when a new version is rolled out

izgzhen:
there is another change https://github.com/ghc/ghc/commit/02ff70563e490d2a7f3141eab7229803c523da57 that I have to fix, so included as another commit in this PR as well

snowleopard:
@izgzhen Thank you, I'm merging this, although it looks like Windows build is broken by further changes in the GHC HEAD.

Issue #246, opened 2016-05-10
Build the User's Guide + man pages


thomie:
See `docs/users_guide/ghc.mk` and `rules/sphinx.mk`.

snowleopard:
Shall this go into the list in #239?

@thomie You are on fire today -- many thanks for your help :+1:

jmct:
There is a ghc-trac ticket that should probably be kept in mind on this which is about being able to build the user guide independently [here](https://ghc.haskell.org/trac/ghc/ticket/12619).

I am not sure if you’d like this independence to be part of the minimum viable product described in #239 or not.

snowleopard:
@jmct Thanks! I think it would be implement independent build rules for documentation, as described in the ticket you linked.

I don't think this should be on the critical path, but it's still an important feature to work on.

snowleopard:
By the way, I just noticed that Hadrian currently builds `mkUserGuidePart` in Stage0. Presumably this is wrong, but `mkUserGuidePart.cabal` doesn't set any restrictions on the version of the `ghc` library, hence the build succeeds. If `mkUserGuidePart` must have access to the latest in-tree `DynFlags` then I guess this should somehow be reflected in the `cabal` file?

@bgamari Shall I move `mkUserGuidePart` to Stage2 as in the Make build system?

snowleopard:
Ah, `mkUserGuidePart` was actually moved to Stage0 by @bgamari's commit: https://github.com/snowleopard/hadrian/commit/a86f2b1e97fb7fa0ef08327f083049a41b278513.

So, I'll leave `mkUserGuidePart` where it is for now :)

bgamari:
Andrey Mokhov <notifications@github.com> writes:

> By the way, I just noticed that Hadrian currently builds `mkUserGuidePart` in Stage0. Presumably this is wrong, but `mkUserGuidePart.cabal` doesn't set any restrictions on the version of the `ghc` library, hence the build succeeds. If `mkUserGuidePart` must have access to the latest in-tree `DynFlags` then I guess this should somehow be reflected in the `cabal` file?
>
> @bgamari Shall I move `mkUserGuidePart` to Stage2 as in the Make build system?
>
Or stage1, either way should work. Thanks for noticing!

I'll add a tighter bound on `ghc`.

snowleopard:
> Or stage1, either way should work.

@bgamari I guess Stage1 would be preferable as it will then be a lot less overhead to build user guide only?

[...]

Issue #414, opened 2017-09-16
Finalise documentation build rule


snowleopard:
After merging #413, we can build documentation in various formats. The remaining two tasks are:
- [x] At the moment we need to provide both `--haddock` and `docs` in the command line to build docs. I think we can probably drop the `--haddock` flag and make the `haddock` package lazy: that is, it will be in the list of active packages, but we will not be demanding it at the top-level for the default build rule. It will therefore only be built when we actually run the Haddock builder from Hadrian, i.e. when Hadrian is run with the `docs` target.
- [x] We also need to update the README.
- [x] Fix the `man` page, see #424.

PR #424, opened 2017-09-30
Build man page

izgzhen:
It addresses #246

### DIFF
--- a/src/Settings/Builders/Sphinx.hs
+++ b/src/Settings/Builders/Sphinx.hs
@@ -16,7 +16,7 @@ sphinxBuilderArgs = do
                 , arg =<< getInput
                 , arg outPath ]
             , builder (Sphinx Man) ? mconcat
-                [ arg "-b", arg "latex"
+                [ arg "-b", arg "man"
                 , arg "-d", arg $ outPath -/- ".doctrees-man"
                 , arg =<< getInput
                 , arg outPath ] ]
--- a/src/Rules/Documentation.hs
+++ b/src/Rules/Documentation.hs
@@ -22,12 +22,17 @@ documentationRules = do
     buildHtmlDocumentation
     buildPdfDocumentation
     buildDocumentationArchives
+    buildManPage
     "docs" ~> do
         root <- buildRoot
         let html = htmlRoot -/- "index.html"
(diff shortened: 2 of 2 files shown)


izgzhen:
The major concern is `buildUsersGuideHtml`: I found the current implementation of `Sphinx` builder can't quite elegantly map to the `sphinx` function as in `rules/sphinx.mk`, so I need to use `cmd`.

snowleopard:
@izgzhen Wait, have you seen this: #413? Aren't you duplicating the functionality?

izgzhen:
> Wait, have you seen this: #413? Aren't you duplicating the functionality?

I thought that PR hasn't implemented the users_guide and man page yet, does it?

bgamari:
> I thought that PR hasn't implemented the users_guide and man page yet, does it?

I believe it does. For instance, grep for `man` in the diff.

snowleopard:
@izgzhen It definitely builds the User's Guide. As far as I understand, documentation is mostly done, we only miss a few bits listed here: #414. 

I've now closed #246 to avoid confusion.

src/Settings/Builders/Sphinx.hs
```diff
@@ -16,7 +16,7 @@ sphinxBuilderArgs = do
                 , arg =<< getInput
                 , arg outPath ]
             , builder (Sphinx Man) ? mconcat
-                [ arg "-b", arg "latex"
+                [ arg "-b", arg "man"
```
snowleopard: This does look like the right change though.

@patrickdoc Was this a typo?

patrickdoc: Ah, yes, my apologies. That should indeed be `man` not `latex`.

izgzhen:
Ah, I see. I've confirmed that `users_guide` is indeed built -- but it still looks that man page is missing.

snowleopard:
> but it still looks that man page is missing

@izgzhen Thanks, I've added this into the todo list in #414. Hopefully, we can fix it in this PR!

izgzhen:
revised

snowleopard:
@izgzhen Thanks! Looks good. Broken CI is due to #426.

Issue #425, opened 2017-09-30
"Rules may not be recursive" error


bgamari:
While attempting to build GHC e515c7f37be97e1c2ccc497ddd0a730e63ddfa82 with Hadrian c6d7b2a33e6ff987e7112c57555425c285c380e9 I encountered the following error,
```
$ hadrian/build.cabal.sh -j
Up to date
/-------------------------------------------------------\
| Run Configure "." (stage = Stage0, package = ghc)     |
|      input: hadrian/cfg/system.config.in (and 2 more) |
|  => output: hadrian/cfg/system.config (and 2 more)    |
\-------------------------------------------------------/
shakeArgsWith   0.000s    0%                           
Function shake  0.319s   65%  =========================
Database read   0.001s    0%                           
With database   0.000s    0%                           
Running rules   0.168s   34%  =============            
Total           0.488s  100%                           
Error when running Shake build system:
* OracleQ (ConfigKey "host-os")
* hadrian/cfg/system.config
* hadrian/cfg/system.config settings mk/config.h
* OracleQ (ConfigKey "host-os")
Build system error - recursion detected:
  Key type:   OracleQ ConfigKey
  Key value:  OracleQ (ConfigKey "host-os")
Rules may not be recursive


```

snowleopard:
@bgamari Oh. I'll try to reproduce and fix asap.

snowleopard:
@bgamari I couldn't reproduce the "Rules may not be recursive" error. Can you try if Hadrian 88a7b4e31616f06ea9c0f75d3565ae11936009e0 has the same problem?

By the way, it looks like the build is broken by recent `libffi` changes. I'll raise another issue.

snowleopard:
@bgamari After fixing the `libffi` issue I still can't reproduce this. Is the build still broken for you?

bgamari:
Indeed I still encounter this with ghc/ghc@429fafb5b2b1ac02fd04d9a98e30b5991125692c and Hadrian ca6ab27204bedc59716040da849a47ef64fcafa9.

snowleopard:
@bgamari That's puzzling.

Do you happen to have a custom `UserSettings.hs` file?

I know what can cause this: trying to compute a setting like `windowsHost` before the `configure` script had a chance to run and collect the information about the system. But as far as I can see, in such situations (when `configure` output is not yet available), we are using `System.Info.os` to determine the OS.

bgamari:
[...]

PR #385, opened 2017-07-28
Disable parallel garbage collection

bgamari:
This brings productivity from roughly 40% to 95%. With parallel GC we generally
spend much of our time synchronizing between the GC threads and relatively
little time doing productive work since we spend most of our time waiting for external processes and generate relatively little garbage.

PR #427, opened 2017-10-10
build.cabal.sh: Use cabal new-run

bgamari:
The previous approach was terribly unreliable, leading me to waste an hour
debugging #425.

### DIFF
--- a/src/Hadrian/Utilities.hs
+++ b/src/Hadrian/Utilities.hs
@@ -119,6 +119,7 @@ unifyPath = toStandard . normaliseEx
 
 -- | Combine paths with a forward slash regardless of platform.
 (-/-) :: FilePath -> FilePath -> FilePath
+_ -/- b@('/':_) = b -- don't prefix if second argument is already absolute
 "" -/- b = b
 a  -/- b
     | last a == '/' = a ++       b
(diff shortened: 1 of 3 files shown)


build.cabal.sh
```diff
@@ -48,7 +48,7 @@ if [ "${CABVER[0]}" -eq 2 -o "${CABVER[0]}" -eq 1 -a "${CABVER[1]}" -ge 24 ]; th
 
     # there's no 'cabal new-run' yet, but it's easy to emulate
```
snowleopard: Shall this comment be dropped then?

snowleopard:
@bgamari Thanks! Travis failed though: https://travis-ci.org/snowleopard/hadrian/jobs/286173747

Travis uses `build.cabal.sh`, so it looks like this change broke it?

bgamari:
Fixed.

Also note that the change in #385 wasn't quite right, which I fixed in the second commit. However, I would be happy to move this to a new PR if you would prefer.

snowleopard:
@bgamari I'm fine to keep the `rtsopts` bit in this PR.

However, I don't see the commit fixing the `new-run` problem with the CI -- it still fails!

bgamari:
Huh, odd; my impression that the reason for the previous failure was that I was missing a `--` in the command line. Apparently not.

bgamari:
Hmm, I wonder if this is due to an outdated `cabal-install`.

bgamari:
Alright, well until I get a chance to sort this out I've split out the rts options fix into #429.

snowleopard:
@bgamari Apologies for messing up the PR with a conflicting commit (#430).

bgamari:
Continuing in #435.

PR #430, opened 2017-10-11
Rearrange unix build scripts.

duog:
Addresses Issue #428 

This isn't identifying the renames properly, sorry about that.

build.stack.nix.sh delegates to build.stack.sh with an environment variable, this seemed simpler than trying to work out how to use getopts.

PR #435, opened 2017-10-12
build.cabal.sh: Use cabal new-run

bgamari:
The previous approach was terribly unreliable, potentially picking up stale executables which lead me to waste an hour debugging #425.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -46,9 +46,8 @@ if [ "${CABVER[0]}" -eq 2 -o "${CABVER[0]}" -eq 1 -a "${CABVER[1]}" -ge 24 ]; th
     # New enough cabal version detected, so
     # let's use the superior 'cabal new-build' mode
 
-    # there's no 'cabal new-run' yet, but it's easy to emulate
     "$CABAL" new-build --disable-profiling --disable-documentation -j exe:hadrian
-    $(find ./dist-newstyle -type f -name hadrian | head -n 1) \
+    "$CABAL" new-run -- hadrian        \
         --lint                         \
         --directory "$absoluteRoot/.." \
         "$@"


snowleopard:
@bgamari Wait, something is wrong. Looks like we messed things up in #430.

Could you do this change to `build.sh` instead? `build.global-db.sh` should actually be a direct invocation of GHC like here: 

https://github.com/snowleopard/hadrian/blob/10b8358882867ebfef0a48b9ba28d08fcf37eedb/build.sh

And `build.sh` is the default build script, which is Cabal-based, so your commit should change it. I'll fix `build.global-db.sh`.

bgamari:

> @bgamari Wait, something is wrong. Looks like we messed things up in #430.
>
> Could you do this change to `build.sh` instead? `build.global-db.sh` should actually be a direct invocation of GHC like here: 
>
> https://github.com/snowleopard/hadrian/blob/10b8358882867ebfef0a48b9ba28d08fcf37eedb/build.sh
>
> And `build.sh` is the default build script, which is Cabal-based, so your commit should change it. I'll fix `build.global-db.sh`.
>
Whoops. Thanks for catching this; I should have looked at the result of
the rebase.

snowleopard:
@bgamari Thanks, merged now!

snowleopard:
Oops, just realised this broke Travis... :-S Will try to fix.

snowleopard:
@bgamari I think the above commit is what you meant.

snowleopard:
Still broken :-S No idea why it's still:
```
cabal: Unknown target '--lint'.
There is no component '--lint'.
The project has no package '--lint'.
```

PR #436, opened 2017-10-12
[WIP] Cross compiling fixes

bgamari:
Here are a few patches fixing some bugs I encountered while trying to build a cross-compiler. In essence our `gmp` configuration wasn't quite aligned with what the `make`-based build system did.

### DIFF
--- a/src/Settings/Builders/Configure.hs
+++ b/src/Settings/Builders/Configure.hs
@@ -9,10 +9,10 @@ configureBuilderArgs = do
     gmpPath    <- expr gmpBuildPath
     libffiPath <- expr libffiBuildPath
     mconcat [ builder (Configure gmpPath) ? do
-                hostPlatform  <- getSetting HostPlatform
                 buildPlatform <- getSetting BuildPlatform
+                targetPlatform <- getSetting TargetPlatform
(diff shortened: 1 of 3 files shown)


src/Settings/Builders/Configure.hs
```diff
@@ -9,10 +9,10 @@ configureBuilderArgs = do
     gmpPath    <- expr gmpBuildPath
     libffiPath <- expr libffiBuildPath
     mconcat [ builder (Configure gmpPath) ? do
-                hostPlatform  <- getSetting HostPlatform
                 buildPlatform <- getSetting BuildPlatform
```
Ericson2314: It would be nice to compute these per stage, so we don't need to do the unintuive renaming below.

snowleopard:
@bgamari Do you expect to bring more commits here (I judge by the `WIP` label)? @angerman is doing a major refactoring in another PR and there is a danger that this PR might get outdated and lost.

bgamari:
Well, I don't believe I ever managed to get this building to my satisfaction. I'll try to have a look tonight.

bgamari:
I believe these should be no longer necessary.

PR #441, opened 2017-10-24
Fix CI

snowleopard:
Trying to finally fix all CI


snowleopard:
Aha, looks like this worked.

@ndmitchell Trying to push via PRs as you suggested. Especially helpful when tweaking CI :)

Issue #379, opened 2017-07-26
Implement install_docs in install rule


izgzhen:
This is blocked on #246.

izgzhen:
done in https://github.com/snowleopard/hadrian/commit/7a85ccf58a5fdd82fa44cc0a2703c4bc621211e1

PR #442, opened 2017-10-25
Implement install_docs

izgzhen:
fixes issue #379

### DIFF
--- a/src/Rules/Documentation.hs
+++ b/src/Rules/Documentation.hs
@@ -23,12 +23,16 @@ documentationRules = do
     buildPdfDocumentation
     buildDocumentationArchives
     buildManPage
+    "//docs//gen_contents_index" %> copyFile "libraries/gen_contents_index"
+    "//docs//prologue.txt" %> copyFile "libraries/prologue.txt"
     "docs" ~> do
         root <- buildRoot
         let html = htmlRoot -/- "index.html"
             archives = map pathArchive docPaths
             pdfs = map pathPdf $ docPaths \\ [ "libraries" ]
         need $ map (root -/-) $ [html] ++ archives ++ pdfs
+        need [ root -/- htmlRoot -/- "libraries" -/- "gen_contents_index" ]
+        need [ root -/- htmlRoot -/- "libraries" -/- "prologue.txt" ]
         need [manPagePath]
 
 manPagePath :: FilePath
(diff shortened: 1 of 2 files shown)


src/Rules/Documentation.hs
```diff
@@ -23,12 +23,16 @@ documentationRules = do
     buildPdfDocumentation
     buildDocumentationArchives
     buildManPage
+    "*//docs//gen_contents_index" %> \out -> copyFile "libraries/gen_contents_index" out
+    "*//prologue.txt" %> \out -> copyFile "libraries/gen_contents_index" out
```
snowleopard: I'm confused by the file patterns above: why not `//docs/gen_contents_index` and `//prologue.txt`?

snowleopard: Also, not a big deal, but there is a slight simplification: `\out -> copyFile "libraries/gen_contents_index" out` can be replaced with `copyFile "libraries/gen_contents_index"`.

src/Rules/Install.hs
```diff
@@ -311,3 +312,22 @@ installIncludes = do
...
+
+installDocs :: Action ()
+installDocs = do
+    destDir <- getDestDir
+    docDir  <- installDocDir
+    root    <- buildRoot
+    installDirectory (destDir ++ docDir)
+
+    let usersGuide = root -/- "docs/pdfs/users_guide.pdf"
+    whenM (liftIO (IO.doesFileExist $ usersGuide)) $ installData [usersGuide] (destDir ++ docDir)
```
snowleopard: Any particular reason to use non-Shake `doesFileExist`? We might as well track the existence `usersGuide`.

snowleopard:
Thanks @izgzhen! I've left a couple of minor comments.

src/Rules/Install.hs
izgzhen: Oh, I see. I will use `need` instead

src/Rules/Documentation.hs
izgzhen: Fixed, thanks!

izgzhen:
BTW, I've spotted another problem (https://github.com/snowleopard/hadrian/issues/443), but it doesn't necessarily block the `install_docs`

src/Rules/Install.hs
izgzhen: > I think you can just use Shake's doesFileExist and it will rerun this build rule if its result changes.

Right. This is still the right approach to go. I mean -- if `docs` related things are not built in the first place, then `install` shouldn't `need` them.

izgzhen:
revised

snowleopard:
@izgzhen Looks fine now, thank you!

PR #445, opened 2017-11-01
WIP: Relocation based build (inplace free)

angerman:
This is essentially a reorganization on how libraries and ghc are built. The idea is to produce
ghc builds in `_build/stage2/{bin,lib}`, that could just be packed up and shipped around. This
also means there should be no more `inplace` files left.

### This also includes a few additional changes:
- gets rid of the Evac_thr, Scar_thr logic, that is not present in the Make base build system
- make the cabal file parser a bit more powerful, such that it can `configure` packages, (it does not run the proper configure logic yet).

### Things that do now yet work:
- There are no wrapper scripts.
- `ghc --interactive` fails to run, due to libgmp relocation issues
   **Note** this was due to hadrian relying on builing the in-tree-gmp on macOS. I believe
    this is equally broken in the make based build system.
- No target prefixes yet. (e.g. `arch-vendor-os-ghc`).

### Noteworthy:
- Requires patches to Cabal (haskell/cabal#4857, haskell/cabal#4874, haskell/cabal#4875)
   **Note**: this is now an collection of serval PRs, which are included in my [zw3rk/cabal/work](https://github.com/zw3rk/cabal/tree/work) branch.
- Requires changes to GHC ([D4121](https://phabricator.haskell.org/D4121), [D4135](https://phabricator.haskell.org/D4135), [D4174](https://phabricator.haskell.org/D4174), [D4175](https://phabricator.haskell.org/D4175), [D4176](https://phabricator.haskell.org/D4176))

### Others:
The value of this has definitely been that I now know more about hacking on Hadrian.

**NOTE**: I do *not* expect this to be merged now.  I want this to be more a basis
for discussion. I would have tried to make the same idea work with the make based
build system, but as Hadrian is the supposed way forward, I did not intend to waste
much time on the old(?) build system. I am going to pursue this direction further.

### TODO:
- [ ] Package Install / Registration Targets (stage+1). E.g. a library built with stage0 should be registers in the packageDb at stage1. Similar to how Programs built in stage0 end up in `stage1/bin`. This will also remove the need to copy the package.conf.d over from stage1 to stage2 to obtain the full stage2 compiler.
- [ ] Reinstate wrapper rules
- [x] Drop the custom bootstrap db logic. (it's just the stage1 db then).
- [x] Reinstate the Install rules.
- [ ] Check running the test suite.
- [ ] Check building documentation.
- [x] Roll more of ghc-cabal into hadrian.


snowleopard:
@angerman Oh, wow, this is a huge PR, many thanks! I had a very brief look at it and everything looks fine. I'll add a few specific comments to the code a bit later.

Is there are general agreement that GHC should get rid of `inplace` logic in future? I hope we won't have to support both approaches.

angerman:
I don't know if there is an agreement or not. I face practical issues from `inplace`. For one building cross-compiler binary distributions does not work properly with the current build system, and instead of sinking even more time into trying to make that work, I'd rather sink time into Hadrian if this is what is likely going to be the future build system.

The other is, that I oftentimes want to build multiple different configurations of form the same source, as such I *really* want out-of-tree builds. Hadrian get's us there a bit already. I'd still need different build roots for different configurations, and get rid of all the other in-place modifications.

To tackle the cross compiler bindist issues, I'm trying to make the build system produce a relocatable installation (+ wrapper scripts as fallback). This requires changes to ghc, cabal and hadrian. But the
general idea is to have cleaner stage separation and less custom logic for "special" packages.

For the full out-of-tree builds, I don't have a clear idea yet. I'll have to look into autotools for this as well, as any autoconf generated file (or any generated file for that matter) must not end up in the source-tree.

As such, expect this PR to even grow a bit more.

My take is that, if we are going to change the build system, maybe we should rethink a few of approaches the make base build system took, and see if we can clean them up or simplify them. This might mean we need changes to the tools that are used to build ghc (e.g. cabal).

Also, it might provide a compelling reason to use Hadrian, if it does things differently and provides a benefit over the current make based build system.

snowleopard:
@angerman Thanks, this all sounds good to me. It's a very ambitious plan, but the benefits are very clear.

> The value of this has definitely been that I now know more about hacking on Hadrian.

I'd like to ask a question: If you had to implement this change in the Make build system with which you are already familiar, would it take you less time, more time, or roughly the same time, because the complexity is in the task itself and not the way the build system is written?

src/Builder.hs
```diff
@@ -111,22 +121,25 @@ instance NFData   Builder
 -- 'Stage' and GHC 'Package').
 builderProvenance :: Builder -> Maybe Context
 builderProvenance = \case
-    DeriveConstants  -> context Stage0 deriveConstants
-    GenApply         -> context Stage0 genapply
-    GenPrimopCode    -> context Stage0 genprimopcode
+    DeriveConstants  -> context Stage1 deriveConstants
```
izgzhen: `Stage0` to `Stage1`? What does this imply?

angerman: The general theme is that what is build in `stageN` is placed in `stage(N+1)` (this does hold for all binaries, but not yet for the package.db, which I'll need to fix.

The result is that every binary is built in `stageN` is in `stage(N+1)/bin` and any library built in `stageN` in `stage(N+1)/lib`, as a result building the binary distribution for stageN is simply zipping the `stageN/{bin, lib}` folders.

For some tools, which are used only during the build, where we know we do not need to rebuild them with the stage1 compiler, and the ones built with the bootstrap (stage0) compiler are good, we can fix the location.

So the stage0 -> stage1 changes are mostly due to the changed layout of where binaries end up.

angerman:
@snowleopard 
> I'd like to ask a question: If you had to implement this change in the Make build system with which you are already familiar, would it take you less time, more time, or roughly the same time, because the complexity is in the task itself and not the way the build system is written?

That's a pretty tough question, which is hard to answer in a meaningful way. I do not have any estimates how long it would have taken with the make based build system. I might not even have tried such an ambitious change. The reason being that the primary benefit of Haskell I see is that it is rather easy to refactor code.  With the make based build system you are often not sure you actually have caught all the cases you change might have influenced. That's the benefit of using haskell for a build system to me.

Now while working with hadrian one of the first confusing parts has been the builder logic. While the abstraction makes it rather pleasant to work with the builders, **after** one understand how they actually work, and where all the flags come from, it also adds a huge layer of abstraction. Which (as we see with the `-DPARSEC` for Cabal, makes some things rather hard). This is something where I'm not sure the abstraction is just right yet.

There are of course other changes that this PR includes that are simply not possible within the make based build system. Hadrian by linking against Cabal has access to the cabal library and can as such read and configure cabal files, where the make based build system relies on ghc-cabal to make cabal file information available to the make based build system.

src/Builder.hs
snowleopard: > The general theme is that what is build in `stageN` is placed in `stage(N+1)`

@angerman But this is not how Hadrian does things now. The `_build/stage0` folder contains all build artefacts build by Stage0 compiler (i.e. by the bootstrapping compiler), including binaries.

I think this is a simple and logical arrangement. Why do you want to place binaries built in Stage0 into `_build/stage1`?

snowleopard:
@angerman Thank you, that's very helpful.

Regarding the abstractions we set up for constructing command lines: I agree, this is a non-trivial part and I am also not sure that the abstraction we have at the moment is optimal. We might change it in future if there is a better one -- as you say, with Haskell it's much easier to do refactoring of such scale.

src/Builder.hs
snowleopard: In particular, I think the Stage1 GHC binary should go somewhere into `_build/stage0/ghc` directory, because it's built during Stage0. 

Similarly, the Stage2 GHC should go somewhere into `_build/stage1/ghc` directory, because it's built during Stage1. 

Or is it confusing for some reason?

angerman: @snowleopard true. `stage0` contains the objects and with the current (not this PR) the products. This however poses a problem when trying to separate the stages cleanly.

Stage0 is our bootstrap stage. Our bootstrap compiler is our stage0 compiler. As such that one should be `_build/stage1/bin/ghc` IMO. Similarly the stage1 compiler built by the stage0 compiler should subsequently be `_build/stage1/bin/ghc`.

The other reason is, if I'm talking about packaging up stageN, I would basically zip `_ build/stageN/{bin,lib}`. And if the stage1 compiler would live in stage0 this wouldn't work anymore. Unless I put bin into stage0, but then I zip `_build/stage0/{bin,lib}` to build the binary distribution of the stage1 compiler, which is kinda confusing I think.

If we separate the notion of build artifacts and products (binaries and installed libraries), then the build artifacts (objects) are in the stage they are build in, but the "installation" of the binaries and libraries happens to stage+1.

angerman:
@snowleopard I try to have the commits be somewhat self-contained. The `Evac` parts should be in 
37b6b91 and 944c894. I guess I might eventually try to squash some commits. Though the periodic merging of upstream into this branch will make this a bit more difficult. We'll have to see.

My current plan is to make this work (proof of concept), and then see if we can simplify / cleanup the history.

src/Builder.hs
snowleopard: > This however poses a problem when trying to separate the stages cleanly.

@angerman What are these problems, specifically? I still don't understand.

> And if the stage1 compiler would live in stage0 this wouldn't work anymore.

Why? Just zip `_build/stage0/{bin,lib}` -- what's wrong with that?

> which is kinda confusing I think

Is everything boils down to the fact that `_build/stage0/bin/ghc-stage1.exe` is confusing? Is that the only 'problem' you are referring to above, or there anything else?

Let's look at this from the documentation perspective.

At the moment we can say: "Everything build in Stage0 is in `_build/stage0`", "Everything build in Stage1 is in `_build/stage1`".

With your approach we have to say: "Everything build in Stage0 is in `_build/stage0` plus `_build/stage1/{bin,lib}`", "Everything build in Stage1 is in `_build/stage1`, minus `_build/stage1/{bin,lib}` plus `_build/stage2/{bin,lib}`".

Now which one is more confusing?

And this doesn't stop in the documentation only. This complexity propagates to the build system itself. For example, to freeze Stage1 we no longer can simply ask Shake to skip rebuilding everything in `_build/stage1`. We will have to exclude `_build/stage1/{bin,lib}` and include `_build/stage2/{bin,lib}`. 

Is all this complexity needed just so that you have `_build/stage1/bin/ghc-stage1.exe` instead of `_build/stage0/bin/ghc-stage1.exe`? I find it a too high price to pay just for the sake of matching 0 and 1 in this path.

snowleopard:
> My current plan is to make this work (proof of concept), and then see if we can simplify / cleanup the history.

Sure, sounds fine with me 👍

angerman:
This now rolls ghc-cabal into hadrian, which requires a few modifications to cabal, unless we want to keep all the hacks in `ghc-cabal`. (I don't.) The few modifications that were needed in cabal to make it sufficiently threadable turned out to be fewer than assumed, but the PRs need to be cleaned up.

This can now successfully build a stage2 compiler for me (in the `_build/stage1/{bin, lib}` folders). 

I think most of the `inplace` logic is gone. Some files are still generated in the wrong location and copied over. As such quite a bit of additional cleanup is needed.

I'll continue by making sure I can build my cross compilers with this (which was my main motivation anyway), and polishing the required patches to Cabal and GHC.

src/GHC.hs
```diff
@@ -204,7 +102,7 @@ stage2Packages = return [haddock]
 -- 'Library', the function simply returns its name.
 programName :: Context -> String
 programName Context {..}
-    | package == ghc      = "ghc-stage" ++ show (fromEnum stage + 1)
+    | package == ghc      = "ghc"
```
snowleopard: After this simplification we can probably just grab the name of the executable from the Cabal file.

snowleopard:
@angerman Many thanks again for your efforts!

Looking forward to seeing the final results -- it seems I might not recognise Hadrian after merging this :)

angerman:
@snowleopard, @izgzhen you might want to take a look at 7b37f15, because I think that's still handling in-tree-gmp incorrectly in the master branch.
```
# Uncomment the following to force `integer-gmp` to use the in-tree GMP 6.1.2
# (other sometimes useful configure-options: `--with-gmp-{includes,libraries}`)
#libraries/integer-gmp_CONFIGURE_OPTS += --configure-option=--with-intree-gmp
```
is from the `mk/build.mk.sample`.

```
ifeq "$(Windows_Host)" "YES"
# Apparently building on Windows fails when there is a system gmp
# available, so we never try to use the system gmp on Windows
libraries/integer-gmp_CONFIGURE_OPTS += --configure-option=--with-intree-gmp
endif
```
is from the `libraries/integer-gmp/gmp/ghc.mk`, note the windows guard.

```
AC_ARG_WITH([intree-gmp],
  [AC_HELP_STRING([--with-intree-gmp],
    [force using the in-tree GMP])],
    [GMP_FORCE_INTREE=YES],
    [GMP_FORCE_INTREE=NO])
```
is from `libraries/integer-gmp/configure.ac`. E.g. we do not want to *force* in-tree-gmp. As it completely prohibits the auto-detect logic for `gmp`.

angerman:
@snowleopard 
> Looking forward to seeing the final results -- it seems I might not recognise Hadrian after merging this :)

Yes, I'm afraid this PR, does things quite a bit different to how hadrian currently does them. It is much more a rethinking of how the build system works and stronger separation of stages.

snowleopard:
> you might want to take a look at 7b37f15, because I think that's still handling in-tree-gmp incorrectly in the master branch

@angerman Sorry, I don't understand what the right logic should be. The commit you linked to never adds the `--with-intree-gmp` flag, but presumably this is not right. Could you explain what needs to be done?

snowleopard:
> Yes, I'm afraid this PR, does things quite a bit different to how hadrian currently does them. It is much more a rethinking of how the build system works and stronger separation of stages.

@angerman I noticed a few places where you are 'rearranging chairs' (e.g. moving files around, which makes it difficult to review/merge the PR in one go), but nothing fundamentally different from what Hadrian does now. Could you point where things are done differently? Of course, getting rid of `ghc-cabal` is a huge and welcome change, but I wouldn't call it a rethinking of how Hadrian works -- you comment made me think I'm missing something important, please let me know what it is :)

angerman:
@snowleopard 
> @angerman Sorry, I don't understand what the right logic should be. The commit you linked to never adds the --with-intree-gmp flag, but presumably this is not right. Could you explain what needs to be done?

Yes, that commit *disables* the logic. The commit I linked to is from the relocatable branch.  Sorry, I could have been more clear about that.

---

I moved the Types around, because they often resulted in import cycles :( And I did not want to add even more boot files.

What this PR does:
- drop ghc-cabal (replace package-data with ConfiguredCabal)
- Tried to get rid of any `inplace` logic.
- Builds binaries in `_build/stageN/bin` and libraries in `_build/stageN`, but **also** installed the libraries into `_build/stageN/lib/package.conf.d`. As such you can just zip `_build/stage1/{bin, lib}` and have a binary distribution.

The "rethinking of how Hadrian works" was meant to mean, that this is a departure from how the current build system works, which Hadrian so far tries to model to some extend.

snowleopard:
> The logic here should be that "in-tree" is a setting and not inferred from the presence of the lib and include dirs.

@angerman I see, so the `inTreeGmp :: Bool` should be a part of the build `Flavour`, just like, for example, `integerLibrary` or `ghcProfiled`?

> I moved the Types around, because they often resulted in import cycles :( And I did not want to add even more boot files.

Aha! That makes sense. I'm personally not a big fan of splitting type definitions into separate modules, but it does often help with import cycles. Let's see how it looks when you finish.

> The "rethinking of how Hadrian works" was meant to mean, that this is a departure from how the current build system works, which Hadrian so far tries to model to some extend.

I see, thanks for clarifying! Nothing to too radical then, at least in my view. We do need to discuss this with GHC devs though.

angerman:
@snowleopard 
>  I see, so the inTreeGmp :: Bool should be a part of the build Flavour, just like, for example, integerLibrary or ghcProfiled?

Yes! It should be user configurable if one want's to force in-tree-gmp or not.

> Aha! That makes sense. I'm personally not a big fan of splitting type definitions into separate modules, but it does often help with import cycles. Let's see how it looks when you finish.

This is something I'm not terribly fond of either, but until GHC can deal with mutually recursive modules, it seems best to have (data) types in their own modules as they are often leaf nodes in the module dependency graph.  Sometimes class instances complicate this a bit. However, the constant running into cycles issue (for me) indicates that putting types into their own modules is a sensible approach for now.

> I see, thanks for clarifying! Nothing to too radical then, at least in my view. We do need to discuss this with GHC devs though.

Definitely!

angerman:
@snowleopard 
> Nice! I haven't thought about the `Rules FilePath` version of `buildRoot`. I like it!
> I didn't know there is `getShakeOptionsRules`. This might simplify some other things too!

I believe an even better solution would be to have
```
class HasShakeOptions a where
  ...
```
and then have two instances for `Action` and `Rules`. But maybe I'm overlooking something
here.

Not sure if that should be in shake proper though. @ndmitchell thoughts?

snowleopard:
@angerman I agree, a type class may be a good solution. I was thinking about a more specific type class that can be part of `Hadrian.Utilities`:

```haskell
class BuildRoot m where
    buildRoot :: m FilePath

instance BuildRoot Action where
    buildRoot = Hadrian.Utilities.buildRoot

instance BuildRoot Rules where
    buildRoot = Hadrian.Utilities.buildRootRules
```

src/Rules/Register.hs
```diff
@@ -1,44 +1,101 @@
...
+import Distribution.Version (Version)
+
+import Hadrian.Haskell.Cabal.Parse as Cabal
+
+parseCabalName :: String -> Maybe (String, Version)
+parseCabalName = readPToMaybe parse
+  where parse = (,) <$> (parsePackageName <* Parse.char '-') <*> parseOptVersion
+
+-- | This rule provides rules for copying packges into the
+-- boot packages db from the installed compiler.
```
bgamari: You'll need to either merge this with the haddock comment below or drop it.

src/Rules/Register.hs
```diff
@@ -1,44 +1,101 @@
...
-        -- TODO: Get rid of this hack.
-        "//" ++ stage0PackageDbDir -/- pkgName package ++ "*.conf" %%>
-            buildConf rs context
+registerPackages :: [(Resource, Int)] -> Context -> Rules ()
+registerPackages rs context@Context {..} = do
+    root <- buildRootRules
+    root -/- inplacePackageDbPath stage %>
+      buildStamp rs context
+
+    root -/- inplacePackageDbPath stage -/- packageDbStamp %> \stamp -> do
```
bgamari: What is going on here? Surely we need to do *something* besides just writing the stamp file, no?

src/Builder.hs
```diff
@@ -46,15 +48,23 @@ instance NFData   CcMode
...
+data GhcCabalMode = Conf | Copy | Reg | HsColour | Check | Sdist
+    deriving (Eq, Generic, Show)
+
+instance Binary   GhcCabalMode
+instance Hashable GhcCabalMode
+instance NFData   GhcCabalMode
+
 -- | GhcPkg can initialise a package database and register packages in it.
-data GhcPkgMode = Init | Update deriving (Eq, Generic, Show)
+data GhcPkgMode = Init | Update | Clone | Dependencies deriving (Eq, Generic, Show)
```
bgamari: Can you document what `Clone` does?

src/Rules/Register.hs
```diff
@@ -1,44 +1,101 @@
...
+    -- in the first place.
+    setupConfig <- (contextPath context) <&> (-/- "setup-config")
+    need [setupConfig]
+
+    need =<< mapM (\pkgId -> packageDbPath stage <&> (-/- pkgId <.> "conf")) depPkgIds
+
+    ways <- interpretInContext context (getLibraryWays <> if package == rts then getRtsWays else mempty)
+    need =<< concatMapM (libraryTargets True) [ context { way = w } | w <- ways ]
+
+    -- might need some package-db resource to limit read/write,
```
bgamari: "might" meaning that this needs to be implemented? Hadrian used to do this, no?

src/Settings/Builders/Common.hs
```diff
@@ -46,13 +49,19 @@ cWarnings =
...
+    top    <- expr topDirectory
+    root   <- getBuildRoot
+    prefix <- ifM (builder Ghc) (return "-package-db ") (return "--package-db=")
+    arg $ prefix ++ top -/- root -/- inplacePackageDbPath stage
+
 bootPackageDatabaseArgs :: Args
 bootPackageDatabaseArgs = do
     stage  <- getStage
     dbPath <- expr $ packageDbPath stage
     expr $ need [dbPath -/- packageDbStamp]
```
bgamari: Is this still needed? Shouldn't `stage0 ? packageDatabaseArgs` be sufficient?

src/Settings/Builders/HsCpp.hs
```diff
@@ -1,6 +1,8 @@
 module Settings.Builders.HsCpp (hsCppBuilderArgs) where
 
 import Settings.Builders.Common
+import Builder ()
```
bgamari: Why is this needed?

src/Settings/Builders/Hsc2Hs.hs
```diff
@@ -1,6 +1,10 @@
 module Settings.Builders.Hsc2Hs (hsc2hsBuilderArgs) where
 
 import Settings.Builders.Common
+import Types.ConfiguredCabal as ConfCabal
+import GHC (autogenPath)
+import Hadrian.Builder (getBuilderPath)
+import Builder ()
```
bgamari: Likewise, why is this needed?

src/Settings/Builders/Xelatex.hs
```diff
@@ -1,6 +1,8 @@
 module Settings.Builders.Xelatex (xelatexBuilderArgs) where
 
 import Settings.Builders.Common
+import Hadrian.Expression (getInput)
+import Builder ()
```
bgamari: Why is this and the one above necessary?

src/Settings/Default.hs-boot
```diff
@@ -3,8 +3,8 @@ module Settings.Default (
     defaultArgs, defaultLibraryWays, defaultRtsWays, defaultFlavour, defaultSplitObjects
     ) where
 
-import Flavour
-import Expression
+import Types.Flavour
+import Types.Expression
```
bgamari: Structuring the module namespace this way seems a bit odd; wouldn't it be more natural for these to be, e.g., `Flavour.Types`? Afterall, `Types.Flavour` has more in common with `Flavour` that, for instance, `Types.Expression`.

src/Rules/Library.hs
```diff
@@ -3,19 +3,44 @@ module Rules.Library (
...
 import Settings
 import Target
 import Utilities
+import GHC.Packages
+
+archive :: Way -> String -> String
+archive way pkgId = "libHS" ++ pkgId ++ (waySuffix way <.> "a")
+
+-- | Building a library consist of building
+-- the artefacts, and copying it somewhere
```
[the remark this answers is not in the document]
snowleopard: I've always been spelling this as "artefacts". Looks like these are just UK vs US spellings:

https://en.oxforddictionaries.com/definition/artefact

I don't mind either way but we might already have a few "artefacts" in Hadrian written by me :)

src/Settings/Default.hs-boot
snowleopard: I second that. We then can also get rid of plural, getting `Expression.Type`, `Flavour.Type`, etc.

snowleopard: I also would like this to be factored out into a separate PR.

src/Rules/Library.hs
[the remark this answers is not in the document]
bgamari: Wow, today I learned. I had no idea. I don't really care; I just thought it was humorous how both spellings were used within five lines of one another.

src/Builder.hs-boot
```diff
@@ -0,0 +1,44 @@
...
+             | Ranlib
+             | Sphinx SphinxMode
+             | Tar TarMode
+             | Unlit
+             | Xelatex
+             | CabalFlags Stage
+
+instance Eq Builder
+
+builderPath' :: Builder -> Action FilePath
```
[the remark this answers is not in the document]
angerman: This is likely due to history. Initially I tried not to move files around as much as possible, and rather tried to have minimal boot files.

src/Builder.hs
```diff
@@ -138,10 +153,35 @@ instance H.Builder Builder where
...
             Make dir      -> need [dir -/- "Makefile"]
             _             -> when (isJust $ builderProvenance builder) $ need [path]
 
+    -- TODO: We would need to encode that asking a builder,
+    --       depending on the "ask" mode, has different return types.
+    --       For now it's the stdout string.
+    --
+    --       This however means that the string -> datatype logic
+    --       needs to reside at the callsite.
+    askBuilderWith :: Builder -> BuildInfo -> Action String
```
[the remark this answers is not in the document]
angerman: I'll have to check if this is still used of if we can just get rid of it altogether. The idea was that you can not only use a builder to do something, but also ask a builder. E.g. ask `ghc` about `--info` or similar.

src/GHC.hs
```diff
@@ -135,19 +36,20 @@ defaultPackages Stage3 = return []
...
     win <- windowsHost
-    ios <- iosHost
     cross <- crossCompiling
     return $ [ binary
              , cabal
              , compareSizes
              , compiler
+             , dataBitcode
+             , dataBitcodeLlvm
+             , dataBitcodeEdsl
```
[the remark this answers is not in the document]
angerman: Yep, this needs to change! It's a bit annoying that layering this ontop of hadrian isn't as trivial as I had hoped. I'll probably need to look into the UserSettings and see if I can extract it into that cleanly, in a way that it doesn't pollute this branch, but keeps letting me use them

src/GHC.hs
```diff
@@ -220,34 +128,20 @@ installStage pkg
...
-    return $ contextPath -/- programName context <.> exe
-  where
-    inplacePath | package `elem` [touchy, unlit, iservBin] = inplaceLibBinPath
-                | otherwise                                = inplaceBinPath
+    path    <- stageBinPath stage
+    pgm     <- programName context
+    return $ path -/- pgm <.> exe
 
 -- | Some contexts are special: their packages do not have @.cabal@ metadata or
 -- we cannot run @ghc-cabal@ on them, e.g. because the latter hasn't been built
```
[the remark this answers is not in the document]
angerman: Yes. Will clean up.

src/Hadrian/Builder.hs
```diff
@@ -42,6 +42,9 @@ class ShakeValue b => Builder b where
     -- | The path to a builder.
     builderPath :: b -> Action FilePath
 
+    -- | Ask the builder for something
+    askBuilderWith :: b -> BuildInfo -> Action String
```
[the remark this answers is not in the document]
angerman: As noted above, this might (hopefully?) not be needed anymore. If it still is, I'll try to make this clearer.

src/Hadrian/Expression.hs
```diff
@@ -71,7 +71,12 @@ class ToPredicate p c b where
 infixr 3 ?
 
 -- | Apply a predicate to an expression.
+
+#if !(MIN_VERSION_base(4,11,0))
```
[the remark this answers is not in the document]
angerman: This as well as the `shake` and `alex` submodules happend when I tried to make HEAD with HEAD work. Turns out you are better off using `./configure --with-ghc=...`

src/Rules.hs
```diff
@@ -20,28 +20,50 @@ import qualified Rules.Program
...
-                 targets <- concatForM allStages $ \stage ->
-                     concatForM (knownPackages \\ [libffi]) $
-                        packageTargets False stage
-                 return $ targets ++ inplaceLibCopyTargets
+topLevelTargets = do
+    phony "binary-dist" $ do
+      -- This is kind of incorrect.  We should not "need" a phony rule.
+      -- Instead we should *need* the libraries and bianries we want to
+      -- put into the bianry distribution.  For now we will just *need*
+      -- stage2 and package up bin and lib.
```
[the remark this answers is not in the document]
angerman: So, this does basically "need" the phony `stage` target. And then just packages up `bin` and `lib`, I believe we should be more explicit about *what* we want in the bin-dist and list it here explicity. e.g. `words "/bin/ghc /bin/ghc-pkg /bin/hsc2hs /lib/packages.conf.d/rts-1.0.conf ...`.
As we might want to add the host `configure` script and other `shipping` items here as well, which are not part of `stage2`.

src/Rules.hs
```diff
@@ -90,16 +112,17 @@ packageRules = do
     let dynamicContexts = liftM3 Context [Stage1 ..] knownPackages [dynamic]
     forM_ dynamicContexts Rules.Library.buildDynamicLib
 
-    forM_ (filter isProgram knownPackages) $
-        Rules.Program.buildProgram readPackageDb
+    Rules.Program.buildProgram readPackageDb
+
+    forM_ [Stage0 .. ] $ \stage -> do
+      Rules.Register.registerPackages writePackageDb (Context stage base vanilla) -- base is only a dummy here.
```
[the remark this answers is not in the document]
angerman: `registerPackages` expects a `Context`, but ignores the package. There is some Context constructor that sets the `package` to `error`, which I had hoped would be fine, but apparently `package` is still forced (I didn't bother looking into this more), even though it's not used and as such errors out. Hence we just set a dummy package (here `base`).

src/Settings/Builders/HsCpp.hs
angerman: Because GHC wanted it there I believe, I'll try to drop it again, and see if GHC complains again.

src/Rules/Register.hs
angerman: will do.

src/GHC.hs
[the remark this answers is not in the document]
snowleopard: Have you tried adding these packages to `UserSettings.userPackages`? 

Here is an example: 

https://github.com/snowleopard/hadrian/blob/master/doc/user-settings.md#packages

src/Rules/Register.hs
angerman: This is a shake short coming iirc. You can not *depend* on folders. You can only depend on files. As such, depending on the `stamp` file, will implicitly build the folder, and call the rule above, which does the initialization.

src/Rules/Program.hs
```diff
@@ -1,116 +1,100 @@
...
+                                 copyFile (srcDir -/- takeFileName bin) bin
+                _ -> buildBinary rs bin =<< programContext stage package
+          -- Rules for the GHC package, which is built 'inplace'
 
 -- TODO: Get rid of the Paths_hsc2hs.o hack.
 buildBinary :: [(Resource, Int)] -> FilePath -> Context -> Action ()
 buildBinary rs bin context@Context {..} = do
     binDeps <- if stage == Stage0 && package == ghcCabal
         then hsSources context
+        -- then do -- this is a hack, but he ghc-cabal packge only list's it's Main
```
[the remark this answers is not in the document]
angerman: YES!

src/Rules/Program.hs
```diff
@@ -1,116 +1,100 @@
...
-buildWrapper :: Context -> Wrapper -> FilePath -> WrappedBinary -> Action ()
-buildWrapper context@Context {..} wrapper wrapperPath wrapped = do
-    contents <- interpretInContext context $ wrapper wrapped
-    writeFileChanged wrapperPath contents
-    makeExecutable wrapperPath
-    putSuccess $ "| Successfully created wrapper for " ++
-        quote (pkgName package) ++ " (" ++ show stage ++ ")."
+              cross <- crossCompiling
+              -- for cross compiler. copy the stage0/bin/<pgm>
+              -- into stage1/bin/
```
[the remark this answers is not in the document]
angerman: So, for cross compilation, we don't really want the stage2 compiler (when host /= target). We want the compiler that runs on the the host (e.g. the one we compile the compiler on) but targets what ever we want to target.

To make the bin-dist logic simpler. And end up with the right `ghc` executable in the final `bin` directory, for cross compilers we take the stage0 binary (used to be called `ghc-stage1`) and copy it into stage1 (used to be called `ghc-stage2`). Similarly for other binaries, but only when building a cross compiler.

This might be at odds when we want to build a cross compiled ghc, as opposed to ghc as a cross compiler.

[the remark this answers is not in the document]
angerman: In that case we would *not* want this logic, and use the stage1 compiler (build == host -> target) to produce the stage2 compiler (host == target /= build)

src/Rules/Program.hs
```diff
@@ -1,116 +1,100 @@
...
+          nameToCtxList <- forM activeProgramPackages $ \pkg -> do
+            let ctx = vanillaContext stage pkg
+            name <- programName ctx
+            return (name <.> exe, ctx)
 
-        -- Rules for the GHC package, which is built 'inplace'
-        when (package == ghc) $ do
-            inplaceBinPath -/- programName context <.> exe %> \bin ->
-                buildBinaryAndWrapper rs bin =<< programContext stage package
+          case lookup (takeFileName bin) nameToCtxList of
```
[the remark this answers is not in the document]
angerman: I'm not exactly sure *how* inefficient this is, but this was also the approach suggested by @ndmitchell in snowleopard/hadrian/issues/444#issuecomment-340183730

src/Rules/Register.hs
angerman: This used to be in a different action. Probably stale. However this action, while it gets the resource info, discards that argument. Not sure if we need to use it here.

src/Settings/Builders/Common.hs
angerman: YES!

angerman: Yes, `stage0 ? packageDatabaseArgs` should be sufficient.

src/Settings/Default.hs-boot
angerman: sure, why not. I don't really mind where the modules live. It's just that data-types and alike are often leaf nodes, and if coupled too tightly with the logic around them quickly lead to cyclic imports.

src/Builder.hs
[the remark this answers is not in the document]
bgamari: Yes, I agree that having a more descriptive type would help a lot

src/Hadrian/Builder.hs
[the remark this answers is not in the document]
bgamari: Ahh, lovely.

src/Rules/Register.hs
bgamari: Where is this rule `need`ed?

bgamari:
Great work, @angerman!

src/Rules/Library.hs
```diff
@@ -3,19 +3,44 @@ module Rules.Library (
...
+-- with cabal, and finally registering it
+-- with the compiler via cabal in the
+-- package database.
+--
+-- So we'll assume rules to build all the
+-- package artifacts, and provide rules for
+-- the any of the lirbary artifacts.
+library :: Context -> Rules ()
+library context@Context{..} = do
```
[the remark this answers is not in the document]
ggreif: "lirbary"

src/Rules/Program.hs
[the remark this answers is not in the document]
ggreif: "packge"

src/Rules/Program.hs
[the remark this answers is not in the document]
ggreif: "compiler. copy" --> "compiler, copy"

src/Rules.hs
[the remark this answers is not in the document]
ggreif: "bianries", "bianry"

src/Rules/Program.hs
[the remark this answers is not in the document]
angerman: @ggreif, you are one hell of a proofreader! 💯

src/Rules/PackageData.hs
```diff
@@ -75,45 +63,27 @@ packageCSources pkg
...
-postProcessPackageData context@Context {..} file = do
-    top     <- topDirectory
-    cmmSrcs <- getDirectoryFiles (pkgPath package) ["cbits/*.cmm"]
-    path    <- buildPath context
-    let len = length (pkgPath package) + length (top -/- path) + 2
-    fixFile file $ unlines
-                 . (++ ["CMM_SRCS = " ++ unwords (map unifyPath cmmSrcs) ])
-                 . map (drop len) . filter ('$' `notElem`) . lines
+    | pkg == base = do
+        sources <- getDirectoryFiles (pkgPath pkg) ["cbits/*.cmm"]
```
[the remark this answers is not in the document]
angerman: We already do. All dead code. Sorry you had to review this.

angerman:
Ok. Did a first pass over the review. Still needs a few more passes.

angerman:
Also, this diff is still a net addition. Need. To. Delete. More. Code.

snowleopard:
> Nuke Oracles.PackageData

❤️

ndmitchell:
@angerman There are indeed a bunch of things that can work in both Action and Rules, including getting options, things that require options (e.g. `putLoud`), and you can see `need`/`want` as members of a type class. Certainly introduce it in Hadrian, and if it really does seem super-valuable, I'll migrate it to Shake. I don't yet feel it's crossed the level where introducing the abstraction in Shake itself pays off for the extra complexity, but it's getting close...

angerman:
Thanks, @ndmitchell I'll have a look. I'm not yet sure though of the performance implications. And how a wild card rule stacks up against a set of generated non-wildcard rules (potentially unused), now how expensive it actually is to calculate the `Context` from the rule's path name.

ndmitchell:
Non-wildcard rules are relatively expensive - it's a linear search. Unless you can do `|%>` on only-concrete files, in which case it turns it into a `Set` lookup (nice and fast).

src/Settings/Default.hs-boot
angerman: @alpmestan is working on a PR for this right now.

alpmestan:
(Just adding a link as some sort of "TODO", for me to revisit my documentation rules patches to add `--quickjump`: https://github.com/ghc/ghc/commit/9ff4cce373765cec064e2a41d4c6ddd84b873f04)

snowleopard:
@alpmestan Oops, sorry, I rushed ahead with #505 to make sure I won't forget this.

alpmestan:
No worries :)

izgzhen:
I’m curious what is left to be done in this PR after #531?

alpmestan:
> I’m curious what is left to be done in this PR after #531?

I think we're pretty much done, aren't we @angerman ? We got rid of the need for ghc-cabal, made the build root configurable, incorporated a bindist rule (it's even cleaner than the code that I wrote in this branch).

snowleopard:
@alpmestan Sounds great! Once you are confident nothing has been lost, please close this PR.

angerman:
Guess we can close this for now. If I'll end up with issues building my cross compilers again, I'll probably start a new branch.

snowleopard:
@angerman Once again, many thanks for this amazing PR :-)

Issue #449, opened 2017-11-07
macOS build failed in parsing gcc-lt-34 configuration flag


izgzhen:
```
Error when running Shake build system:
* _build/stage1/rts/libHSrts_thr.a
* _build/stage1/rts/c/Adjustor.thr_o
Configuration flag 'gcc-lt-34 = @GccLT34@' cannot be parsed.
CallStack (from HasCallStack):
  error, called at src/Oracles/Flag.hs:44:48 in main:Oracles.Flag
```

bgamari:
This is almost certainly fallout from https://phabricator.haskell.org/D4165

izgzhen:
I see. I'll give a fix today

snowleopard:
Poor `GccLt34`. Its life was so short (was added to Hadrian two days ago).

snowleopard:
This was fixed in #450.

PR #450, opened 2017-11-07
Drop GccLtXX flags, require GCC > 4.7 and up

izgzhen:
See https://github.com/snowleopard/hadrian/issues/449.

### DIFF
--- a/cfg/system.config.in
+++ b/cfg/system.config.in
@@ -33,9 +33,6 @@ ar-supports-at-file = @ArSupportsAtFile@
 cc-clang-backend    = @CC_CLANG_BACKEND@
 cc-llvm-backend     = @CC_LLVM_BACKEND@
 gcc-is-clang        = @GccIsClang@
-gcc-lt-34           = @GccLT34@
-gcc-lt-44           = @GccLT44@
-gcc-lt-46           = @GccLT46@
 hs-cpp-args         = @HaskellCPPArgs@
 
 # Build options:
--- a/src/Oracles/Flag.hs
+++ b/src/Oracles/Flag.hs
@@ -11,9 +11,6 @@ import Oracles.Setting
 data Flag = ArSupportsAtFile
           | CrossCompiling
           | GccIsClang
-          | GccLt34
-          | GccLt44
-          | GccLt46
           | GhcUnregisterised
           | LeadingUnderscore
           | SolarisBrokenShld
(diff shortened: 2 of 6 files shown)


snowleopard:
@izgzhen Thanks for the patch!

Looks like you haven't removed all uses of the flags, so CI Hadrian can't be compiled.

izgzhen:
> Looks like you haven't removed all uses of the flags, so Hadrian can't be compiled.

Yes! I was in a hurry at the time of making this PR, sorry.

snowleopard:
We now hit GHC panic:
```
Stderr:
ghc: panic! (the 'impossible' happened)
  (GHC version 8.0.2 for x86_64-unknown-mingw32):
	tcIfaceGlobal (local): not found
  You are in a maze of twisty little passages, all alike.
  While forcing the thunk for TyThing SyntaxExpr
  which was lazily initialized by initIfaceTcRn,
  I tried to tie the knot, but I couldn't find SyntaxExpr
  in the current type environment.
  If you are developing GHC, please read Note [Tying the knot]
  and Note [Type-checking inside the knot].
  Consider rebuilding GHC with profiling for a better stack trace.
  Contents of current type environment: []
Please report this as a GHC bug:  http://www.haskell.org/ghc/reportabug
)
```
:-S Presumably GHC is broken? Or did we break it by a strange combination of flags?

snowleopard:
Ah, wait, this is bootstrapping GHC! So, it's a real bug which we somehow triggered by flags.

izgzhen:
> Ah, wait, this is bootstrapping GHC! So, it's a real bug which we somehow triggered by flags.

O_o interesting

snowleopard:
@izgzhen One possible explanation is that this GHC commit: 

https://github.com/ghc/ghc/commit/0ff152c9e633accca48815e26e59d1af1fe44ceb

Changed `compiler/hsSyn/HsExpr.hs` in such a way that it started to trigger an old GHC bug. It's therefore not the fault of this PR.

snowleopard:
OK, I've confirmed this. GHC appears to be currently broken on 8.0.2.

@bgamari Shall I report this or is this a known issue?

bgamari:

> OK, I've confirmed this. GHC appears to be currently broken on 8.0.2.
>
> @bgamari Shall I report this or is this a known issue?
>
No need, I'm working with @alanz on it.

PR #452, opened 2017-11-10
Widen bounds on Cabal

bgamari:
GHC's `master` branch now sits on a snapshot of Cabal 2.1.0.0

### DIFF
--- a/hadrian.cabal
+++ b/hadrian.cabal
@@ -118,7 +118,7 @@ executable hadrian
                        , TypeFamilies
     build-depends:       base >= 4.8 && < 5
                        , ansi-terminal        == 0.6.*
-                       , Cabal                == 2.0.0.2
+                       , Cabal                >= 2.0.0.2 && < 2.2
                        , containers           == 0.5.*
                        , directory            >= 1.2 && < 1.4
                        , extra                >= 1.4.7


snowleopard:
@bgamari Thanks, merged!

I've just realised that we've broken the CI by the merge, so it failed for this PR, see #453.

snowleopard:
@bgamari There seem to be problems with compiling Alex and Happy against this in-tree Cabal:
```
In the dependencies for alex-3.2.1:
    Cabal-2.1.0.0 must match <2.1 (latest applicable is 2.0.0.2)
needed for unknown reason - stack invariant violated.
In the dependencies for happy-1.19.5:
    Cabal-2.1.0.0 must match <2.1 (latest applicable is 2.0.0.2)
needed for unknown reason - stack invariant violated.
```

See https://ci.appveyor.com/project/snowleopard/hadrian/build/1.0.765

It's probably possible to build Alex/Happy with an older Cabal, but perhaps it's actually worth moving upper bounds in Alex and Happy instead?

bgamari:
Sorry I missed this; it seems you sorted it out? Is this a Cabal bug or have I misunderstood?

snowleopard:
@bgamari I did fix it, but the fix feels like a hack.

The problem was the `stack install alex happy` command, which was run inside the Hadrian directory. The command must have picked up the include path to GHC's version of Cabal, which does not satisfy Alex's upper bound, hence failing to build Alex.

My 'solution' was to run `stack install alex happy` from outside the Hadrian directory, which then correctly built Alex with the bootstrapping Cabal library (which does satisfy Alex's upper bound).

Do you have any better suggestions?

bgamari:

> @bgamari I did fix it, but the fix feels like a hack.
>
> The problem was the `stack install alex happy` command, which was run
> inside the Hadrian directory. The command must have picked up the
> include path to GHC's version of Cabal, which does not satisfy Alex's
> upper bound, hence failing to build Alex.
>
I'll admit I'm not entirely sure I understand how this could happen.
I'll need to try to reproducing the issue locally.

Does cabal-install exhibit the same behavior?

snowleopard:
> Does cabal-install exhibit the same behavior?

Not sure how to tell. For some reason I can't build Hadrian on Windows using the `build.cabal` approach.

Issue #453, opened 2017-11-10
Fix CI


snowleopard:
Merging Hadrian into the GHC tree broke the CI, for example see:

https://circleci.com/gh/snowleopard/hadrian/181

I think it should be easy to fix, but I'll record it here anyway.

snowleopard:
The scripts are fixed, but it appears that Hadrian is still broken by changes in GHC. Will investigate tomorrow.

snowleopard:
Should be completely fixed now.

PR #454, opened 2017-11-10
Fix CI scripts

snowleopard:
See #453


snowleopard:
Looks like CircleCI and Travis bots are fixed (they fail for different reasons), but AppVeyor still requires further work.

PR #459, opened 2017-11-14
hadrian: Disable -Wno-undef in files which include ffi.h

bgamari:
As pointed out in D4181, these headers refer to undefined macros.


snowleopard:
Thanks @bgamari! Merged.

PR #466, opened 2017-11-16
Unbreak `cabal new-build`

hvr:
Using a local `filepath` forces `new-build` to build the
non-local build-tool `alex` as in-place which however isn't supported properly
by `cabal new-build` yet (this will be fixed at the earliest in cabal 2.2) and thus
breaks `hadrian/build.sh` for me.


snowleopard:
Thanks @hvr! Merged.

PR #469, opened 2017-11-16
UserSettings: Document userArgs

bgamari:


src/UserSettings.hs
```diff
@@ -3,8 +3,8 @@
 -- If you don't copy the file your changes will be tracked by git and you can
 -- accidentally commit them.
 module UserSettings (
-    userBuildRoot, userFlavours, userPackages, verboseCommand,
-    buildProgressColour, successColour, stage1Only
+    userBuildRoot, userFlavours, userPackages, userArgs,
```
snowleopard: Why is `userArgs` exported?

src/UserSettings.hs
```diff
@@ -36,6 +36,10 @@ userFlavour = defaultFlavour { name = "user" } -- Modify other settings here.
 userPackages :: [Package]
 userPackages = []
 
+-- | Arguments to pass to builders. See @doc/user-settings.md@.
```
snowleopard: Presumably, you meant to use `userArgs` when defining `userFlavour`?

See example here: https://github.com/snowleopard/hadrian/blob/master/doc/user-settings.md#command-line-arguments

snowleopard:
@bgamari Thanks, we do need to make `UserSettings` easier to use. We don't need to export `userArgs`, we need to use it in this line:
```haskell
userFlavour = defaultFlavour { name = "user" } -- Modify other settings here.
```

bgamari:
Indeed I completely misunderstood the documentation here. Ignore me,

snowleopard:
I refuse to ignore you -- I'll try to improve the docs! :)

bgamari:
Thanks @snowleopard!

PR #473, opened 2017-11-17
Add ways to build hadrian using nix

angerman:
This adds two new files to the hadrian directory

    shell.nix sets up the build envrionment you need to build ghc
    build-nix is a simple wrapper which invokes hadrian in the correct environment

Note: this patch was authored by @mpickering, however it ended up on phabricator due to the subtree as https://phabricator.haskell.org/D4207.

### DIFF
--- a/build-nix
+++ b/build-nix
@@ -1,0 +1,32 @@
+#!/usr/bin/env nix-shell
+#! nix-shell -i bash shell.nix
+
+function rl {
+    TARGET_FILE="$1"
+
+    cd "$(dirname "$TARGET_FILE")"
+    TARGET_FILE="$(basename "$TARGET_FILE")"
+
+    # Iterate down a (possible) chain of symlinks
+    while [ -L "$TARGET_FILE" ]
+    do
+        TARGET_FILE="$(readlink "$TARGET_FILE")"
+        cd "$(dirname "$TARGET_FILE")"
+        TARGET_FILE="$(basename "$TARGET_FILE")"
+    done
+
+    # Compute the canonicalized name by finding the physical path
+    # for the directory we're in and appending the target file.
+    PHYS_DIR="$(pwd -P)"
+    RESULT="$PHYS_DIR/$TARGET_FILE"
+    echo "$RESULT"
+}
+
+absoluteRoot="$(dirname "$(rl "$0")")"
+echo $absoluteRoot
+cd "$absoluteRoot"
+
+hadrian         \
+  --lint                         \
+  --directory="$absoluteRoot/.." \
+  "$@"
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,57 @@
+{ nixpkgs ? import <nixpkgs> {} }:
+
+let
+  haskellPackages = nixpkgs.haskell.packages.ghc821;
+
+  removeBuild = path: type:
+    let baseName = baseNameOf (toString path);
+    in
+        ! (baseName == "_build"
+           || baseName == "dist"
+           || baseName == "dist-newstyle"
+           || baseName == ".stack-work"
+           || baseName == "config.log"
+           || baseName == "config.status"
+           || nixpkgs.lib.hasSuffix ".sh" baseName
+           || !(nixpkgs.lib.cleanSourceFilter path type)) ;
+
+  filterSrc = path: builtins.filterSource removeBuild path;
+
+
+  hadrianPackages = nixpkgs.haskell.packages.ghc821.override {
+    overrides = self: super: let
+        localPackage = name: path: self.callCabal2nix name (filterSrc path) {};
+      in {
+        hadrian = localPackage "hadrian" ./. ;
+        shake = self.callHackage "shake" "0.16" {};
+        Cabal = localPackage "Cabal" ./../libraries/Cabal/Cabal ;
+        filepath = localPackage "filepath" ./../libraries/filepath ;
+        text = localPackage "text" ./../libraries/text  ;
+        hpc = localPackage"hpc" ./../libraries/hpc ;
+        parsec = localPackage "parsec" ./../libraries/parsec ;
+        HUnit = nixpkgs.haskell.lib.dontCheck (self.callHackage "HUnit" "1.3.1.2" {});
+        process = localPackage "process" ./../libraries/process ;
+        directory = localPackage "directory" ./../libraries/directory ;
+      }; };
+
+in
+
+  nixpkgs.stdenv.mkDerivation {
+    name = "ghc-dev";
+    buildInputs = [
+                    hadrianPackages.hadrian
+                    nixpkgs.haskell.compiler.ghc821
+                    haskellPackages.alex
+                    haskellPackages.happy
+                    nixpkgs.python3
+                    nixpkgs.git
+                    nixpkgs.autoconf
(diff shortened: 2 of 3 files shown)


build-nix
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env nix-shell
+#! nix-shell -i bash shell.nix
```
snowleopard: This file looks almost identical to `build.stack.nix.sh` -- what's the difference? Do we need both?

snowleopard:
Can we stick to the convention that shell scripts have `.sh` extension?

A general comment: I am concerned that we keep accumulating too many various build scripts, which can be confusing. Perhaps we should only keep the default `build.sh` and `build.bat` at the top level and create a separate directory called `scripts` for custom build scripts?

snowleopard:
@mpickering @angerman Thanks for the PR and for bringing it here!

build-nix
mpickering: I see, it is quite similar. But the obvious difference being that this doesn't require stack nor cabal to be installed on your system and is more direct than using a stack.yaml file to specify the nix configuration.

alpmestan: I just copied this in my tree for now and would therefore be the second user of this approach.

snowleopard: So, to use this script one will first have to somehow build Hadrian. This feels like an easy way to shoot oneself in the foot by using an outdated Hadrian binary.

It may be a good idea to separate the two concerns somehow: building Hadrian and building GHC, but this PR seems to only address the latter, but not the former, unless I'm missing something?

shell.nix
```diff
@@ -0,0 +1,39 @@
...
+        shake = self.callHackage "shake" "0.16" {};
+        Cabal = self.callCabal2nix "Cabal" ./../libraries/Cabal/Cabal {};
+        filepath = self.callCabal2nix "filepath" ./../libraries/filepath {};
+        text = self.callCabal2nix "text" ./../libraries/text {} ;
+        hpc = self.callCabal2nix "hpc" ./../libraries/hpc {};
+        parsec = self.callCabal2nix "parsec" ./../libraries/parsec {};
+        HUnit = nixpkgs.haskell.lib.dontCheck (self.callHackage "HUnit" "1.3.1.2" {});
+      }; };
+
+  hadrian = hadrianPackages.ghcWithPackages (ps: [ ps.hadrian ]);
```
snowleopard: I'm not familiar with nix, so perhaps this is the second part of the puzzle -- is this the nix-based build script for Hadrian?

build-nix
snowleopard: (I've never used nix, so excuse my naive comments!)

shell.nix
alpmestan: It overrides the default haskell packageset by replacing hadrian, shake, etc by the local ones, except for shake for which we specifically require 0.16, and HUnit 1.3.1.2. Finally, line 19 uses this "new" package set to define a build environment that has a ghc + hadrian and all of its dependencies in it. This is then used along with many other things to provision the entire build environment needed to build _ghc_.

snowleopard: So, if the user edits `hadrian/UserSettings.hs` will the invokation of the `build-nix` script automatically rebuild Hadrian?

alpmestan: Yes. In fact, it will even do it when any file from the hadrian dir changes, but this is a quirk @mpickering told me about and we know how to fix it. If @mpickering doesn't get around to doing it in the upcoming days I'll do it.

mpickering:
There are two files.

The first `shell.nix` can be invoked directly by calling `nix-shell shell.nix` and its goal is to set up an environment in which you can build ghc. As such, it adds dependencies to the environment such as gmp, python3 and so on. One such dependency is hadrian itself, so the instructions about how to build hadrian are automatically calculated from hadrian.cabal. This happens in the call `callCabal2nix` which looks in the given directory for a cabal file and creates a nix expression based on that which is can use to build the library. 

The second `build-nix` is a convenient wrapper around this `shell.nix` which automatically invokes hadrian in the correct environment. 

The advantage of this script for nix users is that they don't have to use `cabal` or `stack` in order to manage their dependencies, they can just use the normal way they manage dependencies.

Hope that clears things up.

snowleopard:
@mpickering Thanks, could you add this clarification as a comment on top of these scripts? I'm sure I'm not the only one who might be confused by all the different scripts we have. I'll try to add comments to other scripts too.

shell.nix
snowleopard: @alpmestan Note that in principle Hadrian should also be rebuilt when the in-tree Cabal library is modified. But this is a rather rare scenario.

mpickering: That is also taken care of correctly but not evident from the script. 

Essentially `callCabal2nix` creates a function which takes all the dependencies of `hadrian` and returns a recipe to construct `hadrian` using these dependencies. It then performs some magic to automatically fill in these arguments correctly and because we specify that we want to use the in-tree `Cabal` on a later line, it passes this version of `Cabal` as an argument. The key principle of `nix` is that when an argument changes, then you have to rebuild the result, so when the in-tree `Cabal` changes, then `Cabal` and hence `hadrian` will be rebuilt.

snowleopard: Aga, that's cool!

build-nix
alpmestan: When you call this script, it enters the environment specified in shell.nix, which "provisions" among other things a local build of hadrian.

alpmestan: (and it has everything needed for building ghc, too)

shell.nix
alpmestan: We got you halfway to using nix haven't we? :)

duog:
@snowleopard: I added `build.stack.nix.sh` to allow `build.stack.sh` to not require nix.

shell.nix
snowleopard: @alpmestan Yes, you have :) I've already heard a lot of good things about nix and I'm tempted!

snowleopard:
> I added `build.stack.nix.sh` to allow `build.stack.sh` to not require nix.

@duog Right. Do we need all `build-nix`, `build.stack.sh` and `build.stack.nix.sh`? Sounds a like there may be potential redundancy here...

build-nix
snowleopard: Thanks!

mpickering:
I improved the script this evening so that it doesn't rebuild hadrian so often and works with ghc802 as well. 

I need to add some comments and non-functional changes tomorrow.

alpmestan:
> Perhaps we should only keep the default build.sh and build.bat at the top level and create a separate directory called scripts for custom build scripts?

Probably a good idea yes. With a short `README` that says "use this one if you want to build hadrian using X" for each specific script.

snowleopard:
@alpmestan Indeed, that will work well. I've opened an issue to track this: #475.

duog:
@snowleopard: Yes, it does seem there is redundancy. Unfortunately I don't understand nix, specifically the stack-nix integration, well enough to have an opinion on what should be done. Presumably the nix section in `stack.yaml` was added to help someone's workflow, we'd need to check with them.

mpickering:
I added some comments now and renamed the file. Anything else left to do?

snowleopard:
@mpickering Many thanks, that's great. Merged.

PR #486, opened 2017-12-04
Extend ghcHEAD derivation rather than make a new derivation

mpickering:
This allows you to use configurePhase/buildPhase as usual and
set the right configureFlags rather than end up with a broken
executable.


snowleopard:
@mpickering Thanks! I can't do any meaningful review, but happy to merge if this is ready and helps you :)

Issue #494, opened 2018-01-15
Run ghc's test can't work


sighingnow:
When I run `build.stack.sh test`, I get the following error:

```
runtests.py: error: unrecognized arguments: --configfile=testsuite/config/ghc
```

The ghc's `--configfile` flag has already been renamed to `--config-file` [six months ago](https://github.com/ghc/ghc/commit/5e940bd3d554729ce650008a72b4f82a78578a7b).

sighingnow:
I would like to push a patch to improve the test support of hadrian. Should I submit the patch to [GHC phabricator](https://phabricator.haskell.org/) or open a pull request here ?

snowleopard:
@sighingnow Many thanks for the report and for offering your help! Please open a PR here.

Note that there is also the `validate` build rule, which runs GHC's validation script, which you might want to try as well. However, in any case we should try to fix the `test` rule as well.

bgamari:
@angerman, @alpmestan, I believe this is relevant to your work.

alpmestan:
@bgamari It is, but as you can see in https://github.com/snowleopard/hadrian/pull/495, @sighingnow and I discussed all the problems that we respectively had identified, both there and by email a little bit. @sighingnow seemed well on the way to getting a good chunk of the tests working. We'll mostly be able to reuse his work directly (even though it might require a tweak or two) with the relocatable approach, nothing scary really. :)

izgzhen:
Running `./build.sh test` now led to an error:

```
/home/zhen/repos/ghc/_build/stage1/rts/build/libCffi_p.a: copyFile: does not exist (No such file or directory)
shakeArgsWith    0.000s    0%                           
Function shake   0.219s    0%                           
Database read    0.232s    0%                           
With database    0.098s    0%                           
Running rules  604.736s   99%  =========================
Total          605.286s  100%                           
Error when running Shake build system:
* test
* _build/stage1/bin/ghc
* OracleQ (PackageDataFile (Context {stage = Stage1, package = Package {pkgLanguage = Haskell, pkgType = Program, pkgName = "ghc-bin", pkgPath = "ghc"}, way = v}))
* _build/stage1/ghc/setup-config
* _build/stage1/lib/package.conf.d/containers-0.5.11.0.conf
* OracleQ (PackageDataFile (Context {stage = Stage1, package = Package {pkgLanguage = Haskell, pkgType = Library, pkgName = "containers", pkgPath = "libraries/containers"}, way = v}))
[...]

Issue #187, opened 2016-01-22
Add validation support


snowleopard:
We need to add support for running automated GHC tests when the build system is run with `test` target.

Below are specific subtasks we identified in this thread (see discussions below):
- [x] Implement a temporary target `validate` which simply runs `make fast` in `testsuite/tests`.
- [ ] Support a flag like `--test-hc=...` to allow specifying a Haskell compiler to use for running tests. Possible options could be `stage0` (bootstrapping compiler), `stage1`, `stage2` (default), `stage3` and `path/to/compiler` to specify any other compiler. We need to support paths with spaces and sometimes expand paths using environment variable `PATH`.
- [ ] Detect correct paths to `ghc-pkg`, `hpc`, `runghc`, `haddock`, `hsc2hs`, `hp2ps` in accordance with the chosen `--test-hc`. See `testsuite/mk/boilerplate.mk`.
- [ ] Don't trust settings like `getLibraryWays` in the build system during testing. Instead probe the file system to see which ways are actually there. See `HAVE_PROFILING` in `testsuite/mk/test.mk`. Other options can be found via `ghc --info` and `ghc +RTS --info`. See `testsuite/mk/ghc-config.hs`, which generates a `Makefile` with options passed to the Python driver from `testsuite/mk/test.mk`.
- [ ] Use `getPackages` to generate all `--rootdir` flags for passing to the python `runtests.py` script.
- [ ] Run only one specific test: `build.sh test --only=...`.
- [ ] Run local tests only when in a specific source folder: `build.sh test --local-only`. If possible detect this automatically, so `--local-only` flag is not required.
- [ ] Support `--way=...` flag. Example: `make TEST='tc002 tc002' WAY='normal optasm'`

snowleopard:
Several quick attempts to reuse the `validate` script have failed, so I decided to start implementing a proper test rule in the new build system. (Note, I still rely on the old Python scripts for testing -- rewriting them seems to be a major undertaking.) The test rule does work for me on Windows, but the functionality is very limited at the moment. You can give it a try by running:

`shake-build/build.sh test`

This should run (some) tests. I added a section on testing to the README: https://github.com/snowleopard/shaking-up-ghc#testing.

thomie:
[...]

PR #495, opened 2018-01-17
[WIP] Support run GHC's test from hadrian.

sighingnow:
1. Necessary command line arguments to run test driver.
   + `--test-only=<TESTS>`
   + `--skip-perf`
   + `--summary=<TEST_SUMMARY>`
   + `--junit=<TEST_SUMMARY_JUNIT>`
   + `--config=<EXTRA_TEST_CONFIG>`
2. Synchronize configurations from test.mk.
3. Synchronize GHC's compilation flags from test.mk (that's very important).

**WIP**

- [x] I'm not sure if it's fine to add these command line arguments for running test.
- [ ] Document about how to use hadrian to run one or more test cases.


sighingnow:
Another question is that I have noticed that currently hadrian just call sytem's `make` to run the `validate` target. Should we move to `make` to run `test` target or call the `runtest.py` script from hadrian directly ?

snowleopard:
@sighingnow Many thanks for this. To answer your questions:

> I'm not sure if it's fine to add these command line arguments for running test.

I think that's fine. Otherwise there is really no way to conveniently use the built-in `test` rule, because one often wants to test just a single test case. However, I think we don't need the `test-` prefixes, because that would lead to a bit of duplication in command lines, e.g. `hadrian test --test-only=123`. Let's try to keep the flags exactly as they are currently in the driver script, e.g. `hadrian test --only=123`. This also has the benefit that GHC developers do not need to learn any new flags.

> Another question is that I have noticed that currently hadrian just call sytem's `make` to run the `validate` target. Should we move to `make` to run `test` target or call the `runtest.py` script from hadrian directly ?

At the moment we run `validate` via `make` because that's the simplest thing to do. However, in future I think it would be best to get rid of `make` completely, and possibly even rewrite the `runtest` logic in Hadrian directly. Does this answer your question?

src/Hadrian/Utilities.hs
```diff
@@ -453,3 +460,21 @@ renderUnicorn ls =
     ponyPadding = "                                            "
     boxLines :: [String]
     boxLines = ["", "", ""] ++ (lines . renderBox $ ls)
+
+-- | These arguments are used by the `test` target.
+data TestArgs = TestArgs
```
snowleopard: I'm not sure why `TestArgs` is defined here, let's move it to `CmdLineArgs`.

Everything in `src/Hadrian` is supposed to be generally useful for build systems, i.e. not GHC-specific, whereas flags passed to the GHC's testsuite are very much GHC-specific.

src/Settings.hs
```diff
@@ -66,3 +66,7 @@ findPackageByName name = find (\pkg -> pkgName pkg == name) knownPackages
 -- | Install's DESTDIR setting.
 getDestDir :: Action FilePath
 getDestDir = fromMaybe "" <$> cmdInstallDestDir
+
+-- | Arguments to run GHC's test script.
+getTestArgs :: Action TestArgs
```
snowleopard: Not clear why need another name for the same function. I'd suggest to drop it.

src/Rules/Test.hs
```diff
@@ -28,45 +28,116 @@ testRules = do
...
 
-            -- , "--config", "hp2ps="    ++ quote ("hp2ps")
-            -- , "--config", "hpc="      ++ quote ("hpc")
-            -- , "--config", "gs=$(call quote_path,$(GS))"
-            -- , "--config", "timeout_prog=$(call quote_path,$(TIMEOUT_PROGRAM))"
+        -- See: https://github.com/ghc/ghc/blob/master/testsuite/mk/test.mk#L291
+        let timeout_prog = "testsuite/timeout/install-inplace/bin/timeout"
+
+        quietly . cmd "python3" $
+            [ "testsuite/driver/runtests.py" ] ++
```
snowleopard: We probably should be constructing this command line using the abstractions available in Hadrian. For example:

https://github.com/snowleopard/hadrian/blob/master/src/Settings/Builders/Haddock.hs

sighingnow:
Thanks for your reply and review. I will fix these problem this weekends. I'll work on the `validate` rule after finishing the `test` rule.

alpmestan:
I recently got started tackling the very same thing, but you went through much more trouble to integrate this properly with command-line flags etc. Can this run .T tests properly too, or more generally can we just run all tests and get the same result that we get with the `make` build system? Some tests have `Makefile`s that rely on some env vars being set etc.

I got started looking into this in the context of https://github.com/snowleopard/hadrian/pull/445 but if you do most of the work, that would be great and I'll then just have to tweak a couple of things to make it work with that branch. So... thanks! :)

simonmar:
@alpmestan aren't you also working on this? cc @bgamari

alpmestan:
@simonmar I am indeed, I left a comment above to ask a question or two. This patch is most likely saving me quite some work.

snowleopard:
@alpmestan Ah, hopefully there wasn't too much work duplication. Note that this PR grew out of solving a bug in the test rule: #494. Perhaps, we should create a more general issue to coordinate the work on test build rules in Hadrian.

alpmestan:
@snowleopard Well if this PR gets us to a point where we can run tests with the same results that we get with the make build system, then I don't think there's much more to do for now, hence my questions above :)

I haven't done a lot of duplicate work, but I did some. It's not all lost though, as this allowed me to look at and understand the testing infra of GHC as well as more code in Hadrian.

sighingnow:
@alpmestan Currently I just have taken flags from `test.mk` to the `test` rule to execute the python script. 

Make hadrian's `test` and `validate` rules fully support GHC's testsuite needs working with GHC team to translate the `Makefile` script in some tests to hadrian/shake's rule. I have run `hadrian validate` and many testcases failed to run their `Makefile` scripts. [test summary log](https://gist.github.com/sighingnow/05bb9076ebaff16700035740be661651)

I'm willing to take the job to integrate hadrian to GHC's `test` and `validate`.

src/Rules/Test.hs
```diff
@@ -28,45 +28,116 @@ testRules = do
...
+
+            , "-e", "ghc_with_threaded_rts=0"               -- TODO: support threaded
+            , "-e", "ghc_with_dynamic_rts=0"                -- TODO: support dynamic
+            , "-e", "config.ghc_dynamic_by_default=False"   -- TODO: support dynamic
+            , "-e", "config.ghc_dynamic=False"              -- TODO: support dynamic
+
+            , "-e", "config.in_tree_compiler=True"          -- Use default value, see https://github.com/ghc/ghc/blob/master/testsuite/mk/boilerplate.mk
+
+            , "--config-file=testsuite/config/ghc"
+            , "--config", "compiler="     ++ show (top -/- compiler)
```
alpmestan: These two lines have been a problem for me, as the config file seems to take priority and it also defines a `config.compiler` field. @sighingnow Can you confirm this is not the case for you and the correct GHC is picked?

src/Rules/Test.hs
```diff
@@ -28,45 +28,116 @@ testRules = do
...
+            , ifMinGhcVer "801" "-fno-diagnostics-show-caret"
+            , pure "-dno-debug-output"
             ]
 
-            -- , "--config", "hp2ps="    ++ quote ("hp2ps")
-            -- , "--config", "hpc="      ++ quote ("hpc")
-            -- , "--config", "gs=$(call quote_path,$(GS))"
-            -- , "--config", "timeout_prog=$(call quote_path,$(TIMEOUT_PROGRAM))"
+        -- See: https://github.com/ghc/ghc/blob/master/testsuite/mk/test.mk#L291
+        let timeout_prog = "testsuite/timeout/install-inplace/bin/timeout"
```
alpmestan: Does this therefore rely on having the timeout prog make rules executed before running hadrian's `test` rule? I do have a rule or two to handle the timeout program in my branch. I don't properly implement the Windows logic yet, I always pick the python script, but that's a start.

src/Rules/Test.hs
sighingnow: I have confirm that this line let me execute the right `ghc` (inplace/bin/ghc-stage2) to compile the test.

Without this config, the `ghc` in `$PATH` will be called.

alpmestan:
@sighingnow I see! So we are both facing the same problem. One idea that I had yesterday to handle this problem without touching GHC's codebase at all might do the trick but requires going through a few hacky hoops.

What if we copied over the makefiles for all the tests that have one, along with the files needed for the test, and put them somewhere under the build root while following the same hierarchy. Except that we could edit the makefiles on the fly. As far as I can tell, many (most? all?) makefiles define `TOP`, include `boilerplate.mk` and `test.mk` and then define rules for the tests, which end up running simple commands like `$(TEST_HC) $(TEST_HC_OPTS) [...] foo.hs -o foo`. What if we simply stripped of the problematic bits (TOP and the includes) and instead defined TEST_HC, TEST_HC_OPTS to point to the GHC we've built and our config etc? I don't know how many such variables have to be defined, but I was looking around yesterday and I've really just seen these 2. I haven't looked everywhere though of course.

It sounds to me like such an approach could work and has the _huge_ advantage that this doesn't require a single change to GHC's codebase. I'm not sure defining TEST_HC and what not on the spot is the best option, but I do think it is necessary to work with variations on those `Makefile`s that don't rely on half of the make based build system, therefore the suggestion to strip off the includes. If we don't do this, then we don't have to just hook into the python driver correctly but also into the make build system, and I suspect this is something we want to avoid as much as possible, or even entirely.

src/Rules/Test.hs
sighingnow: If the `timeout_prog` is not set, I get a `[Error 13] Permission denied` in python script (inside the `runCmd` method).

I'm working with Bash on Windows, it's a Ubuntu 16.04 environment.

sighingnow: We have the following code in `testlib.py`:

```
        r = subprocess.Popen([timeout_prog, timeout, cmd],
                             stdin=stdin_file,
                             stdout=subprocess.PIPE,
                             stderr=hStdErr,
                             env=ghc_env)
```

alpmestan: Right, so this indeed requires that you build it through the make build system first. I have a bit of code that works on Linux for wrapping the python version of the program in a shell script and using that for the tests. It's a bit ugly but it does the trick. You'd need to replicate the Windows logic to make this work on your Windows machine, but as you can see [here](https://github.com/ghc/ghc/blob/master/testsuite/timeout/Makefile#L36) it's not scary enough to prevent us from doing it all in hadrian, I think. Here's the code that I have for the non-Windows case (that is, the part that replicates [this bit](https://github.com/ghc/ghc/blob/master/testsuite/timeout/Makefile#L23) of the Makefile).

``` haskell
-- FIXME: add ".exe" extension on windows
timeoutProgPath :: FilePath
timeoutProgPath = "test" -/- "bin" -/- "timeout"

timeoutPyPath :: FilePath
timeoutPyPath = "test" -/- "bin" -/- "timeout.py"

-- and then in `testRules`...
    root <- buildRootRules

    root -/- timeoutPyPath ~> do -- FIXME: run only when not on Windows
      copyFile "testsuite/timeout/timeout.py" (root -/- timeoutPyPath)

    root -/- timeoutProgPath ~> do
      need [ root -/- timeoutPyPath ]
      let script = unlines
            [ "#!/usr/bin/env sh"
            , "exec python3 $0.py \"$@\""
            ]
      liftIO $ do
        writeFile (root -/- timeoutProgPath) script
        cmd "chmod" [ "+x", root -/- timeoutProgPath ]
```

src/Rules/Test.hs
alpmestan: OK, somehow the `--config` options were not working for me, glad you got that to work!

sighingnow:
Yeah, we just need to remove the `include` in `Makefile` scripts and do some tricks in python's `run_command` to set necessary variables.

We could leave the `.T` file as it is and only modify the `Makefile` in every test cases.

src/Rules/Test.hs
sighingnow: If we can get ride of current python's test driver, we could do the `timeout` directly in Haskell.

alpmestan:
> Yeah, we just need to remove the include in Makefile scripts and do some tricks in python's run_command to set necessary variables.

Either that or define those variables in our modified `Makefile`s, or in the environment in which the python script is executed by Hadrian, if we want to leave GHC's testsuite driver code untouched (which would be a good thing IMO, but it's absolutely not necessary).

> We could leave the .T file as it is and only modify the Makefile in every test cases.

Right, but I think the makefiles rely on having the .hs (and others) files in the same directory, so we would probably have to move those over as well, without changing them though.

Alright @sighingnow, are you willing to give a shot at implementing this or should I?

src/Rules/Test.hs
alpmestan: I'm afraid the python test driver is not going anytime soon, and we might want to aim for a more reasonable goal of just replicating that tiny Makefile's logic in hadrian, so that we can at least properly build/wrap the appropriate timeout program but also `need` it in any rule that will run tests. Really just replicating what's done in the makefile I linked to but in hadrian, so that we don't introduce any (implicit) dependency on them make build system for running the test rules.

sighingnow:
I still don't think we need to move these files, before `run_command` the python driver has already changed to the right directory. For example:

```
cd "testsuite/tests/printer/Ppr017.run" && $MAKE -s --no-print-directory ppr017
```

> are you willing to give a shot at implementing this or should I?

Do you mean modify the test cases' `Makefile` and hack the environment variables before run test ?

I think we can pass necessary variables via system's environment variable. Define these variables in `Makefile` is not a good idea since we may want to run test with another ghc instance.

sighingnow:
> are you willing to give a shot at implementing this or should I?

@alpmestan I'm really not familiar with hadrian's codebase so it would be better if you could submit such a patch :)

I'll try to follow your work.

alpmestan:
> Do you mean modify the test cases' Makefile and hack the environment variables before run test ?

Well, anything that gets all the tests to be executed properly, really :)

alpmestan:
Just to clarify: I'm by no means an expert on Hadrian's code. I've just been working on/with it for several few weeks. I took the liberty to give some feedback and suggestions on ways to fix the test rules because I'm myself facing the same problem. I'm nowhere saying my suggestions are better, just trying to figure out a way to get hadrian to be a fine companion for running tests.

snowleopard:
> I'm really not familiar with hadrian's codebase
> I'm by no means an expert on Hadrian's code

@sighingnow @alpmestan I'm happy to be your go-to expert on Hadrian :) Feel free to submit half-cooked PRs for review and I'll suggest how to make the best use of Hadrian infrastructure and/or do some polishing after merging. Thank you for helping!

alpmestan:
@snowleopard For the record, @sighingnow and I had a little chat by email, and he's willing to take a stab at implementing something along the lines of what we have been suggesting above. By combining everyone's knowledge, fixes and experiments, we _will_ get that testsuite running smoothly sooner or later :)

(While you're around Andrey, I think it might be good that you and (Moritz and/or me) have a little discussion about https://github.com/snowleopard/hadrian/pull/445 and how we should go about merging most/all of that work -- should I start an email thread? Sorry for the off-topic question on this PR.)

snowleopard:
@alpmestan Awesome!

Sure, feel free to start an email thread to discuss #445. There is already a bit of a discussion in the PR's itself, so let's include all participants. We could also setup a Skype chat to speed things up.

sighingnow:
Update summary:

1. Move `TestArg` definition to `CommandLine.hs`.
2. Add `runTestBuilderArgs` under the `Settings/Builders` directory to collect all necessary arguments to run python driver.
3. Add timeout program target in `testRules`.
4. Set environment variables `TEST_HS` and `TEST_HC_OPTS` before run python script, Now the tests which use `Makefile` script can work after deleting `include` lines in their `Makefile`, (for example, `T2578` and `T14626` under `codeGen/should_compile`), without touching or moving the `.T` file.

Question:
1. I'm really not sure if I do the right things in `runTestBuilderArgs` in the newly created `RunTest.hs`.
2. Currently I need to compute `ghcFlags` twice, in `runTestBuilderArgs` and `testRules`. I need to pass these flags to python driver, as well as set these flags to `TEST_HC_OPTS` environment variables. How could I avoid the boilerplate code ?

/cc @snowleopard @alpmestan

alpmestan:
Regarding question number 1: the code so far looks reasonable. It is much nicer to have a proper "builder" for running the tests. If I were you I would not worry much more about that code at least until the tests all run properly. :)

Regarding question number 2, I think @snowleopard or someone else will have a better answer.

src/Rules/Test.hs
```diff
@@ -24,49 +46,51 @@ testRules = do
...
 
-            -- , "--config", "hp2ps="    ++ quote ("hp2ps")
-            -- , "--config", "hpc="      ++ quote ("hpc")
-            -- , "--config", "gs=$(call quote_path,$(GS))"
-            -- , "--config", "timeout_prog=$(call quote_path,$(TIMEOUT_PROGRAM))"
+        -- Set environment variables for test's Makefile.
+        liftIO $ do
+            setEnv "MAKE" makePath
+            setEnv "TEST_HC" ghcPath
+            setEnv "TEST_HC_OPTS" ghcFlags
```
snowleopard: We typically set the environment as follows: 

https://github.com/snowleopard/hadrian/blob/master/src/Rules/Gmp.hs#L41-L44

src/Rules/Test.hs
```diff
@@ -24,49 +46,51 @@ testRules = do
...
+
+        -- Execute the test target.
+        build $ target (vanillaContext Stage2 compiler) RunTest [] []
+
+timeoutPyPath :: FilePath
+timeoutPyPath = "test" -/- "bin" -/- "timeout.py"
```
snowleopard: In such cases we are usually more direct in Hadrian: `test/bin/timeout.py`. As far as I know, this doesn't cause any compatibility issues across different OSs.

src/Settings.hs
```diff
@@ -66,3 +66,23 @@ findPackageByName name = find (\pkg -> pkgName pkg == name) knownPackages
 -- | Install's DESTDIR setting.
 getDestDir :: Action FilePath
 getDestDir = fromMaybe "" <$> cmdInstallDestDir
+
+-- | Arguments to run GHC's test script.
+getTestArgs :: Args
```
snowleopard: Why is this defined here and not in `Settings.Builders.RunTest` together with the rest of the flags?

src/Settings/Builders/RunTest.hs
```diff
@@ -0,0 +1,113 @@
...
+            [ pure " -dcore-lint -dcmm-lint -no-user-package-db -rtsopts"
+            , pure ghcExtraFlags
+            , ifMinGhcVer "711" "-fno-warn-missed-specialisations"
+            , ifMinGhcVer "711" "-fshow-warning-groups"
+            , ifMinGhcVer "801" "-fdiagnostics-color=never"
+            , ifMinGhcVer "801" "-fno-diagnostics-show-caret"
+            , pure "-dno-debug-output"
+            ]
+
+        timeout_prog <- (-/- "test" -/- "bin" -/- "timeout") <$> expr buildRoot
```
snowleopard: This would be more typical: `expr buildRoot <&> "test/bin/timeout"`.

src/Settings/Builders/RunTest.hs
```diff
@@ -0,0 +1,113 @@
...
+        top      <- expr topDirectory
+        compiler <- expr $ builderPath $ Ghc CompileHs Stage2
+        ghcPkg   <- expr $ builderPath $ GhcPkg Update Stage1
+        haddock  <- expr $ builderPath $ Haddock BuildPackage
+        hp2ps    <- expr $ builderPath $ Hp2Ps
+        hpc      <- expr $ builderPath $ Hpc
+
+        let ifMinGhcVer ver opt = do v <- expr ghcCanonVersion
+                                     if ver <= v then pure opt
+                                                 else pure ""
```
snowleopard: Isn't this simply `(ver <= v) ? pure opt`?

snowleopard:
> Currently I need to compute `ghcFlags` twice

@sighingnow Can you simply export `ghcFlags` from `Settings.Builders.RunTest`? That's what I would do. I hope that doesn't lead to an import cycle.

Great job! I've left a few other minor comments above. In general, I think you sometimes could shorten the code by using the operator `?` for adding certain command line flags conditionally.

src/Rules/Test.hs
sighingnow: I have tried

```haskell
        -- Set environment variables for test's Makefile.
        env <- sequence
            [ builderEnvironment "MAKE" $ Make ""
            , builderEnvironment "TEST_HC" $ Ghc CompileHs Stage2
            , AddEnv "TEST_HC_OPTS" <$> runTestGhcFlags
            , AddEnv "MAKE" <$> builderPath (Make "") ]

        -- Execute the test target.
```

But it doesn't work. The env variable `$MAKE` is still undefined when the python driver invoke the `$MAKE` command.

src/Settings.hs
sighingnow: I define this method in `src/Setting.hs` to translate the arguments from command line to `Args` type.

Without this function I'll need to import module `CommandLine` into `Settings.Builders.RunTest` to do the transformation. I don't sure if it's the right way (import `CommandLine` into the builder module).

sighingnow: The `getTestArgs` returns the arguments provided by user from command line. In `Settings.Builders.RunTest` we only process the arguments decided by current GHC's configuration.

src/Rules/Test.hs
sighingnow: Fixed.

src/Settings/Builders/RunTest.hs
sighingnow: Fixed.

src/Settings/Builders/RunTest.hs
sighingnow: Now I move this port to `src/Rules/Tests.hs`. I still use `if ... then ... else` because operator `?` is only available for `Expr` monad, can't be used in `Action` monad.

sighingnow:
> Can you simply export `ghcFlags` from `Settings.Builders.RunTest`? That's what I would do. I hope that doesn't lead to an import cycle.

I have define `ghcFlags` in `Rules.Test` and import it in `Settings.Builders.RunTest` to make it less boilerplate.

src/Rules/Test.hs
snowleopard: Hmm, strange. Thanks for leaving a TODO message, I also suggest to open an issue so we don't forget about this -- the ability to set an environment for a builder is important and we should figure out the best approach for it.

src/Settings.hs
snowleopard: We use `Settings.Builders.X` to define all X-specific arguments. This seems to be relevant only for `RunTest`.

snowleopard: > I don't sure if it's the right way (import CommandLine into the builder module).

Yes, that's fine, as long as this doesn't create nasty import cycles. In fact, we do something very similar in `Settings.Builders.Make` where we rely on `shakeThreads <$> expr getShakeOptions` which essentially comes from Hadrian's command line (but captured by Shake).

snowleopard:
@sighingnow Many thanks! I think we are getting close to merging this. My apologies for slow responses, I am travelling this week.

src/Settings.hs
sighingnow: I have moved the `getTestArgs` from `Settings` into `Settings.Builders.RunTest` and deleted the unused `cmdTestArgs` function.

src/Builder.hs
```diff
@@ -95,10 +96,13 @@ data Builder = Alex
              | Objdump
              | Patch
              | Perl
+             | Python
              | Ranlib
+             | RunTest
              | Sphinx SphinxMode
              | Tar TarMode
              | Unlit
+             | Validate
```
snowleopard: This seems to be irrelevant for this particular PR? I suggest to drop this to avoid confusion.

src/Rules/Test.hs
```diff
@@ -1,17 +1,37 @@
...
+    root -/- timeoutProgPath ~> do
+        python <- builderPath Python
+        need [root -/- timeoutPyPath]
+        let script = unlines
+                [ "#!/usr/bin/env sh"
+                , "exec " ++ python ++ " $0.py \"$@\""
+                ]
+        liftIO $ do
+            writeFile (root -/- timeoutProgPath) script
+            cmd "chmod" [ "+x", root -/- timeoutProgPath ]
```
snowleopard: There is a utility called `makeExecutable` that does this.

src/Rules/Test.hs
```diff
@@ -24,49 +44,62 @@ testRules = do
...
+        , ifMinGhcVer "801" "-fdiagnostics-color=never"
+        , ifMinGhcVer "801" "-fno-diagnostics-show-caret"
+        , pure "-dno-debug-output"
+        ]
+
+timeoutPyPath :: FilePath
+timeoutPyPath = "test/bin/timeout.py"
+
+timeoutProgPath :: FilePath
+timeoutProgPath = "test/bin/timeout"  -- TODO: `.exe` suffix for windows.
```
snowleopard: You can do `<.> exe`.

src/Settings/Builders/RunTest.hs
```diff
@@ -0,0 +1,111 @@
...
+                , arg "--config", arg $ "hpc="          ++ show (top -/- hpc)
+                , arg "--config", arg $ "gs=gs"                           -- Use the default value as in test.mk
+                , arg "--config", arg $ "timeout_prog=" ++ show (top -/- timeoutProg)
+                , arg $ "--threads=" ++ show threads
+                , arg $ "--verbose=" ++ show (fromEnum verbose)
+                , getTestArgs -- User-provided arguments from command line.
+                ]
+
+    -- TODO
+    , builder Validate ? pure [] ]
```
snowleopard: Not sure what should be done here, so I suggest to drop this bit.

snowleopard:
@sighingnow I've left a few more comments.

At the top of the PR one of your tasks is to add docs describing how to run tests. Are you going to add these docs as part of this PR? Or do you prefer to do it in a separate PR?

Thank you!

src/Rules/Test.hs
sighingnow: `<.> exe` will add `.exe` extension to the file[1], however on linux/unix system we usually don't use specific extension for executable.

1. http://hackage.haskell.org/package/filepath-1.4.2/docs/System-FilePath-Windows.html#v:-60-.-62-

snowleopard: It's OS-specific exactly for this reason: https://hackage.haskell.org/package/shake-0.16/docs/Development-Shake-FilePath.html#v:exe

snowleopard: That is, on Linux `"foo" <.> exe` should remain `"foo"`.

sighingnow:
I have read the task list in #187 . I prefer to add doc after #187 get fixed.

Many TODOS still exists in this PR. I need more time to understanding other stuffs in GHC's test script, such as `--way=...` and finish hadrain's 'test' and `validate` rules.

src/Rules/Test.hs
sighingnow: Oops, my bad.

sighingnow: Done.

sighingnow:
I suggest to use the **Squash and merge** option when merge this PR to make a clean diff for others who also is interested in this work. Thanks!

@snowleopard

snowleopard:
@sighingnow Many thanks, I've squashed & merged this. Looking forward to seeing more PRs from you :)

PR #496, opened 2018-02-03
Fix Haddock

### LINKED ISSUE #492: Documentation broken by upstream Haddock change
Haddock commit https://github.com/haskell/haddock/commit/e99aefb broke Hadrian's ability to generate the documentation. In particular, the file `haddock-util.js` was removed from the repository. That file is currently being used as the target of `need`, which causes documentation to break.

https://github.com/snowleopard/hadrian/blob/7d2368d714c7ed3fb8c096882564575aacf2b396/src/Rules/Documentation.hs#L120-L121

https://github.com/snowleopard/hadrian/blob/7d2368d714c7ed3fb8c096882564575aacf2b396/src/Rules/Documentation.hs#L137-L141

Haddock uses all the files in that folder to generate and style the web pages. It contains both style themes and JavaScript for functionality. Because Haddock requires that folder to run (the wrapper script explicitly names it), I propose copying it when Haddock is built rather than `need`ing it when building the documentation.

The fastest solution is to just point `haddockHtmlLib` at a different file, but that feels a little brittle to me.

snowleopard:
Fix #492 and then attempt https://phabricator.haskell.org/D4365.

### DIFF
--- a/src/Rules/Documentation.hs
+++ b/src/Rules/Documentation.hs
@@ -117,8 +117,11 @@ allHaddocks = do
     sequence [ pkgHaddockFile $ vanillaContext Stage1 pkg
              | pkg <- pkgs, isLibrary pkg, isHsPackage pkg ]
 
-haddockHtmlLib :: FilePath
-haddockHtmlLib = "inplace/lib/html/haddock-util.js"
+-- TODO: This is fragile and will break if @README.md@ is removed. We need to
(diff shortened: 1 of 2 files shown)


snowleopard:
Fixing https://phabricator.haskell.org/D4365 [didn't work](https://phabricator.haskell.org/D4365#121780), so I'm merging as is to close #492.

PR #499, opened 2018-02-08
Build timeout program with `buildProgram` rules on windows.

sighingnow:
Hi, @snowleopard, I'm working on the timeout program in testsuite on windows. I'm trying to build the `testsuite/timeout` using the `buildProgram` rules, just like other utils programs.

I have added the program definition in `GHC.hs`:

~~~haskell
timeout             = hsPrg  "timeout"         `setPath` "testsuite/timeout"
~~~

However I'm lost in the following error message when execute `./hadrian-ghc.sh inplace/bin/timeout.exe`:

~~~
* inplace/bin/timeout.exe
* _build/stage0/testsuite/timeout/WinCBindings.o
* _build/stage0/testsuite/timeout/WinCBindings.o _build/stage0/testsuite/timeout/WinCBindings.hi
* OracleQ (KeyValues ("_build/stage0/testsuite/timeout/.dependencies","_build/stage0/testsuite/timeout/WinCBindings.o"))
* _build/stage0/testsuite/timeout/.dependencies
* _build/stage0/testsuite/timeout/Main.hs
No generator for _build/stage0/testsuite/timeout/Main.hs.
CallStack (from HasCallStack):
  error, called at src\Rules\Generate.hs:108:38 in main:Rules.Generate
~~~

Could you provide me some hints about this problem ?

### DIFF
--- a/src/GHC.hs
+++ b/src/GHC.hs
@@ -109,7 +109,8 @@ testsuitePackages = return [ checkApiAnnotations
                            , checkPpr
                            , ghcPkg
                            , parallel
-                           , hp2ps              ]
+                           , hp2ps
+                           , timeout         ]
 
 -- | Given a 'Context', compute the name of the program that is built in it
 -- assuming that the corresponding package's type is 'Program'. For example, GHC
--- a/src/GHC/Packages.hs
+++ b/src/GHC/Packages.hs
@@ -17,7 +17,8 @@ ghcPackages =
     , ghcHeap, ghci, ghcPkg, ghcPrim, ghcTags, haddock, haskeline, hsc2hs, hp2ps
     , hpc, hpcBin, integerGmp, integerSimple, iserv, libffi, libiserv, mtl
     , parsec, parallel, pretty, process, rts, runGhc, stm, templateHaskell
-    , terminfo, text, time, touchy, transformers, unlit, unix, win32, xhtml ]
+    , terminfo, text, time, touchy, transformers, unlit, unix, win32, xhtml
+    , timeout ]
 
 -- TODO: Optimise by switching to sets of packages.
 isGhcPackage :: Package -> Bool
@@ -81,6 +82,7 @@ unlit               = hsUtil "unlit"
 unix                = hsLib  "unix"
 win32               = hsLib  "Win32"
 xhtml               = hsLib  "xhtml"
+timeout             = hsUtil "timeout"         `setPath` "testsuite/timeout"
 
 -- | Construct a Haskell library package, e.g. @array@.
 hsLib :: PackageName -> Package
--- a/src/Rules/Test.hs
+++ b/src/Rules/Test.hs
@@ -3,6 +3,7 @@ module Rules.Test (testRules, runTestGhcFlags, timeoutProgPath) where
 import Base
 import Expression
 import GHC
+import GHC.Packages (timeout)
 import Oracles.Flag
 import Oracles.Setting
 import Settings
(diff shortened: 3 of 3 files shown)


angerman:
Let's look at the timeout folder first:
```
Makefile          Setup.hs          TimeMe.hs         WinCBindings.hsc  calibrate         timeout.cabal     timeout.hs        timeout.py
```
what happens no is the following, we try to build `inplace/bin/timeout.exe`, which looks into the cabal file, and finds it needs `_build/stage0/testsuite/timeout/WinCBindings.o`.

we keep on chasing dependencies, end end up looking for `timeout/Main.hs`, however the `timeout.cabal` does specify it's `main` as `timeout.hs` and not `Main.hs` (which hadrian here seems to simply assume).

And as such it just falls over.

You could probably try renaming `timeout.hs` to `Main.hs` and see if hadrian is happy for your local build. But the correct solution is to teach hadrian about no-Main.hs main modules.

sighingnow:
Thanks @angerman, It does work!

Now the problem is to teach hadrian read `main-hs` from Cabal file. I'll do this!

src/GHC.hs
```diff
@@ -162,6 +163,7 @@ stage0Packages = do
              , unlit                       ]
           ++ [ terminfo | not win, not ios, not cross ]
           ++ [ touchy   | win              ]
+          ++ [ timeout ]
```
snowleopard: Does the Make build system always build `timeout` or only when running tests?

snowleopard:
@sighingnow Thank you, it would be great to add support for parsing `main-is` from Cabal files! If you get stuck, don't hesitate to get in touch.

src/GHC.hs
sighingnow: I just trying to make it buildable.... Indeed I'm not very sure what I was doing...

angerman:
@sighingnow the reloc branch might be doing that already. For the current master branch it might be a combination of verifying the data that `ghc-cabal` produces contains the proper main-module (which I expect it should), and then modifying the `cabal` oracle to actually extract the main module properly and use it instead of assuming it's `Main`.

sighingnow:
@angerman Thanks for pointing me there. The reloc branch have considered more cases about `main-is` than I have done. Once it merged to master, I will rebase this PR.

angerman:
@sighingnow if you feel like opening a PR against the reloc branch, feel free to do so.

If you feel remotely inclined to chip some of the reloc branch stuff off into a smaller self contained PR against master, please also feel free to do so. Merging of the reloc branch is currently mostly blocked on it being just a massive change. If we can chip of smaller self contained parts into PR's against the master branch, reviewing and mergeing the reloc branch becomes easier. @alpmestan is also looking into that.

src/GHC.hs
alpmestan: It's only used for the tests indeed, we might get away with simply throwing the rules for the timeout program in `Rules/Test.hs` if they're not too complicated (the non-windows case is trivial, but the windows case a little less so).

snowleopard: This is fine for debugging to cause `timeout` to be built by default. But I agree with @alpmestan that when the build works, we'll just need to add `timeout` to the list of all packages and `need` it from the test rule.

snowleopard:
I second @angerman's suggestion: it would be great to factor out the `main-is` bits from the reloc branch to simplify its review and merge.

snowleopard:
@sighingnow Are you planning to continue working on this PR? It's been accumulating merge conflicts for some time now. Shall we revive it?

sighingnow:
I will update this PR this weekends to resolve the merge conflicts and make sure it works.

snowleopard:
@sighingnow Awesome, thanks!

sighingnow:
I have resolved the conflicts and refactored the changes into a single commit. Now this PR is blocked by #627, which fix the `main-is` trouble from the `testsuite/timeout` program.

Now this PR is ready to review. I have tested `hadrian test` on my laptop and everything works fine. I believe the failures on travis-ci, as well as appveyor are't related to this PR. The ghc-8.2.2 build case is green on travis-ci.

/cc @snowleopard @alpmestan

sighingnow:
I think [Phab:4212](https://phabricator.haskell.org/D4212#change-qfmxhVHXQ4A6) may be related to the failure.

I have also noticed that we use `integer-simple` for ghc-8.4.1, but not for ghc-8.2.2. I will try local build with ghc-8.4.2 and hadrian.

sighingnow:
However I did't experience the same failure on my local Ubuntu with ghc-8.4.2.....

sighingnow:
Move the discussion about the ghc-8.4.2 failure to #629.

sighingnow:
Now #627 is merged so we can land this as well.

snowleopard:
@sighingnow Thank you, merged!

PR #503, opened 2018-02-15
Relax Cabal constraint

angerman:
As the Cabal 2.2 release branch has been cut, we need to relax the constraint on Cabal in hadrian.


snowleopard:
Thanks, merged!

PR #505, opened 2018-02-18
Add --quickjump to Haddock

snowleopard:
Also reorder other flags as in the Make build system

See https://phabricator.haskell.org/D4365


snowleopard:
This works fine on my Windows machine.

PR #507, opened 2018-02-20
Undo iserv changes

snowleopard:
See #504 and https://phabricator.haskell.org/D4377


snowleopard:
@bgamari I'm undoing the corresponding changes in Hadrian. It's fairly easy to bring this back once D4377 is finalised.

PR #511, opened 2018-02-23
Pin nixpkgs and all-cabal-hashes in shell.nix

mpickering:
This will make builds more reproducible and also fixes the hadrian build.


mpickering:
@sgraf812 @alpmestan

sgraf812:
Wow, that actually fixes the hadrian build! Thank you

PR #525, opened 2018-03-10
Added libraries for validate

chitrak7:


chitrak7:
Added libraries check-ppr and check-api-annotations to stage2 packages.

src/GHC.hs
```diff
@@ -199,7 +201,7 @@ stage1Packages = do
           ++ [ win32    | win     ]
 
 stage2Packages :: Action [Package]
-stage2Packages = return [haddock]
+stage2Packages = return [haddock, checkPpr, checkApiAnnotations]
```
snowleopard: We don't want to build these two packages by default, so they shouldn't be added here.

Instead, `need` then in the `validate` rule.

snowleopard:
Thanks @chitrak7! See my comment above. Adding these two utilities into the default build will slow down Hadrian unnecessarily. We only want to build them in the `validate` rule, before running the testsuite.

src/GHC.hs
chitrak7: I think that intead, I can export packages as an [package] and make it as a need for  validate rule. It should do this for us.

snowleopard: @chitrak7 Are you saying you'd like to do something like this?
```haskell
-- In GHC.hs
testsuitePackages :: [Package]
testsuitePackages = [checkPpr, checkApiAnnotations]

-- Rules/Test.hs
...
needPackages testsuitePackages
...
```
I think I like this. You'll need to define a helper function `needPackages` for that.

chitrak7: Yes, exactly what i wish to do. Will implement this after recent pulls are merged

snowleopard: Great, thank you!

snowleopard:
@chitrak7 Can you rebase this PR and implement the above suggestion? I'd like to merge this so we can reduce the number of `validate` failures.

chitrak7:
@snowleopard Can I do this task on 1st as I have my end semesters going on now.

snowleopard:
@chitrak7 Sure, no problem!

This is a small fix so if we get blocked on this, we'll just complete this PR ourselves and merge it.

Issue #523, opened 2018-03-05
Error: at least ftruncate or _chsize functions are required to build


izgzhen:
```
| Run Ghc LinkHs Stage0: _build/stage0/utils/ghc-pkg/Main.o (and 1 more) => inplace/lib/bin/ghc-pkg
[...]

PR #526, opened 2018-03-13
Add extra include paths when invoking ghc-cabal

snowleopard:
See #523


snowleopard:
So, I think this has worked, but we hit another unrelated issue.

Issue #534, opened 2018-03-20
Document how to turn off dynamic linking


alpmestan:
Following up with Ben's request from [here](https://ghc.haskell.org/trac/ghc/ticket/10536#comment:8), I looked into how we can achieve the same thing with hadrian. If I'm not mistaken, all we have to do is set these two fields from the `Flavour` data type (and possibly a little more, `dynamicGhcPrograms` to false maybe?) as we please:

``` haskell
    -- | Build libraries these ways.
    libraryWays :: Ways,
    -- | Build RTS these ways.
    rtsWays :: Ways,
```

in the `UserSettings.hs` file to define a `dynamic`-less flavour from scratch or by changing an existing one. Maybe we could augment `flavours.md` and/or `user-settings.md` with a tiny section about this, with an example that e.g takes "quick" and turns off dynamic linking, hence not building anything the dyn way. This would be the best answer hadrian can provide right? Any objection to me going to a documentation PR that adds such a section?

(I guess we could provide a flag for this particular need, but then why not for when we want to disable prof and what not? If we start it's hard to stop. Plus, teaching how to tweak `UserSettings.hs` for this purpose would be a nice way for users to get to know some really basic hadrian types and functions, which can't hurt if one day they want to extend hadrian or fix a problem.)

snowleopard:
@alpmestan Do be honest I never looked dynamic linking in detail, so my understanding of this part of Hadrian is vert hazy. @izgzhen did some work on this as part of his project, perhaps, he could comment?

snowleopard:
> Any objection to me going to a documentation PR that adds such a section?

Of course, no objections from me on improving documentation! That would be awesome :-)

If the solution you suggested works as intended, then by all means document it providing some examples for users.

izgzhen:
@alpmestan It is true that the `libraryWays` and `rtsWays` in `Flavour` is a proper way to users to turn on/off dynamic compilation, and it is useful to add more documentation as well. Thanks!

alpmestan:
Turns out most of what I had in mind is already explained in user-settings.md (I had not read it carefully before, I got most of that knowledge from looking at the code directly). PR at #535.

alpmestan:
The PR has been merged, closing this issue now.

PR #535, opened 2018-03-21
docs: add user-defined flavour example for turning off dynamic linking

alpmestan:
See #534 and https://ghc.haskell.org/trac/ghc/ticket/10536
Turns out most of the content I was thinking of was already explained, so I just kept it simple.

cc @snowleopard @izgzhen


snowleopard:
@alpmestan Thank you! Merged.

PR #542, opened 2018-03-31
Move rts I/O Functions in different namespace.

Mistuke:
Hi,

I'll be landing https://phabricator.haskell.org/D4416 soon but that requires some new build arguments for the `rts` due to Linux. The build will fail without this change.

The flag controls name-mangling of some `I/O` functions. Without it they'll be in the default namespace and the build will fail with an unknown symbol.

Also I noticed `xxhash.c`'s optimization flags were not the same as Make, so I've corrected those.


angerman:
LGTM.

angerman:
Build failures are not due to this PR.

snowleopard:
@Mistuke Many thanks for the PR! Hope we'll soon fix Hadrian on all major platforms and will have reliable CI back.

@angerman And thank you for reviewing and merging.

Issue #544, opened 2018-04-02
clean up warnings


izgzhen:
as of https://github.com/snowleopard/hadrian/commit/9a46249e974c99a2015d7bba514b252c31ff8575

```
src/CommandLine.hs:14:1: warning: [-Wunused-imports]
    The qualified import of ‘UserSettings’ is redundant
      except perhaps to import instances from ‘UserSettings’
    To import instances alone, use: import UserSettings()
   |
14 | import qualified UserSettings
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[37 of 86] Compiling Settings.Flavours.Quickest ( src/Settings/Flavours/Quickest.hs, dist/build/hadrian/hadrian-tmp/Settings/Flavours/Quickest.o )
[38 of 86] Compiling Settings.Flavours.QuickCross ( src/Settings/Flavours/QuickCross.hs, dist/build/hadrian/hadrian-tmp/Settings/Flavours/QuickCross.o ) [Expression changed]
[39 of 86] Compiling Settings.Flavours.Quick ( src/Settings/Flavours/Quick.hs, dist/build/hadrian/hadrian-tmp/Settings/Flavours/Quick.o ) [Expression changed]
[40 of 86] Compiling Settings.Flavours.Profiled ( src/Settings/Flavours/Profiled.hs, dist/build/hadrian/hadrian-tmp/Settings/Flavours/Profiled.o ) [Expression changed]
[41 of 86] Compiling Settings.Flavours.Performance ( src/Settings/Flavours/Performance.hs, dist/build/hadrian/hadrian-tmp/Settings/Flavours/Performance.o ) [Expression changed]
[42 of 86] Compiling Settings.Flavours.Development ( src/Settings/Flavours/Development.hs, dist/build/hadrian/hadrian-tmp/Settings/Flavours/Development.o ) [Expression changed]
[43 of 86] Compiling Settings         ( src/Settings.hs, dist/build/hadrian/hadrian-tmp/Settings.o )
[44 of 86] Compiling Settings.Warnings ( src/Settings/Warnings.hs, dist/build/hadrian/hadrian-tmp/Settings/Warnings.o )
[45 of 86] Compiling Settings.Packages.Rts ( src/Settings/Packages/Rts.hs, dist/build/hadrian/hadrian-tmp/Settings/Packages/Rts.o )
[...]

PR #547, opened 2018-04-03
Fix warnings

snowleopard:
See #544

### DIFF
--- a/src/GHC.hs
+++ b/src/GHC.hs
@@ -156,19 +156,6 @@ autogenPath context@Context {..}
   where
     autogen dir = contextPath context <&> (-/- dir -/- "autogen")
 
--- ref: mk/config.mk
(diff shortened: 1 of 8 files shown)


src/Hadrian/Haskell/Cabal/Parse.hs
```diff
@@ -1,3 +1,4 @@
+{-# OPTIONS_GHC -Wno-deprecations #-}
```
snowleopard: See #548.

src/GHC.hs
```diff
@@ -156,19 +156,6 @@ autogenPath context@Context {..}
   where
     autogen dir = contextPath context <&> (-/- dir -/- "autogen")
 
--- ref: mk/config.mk
```
snowleopard: See #549.

Issue #543, opened 2018-04-02
ghc-pkg register failed with "already installed"


izgzhen:
```
| Run GhcPkg Dependencies Stage0: integer-gmp
rts-1.0: package(s) with this id already exist: rts
rts-1.0: package rts-1.0 is already installed
rts-1.0: Package names may be treated case-insensitively in the future.
Package rts-1.0 overlaps with: rts-1.0 (use --force to override)
shakeArgsWith   0.000s    0%                           
Function shake  8.289s   50%  =========================
Database read   0.003s    0%                           
With database   0.000s    0%                           
Running rules   8.154s   49%  ======================== 
Total          16.446s  100%                           
Error when running Shake build system:
* _build/stage1/bin/unlit
* OracleQ (PackageDataFile (Context {stage = Stage1, package = Package {pkgLanguage = Haskell, pkgType = Program, pkgName = "unlit", pkgPath = "utils/unlit"}, way = v}))
* _build/stage1/utils/unlit/setup-config
* _build/stage0/bin/ghc
* OracleQ (PackageDataFile (Context {stage = Stage0, package = Package {pkgLanguage = Haskell, pkgType = Program, pkgName = "ghc-bin", pkgPath = "ghc"}, way = v}))
* _build/stage0/ghc/setup-config
* _build/stage0/lib/package.conf.d/ghc-8.5.conf
* OracleQ (PackageDataFile (Context {stage = Stage0, package = Package {pkgLanguage = Haskell, pkgType = Library, pkgName = "ghc", pkgPath = "compiler"}, way = v}))
* _build/stage0/compiler/setup-config
* _build/stage0/lib/package.conf.d/transformers-0.5.5.0.conf
* _build/stage0/lib/package.conf.d/base-4.10.1.0.conf
* _build/stage0/lib/package.conf.d/rts.conf
user error (Development.Shake.cmd, system command failed
Command: /usr/local/bin/ghc-pkg --global-package-db _build/stage0/lib/package.conf.d register -v0 -
Exit code: 1
Stderr:
rts-1.0: package(s) with this id already exist: rts
rts-1.0: package rts-1.0 is already installed
rts-1.0: Package names may be treated case-insensitively in the future.
Package rts-1.0 overlaps with: rts-1.0 (use --force to override)
)
```

probably one of #540

izgzhen:
there are current two solutions as I see:

1. try to `unregister` and then `register`
2. try to `register`
[...]

PR #552, opened 2018-04-03
Unregister stage0 package first if it needs to be cloned

izgzhen:
fixs #543

### DIFF
--- a/src/Rules/Register.hs
+++ b/src/Rules/Register.hs
@@ -83,8 +83,15 @@ copyConf rs context@Context {..} conf = do
     depPkgIds <- fmap stdOutToPkgIds . askWithResources rs $
       target context (GhcPkg Dependencies stage) [pkgName package] []
     need =<< mapM (\pkgId -> packageDbPath stage <&> (-/- pkgId <.> "conf")) depPkgIds
-    buildWithResources rs $
-      target context (GhcPkg Clone stage) [pkgName package] [conf]
+    -- we should unregister if the file exists since ghc-pkg will complain
+    -- about existing pkg id (See https://github.com/snowleopard/hadrian/issues/543)
+    -- also, we don't always do the unregistration + registration to avoid
+    -- repeated work after a full build
+    unlessM (doesFileExist conf) $ do
+      buildWithResources rs $
+        target context (GhcPkg Unregister stage) [pkgName package] []
+      buildWithResources rs $
+        target context (GhcPkg Clone stage) [pkgName package] [conf]
 
   where
     stdOutToPkgIds :: String -> [String]
--- a/src/Builder.hs
+++ b/src/Builder.hs
@@ -65,6 +65,7 @@ instance NFData   GhcCabalMode
 data GhcPkgMode = Init         -- initialize a new database.
                 | Update       -- update a package.
                 | Clone        -- clone a package from one pkg database into another. @Copy@ is already taken by GhcCabalMode.
+                | Unregister   -- unregister a package
                 | Dependencies -- compute package dependencies.
                 deriving (Eq, Generic, Show)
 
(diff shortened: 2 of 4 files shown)


src/Rules/Register.hs
```diff
@@ -83,8 +83,11 @@ copyConf rs context@Context {..} conf = do
     depPkgIds <- fmap stdOutToPkgIds . askWithResources rs $
       target context (GhcPkg Dependencies stage) [pkgName package] []
     need =<< mapM (\pkgId -> packageDbPath stage <&> (-/- pkgId <.> "conf")) depPkgIds
-    buildWithResources rs $
-      target context (GhcPkg Clone stage) [pkgName package] [conf]
+    unlessM (doesFileExist conf) $ do
+      buildWithResources rs $
+        target context (GhcPkg Unregister stage) [pkgName package] []
```
angerman: Why do we need to unregister the package if the file doesn't exist? How would the package end up being registered even if the file does not exist?

I would have thought we wanted:
```haskell
whenM (doesFileExist conf) {- unregister -}
{- register -}
```

angerman: Looking at #543 I kind of see the reasoning for the unconditional `unregister` now.
That might warrant a comment though.

izgzhen: > Why do we need to unregister the package if the file doesn't exist? How would the package end up being registered even if the file does not exist?

only if you deliberately deleted `[pkg-id].conf` file and then `register`. I think it is a rare case though...

angerman: I'm also unsure if we just want to skip this, or rather always *reregister* the package. In case the `conf` file has changed, even though we likely want to assume that no-one really changes the bootstrap compiler while building right? In that case we can just always unregister+register.

izgzhen: > just always unregister+register.

the problem is if so, when you run `./build.sh` after a full build, the `unregister+register`s will be repeated, which doesn't look sane to me.

angerman: Can we have Shake somehow verify that the `.conf` file is identical?

izgzhen: > Can we have Shake somehow verify that the .conf file is identical?

Do you mean the result of `describe`? I doubt if this kind of `Stdout` can be tracked...

angerman: Alright. Let's just add a comment why we do the unreg+reg (e.g. the comment from #543) and what the justification for not re-registering (full build should not repeatedly unregister/register).

alpmestan:
At least 2 travis builds fail with:

```
| Copy file: _build/stage1/libffi/build/inst/include/ffi.h => _build/stage1/rts/build/ffi.h
| Copy file: _build/stage1/libffi/build/inst/include/ffitarget.h => _build/stage1/rts/build/ffitarget.h
| Copy file (untracked): _build/stage1/libffi/build/inst/lib/libffi.a => _build/stage1/rts/build/libCffi.a
| Copy file (untracked): _build/stage1/libffi/build/inst/lib/libffi.a => _build/stage1/rts/build/libCffi_l.a
| Copy file (untracked): _build/stage1/libffi/build/inst/lib/libffi.a => _build/stage1/rts/build/libCffi_debug.a
| Copy file (untracked): _build/stage1/libffi/build/inst/lib/libffi.a => _build/stage1/rts/build/libCffi_thr.a
| Copy file (untracked): _build/stage1/libffi/build/inst/lib/libffi.a => _build/stage1/rts/build/libCffi_thr_debug.a
| Copy file (untracked): _build/stage1/libffi/build/inst/lib/libffi.a => _build/stage1/rts/build/libCffi_thr_l.a
| Successfully built custom library 'libffi'
Installing library in /home/travis/build/snowleopard/hadrian/ghc/_build/stage1/lib/../lib/x86_64-linux-ghc-8.5.20180402/rts-1.0
hadrian: can't find include file rts/fs.h
shakeArgsWith     0.000s    0%                           
Function shake   14.669s    1%                           
Database read     0.000s    0%                           
With database     0.001s    0%                           
Running rules  1314.416s   98%  ======================== 
Total          1329.086s  100%                           
* _build/stage1/lib/package.conf.d/rts-1.0.conf
ExitFailure 1
```

snowleopard:
@izgzhen Thank you! Let me merge this to make quicker progress, but at some point we might come back to this to see if we can simplify the logic.

PR #553, opened 2018-04-03
Fix Circle CI

izgzhen:


izgzhen:
it made to the final step before following failure:

```
hadrian: can't find include file rts/fs.h

shakeArgsWith    0.000s    0%                           
Function shake  17.862s    2%                           
Database read    0.000s    0%                           
With database    0.000s    0%                           
Running rules  722.812s   97%  ======================== 
Total          740.675s  100%                           
* _build/stage1/lib/package.conf.d/rts-1.0.conf
ExitFailure 1
```

alpmestan:
Note that the travis builds from one of your other PRs fails with the same problem, so this is probably not due to anything you're doing in the Circle CI script.

snowleopard:
@izgzhen Thanks! Is there a way to somehow record the `happy >= 1.19.4` requirement in the Cabal or Stack yaml files? Presumably, someones local build could also be failing because of this.

Anyway, I'm merging this for now.

PR #558, opened 2018-04-03
Preliminary bindist

alpmestan:
This is #555 with the missing `src/Rules/BinaryDist.hs`, sorry for the inconvenience @snowleopard and others!


snowleopard:
@alpmestan Thanks! And don't worry -- we need to fix the CI asap to avoid such glitches :-)

Issue #570, opened 2018-04-16
Touchy missing on Windows


snowleopard:
After #568 we have a new Windows-specific failure:

```
Command: _build/stage0/bin/ghc.exe -Wall -hisuf hi -osuf o -hcsuf hc -static -hide-all-packages -no-user-package-db "-package-db _build/stage1/lib/package.conf.d" "-this-unit-id base-4.11.1.0" "-package-id ghc-prim-0.5.2.0" "-package-id integer-gmp-1.0.2.0" "-package-id rts-1.0" -i -i_build/stage1/libraries/base/build -i_build/stage1/libraries/base/build/autogen -ilibraries/base/. -Iincludes -I_build/generated -I_build/stage1/libraries/base/build -I_build/stage1/libraries/base/build/include -Ilibraries/base/include -IC:/ghc/_build/stage1/x86_64-windows-ghc-8.5.20180413/integer-gmp-1.0.2.0/include -IC:/ghc/_build/stage1/x86_64-windows-ghc-8.5.20180413/rts-1.0/include -I_build/generated -optc-I_build/generated -optP-include -optP_build/stage1/libraries/base/build/autogen/cabal_macros.h -optc-fno-stack-protector -optP-DOPTIMISE_INTEGER_GCD_LCM -odir _build/stage1/libraries/base/build -hidir _build/stage1/libraries/base/build -stubdir _build/stage1/libraries/base/build -Wnoncanonical-monad-instances -optc-Wno-error=inline -c libraries/base/GHC/Real.hs-boot -o _build/stage1/libraries/base/build/GHC/Real.o-boot -O0 -H64m -this-unit-id base -XHaskell2010 -ghcversion-file=C:/ghc/_build/generated/ghcversion.h -Wno-deprecated-flags -Wno-trustworthy-safe
Exit code: 1
Stderr:
ghc.exe: could not execute: C:\ghc\_build\stage0\lib/bin/touchy.exe
)
```

Looks like in #531 `touchy` was relocated to the `bin` directory. I guess we should bring it back to `lib/bin`.

angerman:
I'd rather argue that we should locate `touchy.exe` in `bin` rather than `lib`.

snowleopard:
@angerman I agree, but this requires changes to GHC.

In the meantime I suggest we follow what GHC does. It looks a bit inconvenient but not a big deal.

alpmestan:
[...]

PR #572, opened 2018-04-16
Fix the path to touchy

snowleopard:
See #570


snowleopard:
OK, the `touchy` problem is fixed, but we hit yet another one:
```
* _build/stage1/lib/package.conf.d/ghc-8.5.conf
* _build/stage1/compiler/build/libHSghc-8.5-0.a
  _build/stage1/compiler/build/libHSghc-8.5-0.a
CallStack (from HasCallStack):
  error, called at src\Development\Shake\Internal\Rules\File.hs:171:58 in shake-0.16-BDmtKOMxPdtLRnDkDJohXq:Development.Shake.Internal.Rules.File
```

PR #583, opened 2018-04-26
fix missing symbols when cross-compiling

### LINKED ISSUE #581: missing symbols defined in includes/rts/Bytecodes.h when cross-compiling
according to steps in `docs/cross-compile.md`

izgzhen:
fixes #581

### DIFF
--- a/src/Settings/Builders/Common.hs
+++ b/src/Settings/Builders/Common.hs
@@ -29,9 +29,7 @@ cIncludeArgs = do
     gmpIncludeDir   <- getSetting GmpIncludeDir
     ffiIncludeDir   <- getSetting FfiIncludeDir
 
-    cross   <- expr crossCompiling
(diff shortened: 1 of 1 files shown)


izgzhen:
I am actually reverting a change I've done in https://github.com/snowleopard/hadrian/pull/401/files. I tried to understand what is my motivation but found it hard to justify this by comparing to the latest HEAD. I think we can merge this as long as CI passed and no one jumps out and objects it.

snowleopard:
I think we should just make CI fail if there are any warnings. In this case there are warnings about unused code, which is difficult to spot while reviewing a PR, but easy to check by the compiler. Perhaps, we could also hook up @ndmitchell's Weeder :)

snowleopard:
@izgzhen Thanks, merged!

PR #591, opened 2018-05-09
Fix unlit by placing it under lib/bin/ instead of bin/

alpmestan:
A nicer fix would involve patching GHC to not just look under
$libexec/ but also under the directory where the GHC binary itself lives
(bin/ for hadrian), so that we can leave all binaries under bin/.

Addresses [Trac #15132](https://ghc.haskell.org/trac/ghc/ticket/15132).

Without this patch, this used to fail:
``` sh
$ echo '> main = putStrLn "hello"' > foo.lhs 
$ _build/stage1/bin/ghc foo.lhs
```

with an error saying that ghc could not find `unlit` at `_build/stage1/lib/bin/unlit`, where it expects it. With this patch, GHC successfully compiles that file.

### DIFF
--- a/src/GHC.hs
+++ b/src/GHC.hs
@@ -131,8 +131,10 @@ programPath context@Context {..} = do
     -- The @touchy@ utility lives in the @lib/bin@ directory instead of @bin@,
     -- which is likely just a historical accident that will hopefully be fixed.
     -- See: https://github.com/snowleopard/hadrian/issues/570
-    path <- if package /= touchy then stageBinPath stage
-                                 else stageLibPath stage <&> (-/- "bin")
+    -- Likewise for 'unlit'.
+    path <- if package `elem` [touchy, unlit]
+      then stageLibPath stage <&> (-/- "bin")
+      else stageBinPath stage
     pgm  <- programName context
     return $ path -/- pgm <.> exe
 


src/GHC.hs
```diff
@@ -131,8 +131,10 @@ programPath context@Context {..} = do
     -- The @touchy@ utility lives in the @lib/bin@ directory instead of @bin@,
```
snowleopard: Could you update the comment as well?

snowleopard:
@alpmestan Thanks, happy to merge this, see a minor comment above. I've restarted the CI since Travis failed in a mysterious way.

alpmestan:
Two of the travis jobs still failed in an odd way (with an `!`, not a `X`). The others are green. Do you perhaps want to restart the two failing ones one last time?

src/GHC.hs
alpmestan: I added "Likewise for unlit" 2-3 lines below, as a comment to the same function, is that unsatisfactory / not clear enough?

snowleopard: Ah, indeed, I didn't notice.

snowleopard:
@alpmestan Merged, many thanks!

snowleopard:
@alpmestan I've noticed that both Make build system and Hadrian build `unlit` twice -- do you know if this is necessary? Can we build it just once, say in `Stage0`?

alpmestan:
@snowleopard It's a C program, so I don't think there is a lot to be gained by building `unlit` twice, indeed.

snowleopard:
@alpmestan Thanks, I've added an issue: #595.

Issue #594, opened 2018-05-11
nofib support


alpmestan:
Andreas (in [#15133](https://ghc.haskell.org/trac/ghc/ticket/15133)) and @tdammers (on the #ghc IRC channel) reported the problems they encountered while trying to run nofib on a GHC built by hadrian.

By default, nofib expects the usual inplace + ghc-stage<N> setup. We can point it to a separate GHC binary with commands like the ones I gave in #15133 (tl;dr: override `WithNofibHc` when running make commands).

Do we perhaps want to make this simpler and set those variables on behalf of hadrian users through dedicated `nofib` rule perhaps?

(Note: my "trick" does not allow me to complete the nofib run, see my first comment on the ticket mentionned above.)

snowleopard:
@alpmestan Thanks, I like the idea of adding the `nofib` rule to Hadrian. Will it simply call the `Make` builder in the `nofib` directory with appropriate arguments? This sounds like it should be easy to implement.

alpmestan:
@snowleopard Yes, that would/will in fact be very similar to how we run the tests.

snowleopard:
@alpmestan OK, I've labelled this as `easy` :) I'm unlikely to find time for this in the coming two weeks as I am travelling.

If anyone would like to pick this up, please add a comment.

alpmestan:
I took a quick stab at it, some further work is needed but it's a good start: https://github.com/snowleopard/hadrian/pull/599

PR #599, opened 2018-05-11
Preliminary nofib rule

alpmestan:
This doesn't yet capture stdout/stderr (nor does it `clean` or `boot`, even though I think & hope that the latter is done automatically), so we can't really use `nofib-analyse` yet (see [the nofib trac page](https://ghc.haskell.org/trac/ghc/wiki/Building/RunningNoFib)).

Still a WIP. Meant to address https://github.com/snowleopard/hadrian/issues/594.

@snowleopard Any guidance for the output capture bit? I would have thought the code in Nofib.hs would accomplish what I want (well, eventually I'll want stdout and stderr in the same file), but I can't see any `nofib.stdout`/`nofib.stderr` under my build root.

### DIFF
--- a/src/Rules/Nofib.hs
+++ b/src/Rules/Nofib.hs
@@ -1,0 +1,37 @@
+module Rules.Nofib where
+
+import Base
+import Expression
+import GHC.Packages
+import Oracles.Setting
+import Target
+import Utilities
+
+import System.Environment
+
+nofibRules :: Rules ()
+nofibRules = do
+  "nofib" ~> do
+    needNofibBuilders
+    makePath <- builderPath (Make "")
+    top <- topDirectory
+    root <- buildRoot
+    liftIO (setEnv "MAKE" makePath)
+
+    let out = top -/- root -/- "nofib.stdout"
+        err = top -/- root -/- "nofib.stderr"
+
+    buildWithCmdOptions [FileStdout out, FileStderr err] $
+      target (vanillaContext Stage2 compiler)
+             (Make "nofib")
+             []
+             [out, err]
+
+nofibHC :: FilePath -> String
+nofibHC ghcPath = "WithNofibHc=" ++ ghcPath
+
+perlProg :: FilePath -> String
+perlProg perlPath = "PERL=" ++ perlPath
+
+needNofibBuilders :: Action ()
+needNofibBuilders = needBuilder $ Ghc CompileHs Stage2
--- a/src/Settings/Builders/Make.hs
+++ b/src/Settings/Builders/Make.hs
@@ -1,5 +1,6 @@
 module Settings.Builders.Make (makeBuilderArgs) where
 
+import Builder
 import Rules.Gmp
 import Rules.Libffi
 import Settings.Builders.Common
@@ -9,8 +10,16 @@ makeBuilderArgs = do
     threads    <- shakeThreads <$> expr getShakeOptions
     gmpPath    <- expr gmpBuildPath
     libffiPath <- expr libffiBuildPath
+    ghcPath    <- expr $
+      (-/-) <$> topDirectory <*> builderPath (Ghc CompileHs Stage2)
+    perlPath   <- expr $ builderPath Perl
     let t = show $ max 4 (threads - 2) -- Don't use all Shake's threads
     mconcat
         [ builder (Make gmpPath          ) ? pure ["MAKEFLAGS=-j" ++ t]
         , builder (Make libffiPath       ) ? pure ["MAKEFLAGS=-j" ++ t, "install"]
-        , builder (Make "testsuite/tests") ? pure ["THREADS=" ++ t, "fast"] ]
+        , builder (Make "testsuite/tests") ? pure ["THREADS=" ++ t, "fast"]
+        , builder (Make "nofib"          ) ? pure
+            [ "WithNofibHc=" ++ ghcPath
+            , "PERL=" ++ perlPath
+            ]
+        ]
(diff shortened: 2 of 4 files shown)


snowleopard:
@alpmestan Thanks!

Perhaps you can capture `stdout` and `stderr` directly without intermediate files as in the examples below?

https://hackage.haskell.org/package/shake-0.16.4/docs/Development-Shake.html#v:command

src/Rules/Nofib.hs
```diff
@@ -0,0 +1,37 @@
...
+import Target
+import Utilities
+
+import System.Environment
+
+nofibRules :: Rules ()
+nofibRules = do
+  "nofib" ~> do
+    needNofibBuilders
+    makePath <- builderPath (Make "")
```
snowleopard: I guess `Make "nofib"` would be a bit more appropriate (but equivalent).

alpmestan:
> Perhaps you can capture stdout and stderr directly without intermediate files as in the examples below?

So you mean something like:

``` hs
(Stdout out, Stderr err) <- cmd ...
```

?

alpmestan:
Oh, there even is an [`Stdouterr`](https://hackage.haskell.org/package/shake-0.16.4/docs/Development-Shake.html#t:Stdouterr) newtype with a suitable `CmdResult` instance. This would mirror the first part of the following command:

``` sh
$ make 2>&1 | tee nofib-log
```

that is required to run `nofib` (see [here](https://ghc.haskell.org/trac/ghc/wiki/Building/RunningNoFib)).

I can then write the resulting bytestring to a file under `<build root>` myself and achieve what I want. The entire process for running nofib however is:

``` sh
$ cd nofib
$ make clean
$ make boot                  # Generates Makefile dependencies
$ make 2>&1 | tee nofib-log  # Compiles and runs benchmarks one by one
```

So I suppose I would be better off with a dedicated builder for `nofib` that runs all those commands, instead of having to special-case on `Make "nofib"`. Getting this whole set of commands to leave the source tree untouched and to put everything under the build root is not something I'm going to look into right now. :-)

snowleopard:
@alpmestan Sounds good :) Please ping me when you think this PR is ready.

alpmestan:
@snowleopard This PR in its current state lets me run nofib to completion with `./build.sh -j4 --build-flavour=_foo nofib` from a booted & configured source tree. nofib log attached available [here](https://gist.github.com/alpmestan/f27274ecb59930a2882c40e5bafae54b).

CI seems to be failing for an unrelated reason (and fails with the same error in the validation related PR). This patch is ready for review & possibly landing if you're satisfied.

src/Rules/Nofib.hs
```diff
@@ -0,0 +1,55 @@
...
+
+  -- a phony "nofib" rule that just triggers
+  -- the rule below.
+  "nofib" ~> need [root -/- nofibLogFile]
+
+  -- a rule to produce <build root>/nofig-log
+  -- by running the nofib suite and capturing
+  -- the relevant output.
+  root -/- nofibLogFile %> \fp -> do
+    needNofibDeps
```
snowleopard: I'm confused: why don't you simply execute all required commands here? What is the benefit of having a separate builder?

In the rest of Hadrian, we use builders primarily to have a unified way to track builder-specific command line arguments. But you are not using `Args` here, so it seems that defining a new builder is not required.

alpmestan: OK, I'll simplify this.

alpmestan: Done.

snowleopard:
@alpmestan The PR looks good to me now -- do you think it is ready to be merged?

alpmestan:
Yes, I'm happy with this patch as it is now. Merge when you want :-)

snowleopard:
@alpmestan Wait, it looks like this PR introduced recursive rules -- could you please have a look and fix?

snowleopard:
Once again, we are hit by now working CI -- we should give the highest priority to fixing it!

alpmestan:
@snowleopard I think I was seeing recursive rule errors on other PRs before I made this one, and I'm definitely not seeing anything like this locally. Could you give me commands + git commit for reproducing the issue?

snowleopard:
@alpmestan Well, both AppVeyor and Travis bots fail in the same way, and I think this is the first PR where this happened, but I may be mistaken, of course. I think it is sufficient to run `build selftest` to reproduce (but I won't have a chance to try on my machine until evening).

src/Settings/Builders/Make.hs
```diff
@@ -9,8 +10,16 @@ makeBuilderArgs = do
     threads    <- shakeThreads <$> expr getShakeOptions
     gmpPath    <- expr gmpBuildPath
     libffiPath <- expr libffiBuildPath
+    ghcPath    <- expr $
+      (-/-) <$> topDirectory <*> builderPath (Ghc CompileHs Stage2)
```
snowleopard: @alpmestan This could be the reason? You might want to move this into the `nofib` subexpression.

snowleopard:
@alpmestan To be more precise, you should try `build -c selftest` to reproduce.

src/Settings/Builders/Make.hs
snowleopard: Here is the error, which mentions `ghc-source-path`:
```
  Key value 3:  hadrian/cfg/system.config settings mk/config.h compiler/ghc.cabal
  Key value 4:  OracleQ (KeyValue ("hadrian/cfg/system.config","ghc-source-path"))
Rules may not be recursive
```

alpmestan: I don't even need this anymore. Let me try getting rid of the changes to this file altogether.

alpmestan:
I managed to reproduce with `build -c selftest`. The patch at #606 makes the problem go away for me.

PR #602, opened 2018-05-13
Added support for testsuite

chitrak7:
@snowleopard @angerman 
Validation result after this push: 
SUMMARY for test run started at Mon May 14 00:32:42 2018 IST
 0:20:37 spent to go through
    6359 total tests, which gave rise to
   23523 test cases, of which
   17231 were skipped

      42 had missing libraries
    6098 expected passes
      94 expected failures

       0 caused framework failures
       0 caused framework warnings
       1 unexpected passes
      55 unexpected failures
       2 unexpected stat failures

### DIFF
--- a/src/CommandLine.hs
+++ b/src/CommandLine.hs
@@ -44,6 +44,7 @@ defaultCommandLineArgs = CommandLineArgs
 data TestArgs = TestArgs
     { testOnly     :: Maybe String
     , testSkipPerf :: Bool
+    , testSpeed    :: TestSpeed
     , testSummary  :: Maybe FilePath
     , testJUnit    :: Maybe FilePath
     , testConfigs  :: [String] }
@@ -54,6 +55,7 @@ defaultTestArgs :: TestArgs
 defaultTestArgs = TestArgs
     { testOnly     = Nothing
     , testSkipPerf = False
+    , testSpeed    = Average
     , testSummary  = Nothing
     , testJUnit    = Nothing
     , testConfigs  = [] }
@@ -116,6 +118,18 @@ readTestOnly tests = Right $ \flags -> flags { testArgs = (testArgs flags) { tes
 readTestSkipPerf :: Either String (CommandLineArgs -> CommandLineArgs)
 readTestSkipPerf = Right $ \flags -> flags { testArgs = (testArgs flags) { testSkipPerf = True } }
 
+readTestSpeed :: Maybe String -> Either String (CommandLineArgs -> CommandLineArgs)
+readTestSpeed ms =
+    maybe (Left "Cannot parse test-speed") (Right . set) (go =<< lower <$> ms)
+  where
+    go :: String -> Maybe TestSpeed
+    go "fast"    = Just Fast
+    go "slow"    = Just Slow
+    go "average" = Just Average
+    go _         = Nothing
+    set :: TestSpeed -> CommandLineArgs -> CommandLineArgs
+    set flag flags = flags { testArgs = (testArgs flags) {testSpeed = flag} }
+
 readTestSummary :: Maybe String -> Either String (CommandLineArgs -> CommandLineArgs)
 readTestSummary filepath = Right $ \flags -> flags { testArgs = (testArgs flags) { testJUnit = filepath } }
 
@@ -128,7 +142,7 @@ readTestConfig config =
          Nothing -> Right id
          Just conf -> Right $ \flags ->
                         let configs = conf : testConfigs (testArgs flags)
-                         in flags { testArgs = (testArgs flags) { testConfigs = configs } }
+                        in flags { testArgs = (testArgs flags) { testConfigs = configs } }
 
 -- | Standard 'OptDescr' descriptions of Hadrian's command line arguments.
 optDescrs :: [OptDescr (Either String (CommandLineArgs -> CommandLineArgs))]
@@ -155,6 +169,8 @@ optDescrs =
       "Test cases to run."
     , Option [] ["skip-perf"] (NoArg readTestSkipPerf)
       "Skip performance tests."
+    , Option [] ["test-speed"] (OptArg readTestSpeed "SPEED")
+      "fast, slow or normal. Normal by default"
     , Option [] ["summary"] (OptArg readTestSummary "TEST_SUMMARY")
       "Where to output the test summary file."
     , Option [] ["summary-junit"] (OptArg readTestJUnit "TEST_SUMMARY_JUNIT")
--- a/src/CommandLine.hs
+++ b/src/CommandLine.hs
@@ -52,7 +52,6 @@ data TestArgs = TestArgs
     , testSkipPerf :: Bool
     , testSpeed    :: TestSpeed
     , testSummary  :: Maybe FilePath
-    , testThreads  :: Maybe String
     , testVerbosity:: Maybe String
     , testWays     :: [String] }
     deriving (Eq, Show)
@@ -67,7 +66,6 @@ defaultTestArgs = TestArgs
     , testSkipPerf = False
     , testSpeed    = Average
     , testSummary  = Nothing
-    , testThreads  = Nothing
     , testVerbosity= Nothing
     , testWays     = [] }
 
@@ -158,9 +156,6 @@ readTestSpeed ms =
 readTestSummary :: Maybe String -> Either String (CommandLineArgs -> CommandLineArgs)
 readTestSummary filepath = Right $ \flags -> flags { testArgs = (testArgs flags) { testJUnit = filepath } }
 
-readTestThreads :: Maybe String -> Either String (CommandLineArgs -> CommandLineArgs)
-readTestThreads thread = Right $ \flags -> flags { testArgs = (testArgs flags) { testThreads = thread } }
-
 readTestVerbose :: Maybe String -> Either String (CommandLineArgs -> CommandLineArgs)
 readTestVerbose verbose = Right $ \flags -> flags { testArgs = (testArgs flags) { testVerbosity = verbose } }
 
@@ -207,8 +202,6 @@ optDescrs =
       "fast, slow or normal. Normal by default"
     , Option [] ["summary"] (OptArg readTestSummary "TEST_SUMMARY")
       "Where to output the test summary file."
-    , Option [] ["test-threads"] (OptArg readTestThreads "TEST_THREADS")
-      "Number of concurrent parallel jobs"
     , Option [] ["test-verbose"] (OptArg readTestVerbose "TEST_VERBOSE")
       "A verbosity value between 0 and 5. 0 is silent, 4 and higher activates extra output."
     , Option [] ["test-way"] (OptArg readTestWay "TEST_WAY")
--- a/src/CommandLine.hs
+++ b/src/CommandLine.hs
@@ -42,23 +42,31 @@ defaultCommandLineArgs = CommandLineArgs
 
 -- | These arguments are used by the `test` target.
 data TestArgs = TestArgs
-    { testOnly     :: Maybe String
+    { testConfigs  :: [String]
+    , testJUnit    :: Maybe FilePath
+    , testOnly     :: Maybe String
+    , testOnlyPerf :: Bool
     , testSkipPerf :: Bool
     , testSpeed    :: TestSpeed
     , testSummary  :: Maybe FilePath
-    , testJUnit    :: Maybe FilePath
-    , testConfigs  :: [String] }
+    , testThreads  :: Maybe String
+    , testVerbosity:: Maybe String
+    , testWays     :: [String] }
     deriving (Eq, Show)
 
 -- | Default value for `TestArgs`.
 defaultTestArgs :: TestArgs
 defaultTestArgs = TestArgs
-    { testOnly     = Nothing
+    { testConfigs  = []
+    , testJUnit    = Nothing
+    , testOnly     = Nothing
+    , testOnlyPerf = False
     , testSkipPerf = False
     , testSpeed    = Average
     , testSummary  = Nothing
-    , testJUnit    = Nothing
-    , testConfigs  = [] }
+    , testThreads  = Nothing
+    , testVerbosity= Nothing
+    , testWays     = [] }
 
 readConfigure :: Either String (CommandLineArgs -> CommandLineArgs)
 readConfigure = Right $ \flags -> flags { configure = True }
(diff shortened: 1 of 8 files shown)


src/Rules/Test.hs
```diff
@@ -63,13 +64,24 @@ testRules = do
...
+  where
+    f :: Stage -> Package -> Action FilePath
+    f stage pkg | isLibrary pkg = pkgConfFile (Context stage pkg (read "v"))
+                | otherwise     = programPath =<< programContext stage pkg
+
 needTestBuilders :: Action ()
 needTestBuilders = do
     needBuilder $ Ghc CompileHs Stage2
     needBuilder $ GhcPkg Update Stage1
     needBuilder Hp2Ps
```
snowleopard: `Hp2Ps` is already included into `testsuitePackages`, so I guess we can drop this line?

src/Settings/Builders/Make.hs
```diff
@@ -1,16 +1,29 @@
...
-        , builder (Make "testsuite/tests") ? pure ["THREADS=" ++ t, "fast"] ]
+        , builder (Make "testsuite/tests") ? pure (["THREADS=" ++ t, "fast"] ++ testsuiteFlags) ]
+
+makeTestsuiteFlags :: Action [String]
+makeTestsuiteFlags = do
+    test_hc               <- f ghc
+    check_ppr             <- f checkPpr
+    check_api_annotations <- f checkApiAnnotations    
+    return ["TEST_HC=" ++ test_hc, "CHECK_PPR=" ++ check_ppr, "CHECK_API_ANNOTATIONS=" ++ check_api_annotations]
+  where
```
snowleopard: Let's rename `f` to something more informative, perhaps, `fullPath`.

snowleopard: Also, a type signature might help.

src/Rules/Test.hs
```diff
@@ -63,13 +64,24 @@ testRules = do
...
         buildWithCmdOptions env $ target (vanillaContext Stage2 compiler) RunTest [] []
 
+-- | Build extra programs required by testsuite
+needTestsuiteBuilders :: Action ()
+needTestsuiteBuilders = do
+    targets <- mapM (f Stage1) =<< testsuitePackages
+    need targets
+  where
+    f :: Stage -> Package -> Action FilePath
+    f stage pkg | isLibrary pkg = pkgConfFile (Context stage pkg (read "v"))
```
snowleopard: I think you need `vanillaContext` here.

snowleopard: Btw, are there testsuite-only libraries? If yes, are you sure `vanilla` is sufficient for the testsuite?

src/Rules/Test.hs
```diff
@@ -63,13 +64,24 @@ testRules = do
         -- Execute the test target.
         buildWithCmdOptions env $ target (vanillaContext Stage2 compiler) RunTest [] []
 
+-- | Build extra programs required by testsuite
+needTestsuiteBuilders :: Action ()
+needTestsuiteBuilders = do
+    targets <- mapM (f Stage1) =<< testsuitePackages
+    need targets
+  where
+    f :: Stage -> Package -> Action FilePath
```
snowleopard: Please choose a more informative name instead of `f`.

src/Rules/Test.hs
```diff
@@ -63,13 +64,24 @@ testRules = do
         -- Execute the test target.
         buildWithCmdOptions env $ target (vanillaContext Stage2 compiler) RunTest [] []
 
+-- | Build extra programs required by testsuite
```
snowleopard: The code below suggests that there are not only testsuite programs, but also testsuite libraries.

snowleopard:
@chitrak7 Thanks! I've added some comments.

src/GHC.hs
```diff
@@ -1,14 +1,14 @@
...
-    touchy, transformers, unlit, unix, win32, xhtml, ghcPackages, isGhcPackage,
-    defaultPackages, testsuitePackages,
+    array, base, binary, bytestring, cabal, checkApiAnnotations, checkPpr, 
+    compareSizes, compiler, containers, deepseq, deriveConstants, directory,
+    filepath, genapply, genprimopcode, ghc, ghcBoot, ghcBootTh, ghcCabal,
+    ghcCompact, ghci, ghcPkg, ghcPrim, ghcTags, ghcSplit, haddock, haskeline,
+    hsc2hs, hp2ps, hpc, hpcBin, integerGmp, integerSimple, iservBin, libffi, 
+    mtl, parsec, parallel, pretty, primitive, process, rts, runGhc, stm,
+    templateHaskell, terminfo, text, time, touchy, transformers, unlit, unix,
+    win32, xhtml, ghcPackages, isGhcPackage, defaultPackages, testsuitePackages,
```
izgzhen: nit: this diff looks unnecessarily big

snowleopard: @izgzhen At the moment this is unavoidable. We might refactor this to one package per line at some point if such large diffs look distracting, but it would be better to do it in a separate PR.

src/Rules/Test.hs
chitrak7: This bit of code is taken directly from Rules.hs.
There are many errors coming from rts library (40). Although I still couldn't figure out from where errors are coming, I have added this part just in case.

chitrak7:
@alpmestan I do not exactly know that, Can you give me pointers on where to look.  More importantly, why does testsuite decide to skip a test.

chitrak7:
@snowleopard Can you please see why recursive rule failure is coming?

alpmestan:
> @alpmestan I do not exactly know that, Can you give me pointers on where to look. More importantly, why does testsuite decide to skip a test.

You should see some information about this in the testsuite driver initialization part of the output (early on, when you look at the full log).

chitrak7:
@alpmestan I believe skipped tests are because of using "fast" option with testsuite.

alpmestan:
@chitrak7 OK. If you ever get a chance to, it would be terrific to add a way to run the equivalent of `./validate --slow`. If not, I'll do it. But not in this PR of course. Thanks for the reply and taking a look at what was going on :+1:

src/Rules/Test.hs
snowleopard: @chitrak7 So, do you really need to handle libraries here?

chitrak7: I have to handle rts library. I just cannot yet figure out how.

snowleopard: Why do you need to handle `rts` here? How is it related to the testsuite?

snowleopard:
@chitrak7 Please resolve merge conflicts.

snowleopard:
> Can you please see why recursive rule failure is coming?

@chitrak7 The recursive rule failure is unrelated to your PR. I think the reason is #599.

snowleopard:
@chitrak7 Something is wrong, looks like you've duplicated the recent commits instead of rebasing.

chitrak7:
@snowleopard I have reset my repository and made necessary changes. It is currently failing due to "llvm" issue. I will resend it once that PR Lands.

sighingnow:
Should the title of this PR be "Add support for nofib testsuite" ? We already support `hadrian test`.

alpmestan:
No, support for nofib was the goal of #599, not this PR. This PR is about supporting the execution of more tests, and about building things that are required by some of them which were not built previously. It still skips a whole bunch of tests but maybe @chitrak7 could provide an overview of the state of things with/without the commits from his branch? In terms of # of tests being run, # of broken test expectations.

sighingnow:
I just noticed that this PR also add a new file `src/Rules/Nofib.hs`...

alpmestan:
Right, my commits somehow did not get merged properly apparently. But you cann tell they've already been merged because they refer to the PR number.

chitrak7:
Hi @snowleopard @angerman, I have pushed the work done with validation implementation.

snowleopard:
@chitrak7 Thanks! Let's come back to this PR once we fix Hadrian.

src/Hadrian/Utilities.hs
```diff
@@ -481,3 +484,6 @@ renderUnicorn ls =
     ponyPadding = "                                            "
     boxLines :: [String]
     boxLines = ["", "", ""] ++ (lines . renderBox $ ls)
+
+data TestSpeed = Slow | Average | Fast deriving (Show, Eq)
```
snowleopard: I'm not sure `TestSpeed` is a generally useful data type. It's certainly useful for the GHC build system, but other build systems may have different settings or may not even need a testsuite (remember, `Hadrian.*` namespace is for generic functionality that may be useful not only for building GHC).

I suggest to move `TestSpeed` to the `CommandLine` module.

src/Rules/Test.hs
snowleopard: I still don't understand why `needTestsuiteBuilders` needs libraries. Can `testsuitePackages` contain library packages? Right now they only contain programs. If `testsuitePackages` can contain libraries, then you need to rename this function to something like `needTestsuitePackages` (instead of `Builders`).

src/Settings/Builders/Make.hs
```diff
@@ -13,5 +15,23 @@ makeBuilderArgs = do
...
         ]
+
+validateBuilderArgs :: Args
+validateBuilderArgs = builder (Make "testsuite/tests") ? do
+    threads             <- shakeThreads <$> expr getShakeOptions
+    top                 <- expr topDirectory
+    compiler            <- expr $ fullpath ghc
+    checkPpr            <- expr $ fullpath checkPpr
+    checkApiAnnotations <- expr $ fullpath checkApiAnnotations
+    let t = show $ max 4 (threads - 2)
```
snowleopard: Since the validation script will likely be the only running thread, we do not need to reduce parallelism like we do with other Make invocations. I suggest we use the `threads` settings as is.

src/CommandLine.hs
```diff
@@ -110,26 +120,55 @@ readProgressInfo ms =
...
-    case config of
-         Nothing -> Right id
-         Just conf -> Right $ \flags ->
-                        let configs = conf : testConfigs (testArgs flags)
-                         in flags { testArgs = (testArgs flags) { testConfigs = configs } }
+readTestVerbose :: Maybe String -> Either String (CommandLineArgs -> CommandLineArgs)
+readTestVerbose verbose = Right $ \flags -> flags { testArgs = (testArgs flags) { testVerbosity = verbose } }
 
+readTestWays :: Maybe String -> Either String (CommandLineArgs -> CommandLineArgs)
+readTestWays ways =
```
snowleopard: I don't understand, does this argument correspond to (multiple) ways or to a single way? Looking below, you expect to see a single way in `Just`. Shall this be renamed? If you worry about the clash of names, you can use `maybeWay` here.

src/CommandLine.hs
```diff
@@ -151,17 +190,27 @@ optDescrs =
...
-      "Output testsuite summary in JUnit format."
-    , Option [] ["config"] (OptArg readTestConfig "EXTRA_TEST_CONFIG")
-      "Configurations to run test, in key=value format." ]
-
+    , Option [] ["test-threads"] (OptArg readTestThreads "TEST_THREADS")
+      "Number of concurrent parallel jobs"
+    , Option [] ["test-verbose"] (OptArg readTestVerbose "TEST_VERBOSE")
+      "A verbosity value between 0 and 5. 0 is silent, 4 and higher activates extra output."
+    , Option [] ["test-way"] (OptArg readTestWays "TEST_WAY")
+      "only run these ways" ]
```
snowleopard: We need to document all these flags. You don't need to do this right in this PR, but at least please create an issue so that we don't forget about this.

src/CommandLine.hs
chitrak7: Every argument takes a single way. Multiple ways can be specified by this taking one way at a time.
eg:  --test-way=way1 --test-way=way2

src/Rules/Test.hs
chitrak7: Some files are missing in rts library causing multiple failures. At the moment, I cannot figure out which files are missing. Or they may be missing includes. I have to look into this matter further.

src/CommandLine.hs
snowleopard: Then this function should be renamed to `readTestWay`, since it reads a single way.

src/Rules/Test.hs
snowleopard: I don't understand how `rts` relates to this question. You don't have `rts` in `testsuitePackages`.

The comment for this function says "Build extra programs required by testsuite", but the implementation can also build libraries. This is an inconsistency that should be fixed.

src/CommandLine.hs
```diff
@@ -42,21 +42,31 @@ defaultCommandLineArgs = CommandLineArgs
...
+    , testJUnit    :: Maybe FilePath
+    , testOnly     :: Maybe String
+    , testOnlyPerf :: Bool
     , testSkipPerf :: Bool
+    , testSpeed    :: TestSpeed
     , testSummary  :: Maybe FilePath
-    , testJUnit    :: Maybe FilePath
-    , testConfigs  :: [String] }
+    , testThreads  :: Maybe String
+    , testVerbosity:: Maybe String
```
sighingnow: Hadrian (or Shake) already has `-jN` option, which is enough to control the speed of running testsuite. We don't need the `testThreads` here.

snowleopard: @sighingnow Indeed! Well spotted.

src/Rules/Test.hs
chitrak7: Resolved

src/Hadrian/Utilities.hs
```diff
@@ -30,7 +30,7 @@ module Hadrian.Utilities (
     (<&>), (%%>), cmdLineLengthLimit,
 
     -- * Useful re-exports
-    Dynamic, fromDynamic, toDyn, TypeRep, typeOf
+    Dynamic, fromDynamic, toDyn, TypeRep, typeOf,
```
snowleopard: Drop the comma

src/Settings/Builders/Make.hs
```diff
@@ -13,5 +15,22 @@ makeBuilderArgs = do
...
         ]
+
+validateBuilderArgs :: Args
+validateBuilderArgs = builder (Make "testsuite/tests") ? do
+    threads             <- shakeThreads <$> expr getShakeOptions
+    top                 <- expr topDirectory
+    compiler            <- expr $ fullpath ghc
+    checkPpr            <- expr $ fullpath checkPpr
+    checkApiAnnotations <- expr $ fullpath checkApiAnnotations
+    return [ "fast"
```
snowleopard: Is the `"fast"` above related to `TestSpeed.Fast`?

chitrak7: No, this is the default validate argument. Speed can be specified by running ./buils.dh test and not validate.

alpmestan: Worst case, we can later arrange to support `--slow` and `--fast` flags, like the actual `validate` script, can't we?

chitrak7: Yeah. I can do that right now.

snowleopard: OK, I guess we can leave this for later.

snowleopard:
OK, I think we can merge this -- there are a few minor issues but we can fix them later.

@alpmestan Do you have any further comments?

src/CommandLine.hs
```diff
@@ -42,21 +45,29 @@ defaultCommandLineArgs = CommandLineArgs
...
+    { testConfigs  :: [String]
+    , testJUnit    :: Maybe FilePath
+    , testOnly     :: Maybe String
+    , testOnlyPerf :: Bool
     , testSkipPerf :: Bool
+    , testSpeed    :: TestSpeed
     , testSummary  :: Maybe FilePath
-    , testJUnit    :: Maybe FilePath
-    , testConfigs  :: [String] }
+    , testVerbosity:: Maybe String
```
sighingnow: Shake already provides a `--verbose` option, let's reuse it.

chitrak7: The verbose option of Shake is boolean, whereas test suite supports 5 levels of verbosity. This is why I chose to pass a separate verbose argument.

snowleopard: I think that it may be convenient to control these two verbosity settings separately: you might want verbose testing, but not Shake diagnostic info. And vice versa. So, let's keep this flag.

snowleopard:
Merged. Many thanks @chitrak7, and everyone else for the review!

PR #606, opened 2018-05-18
remove (unnecessary) recursive dependency involving ghcPath

alpmestan:
This should fix the issues we've been talking about in #599. Let's see what CI says.

### DIFF
--- a/src/Settings/Builders/Make.hs
+++ b/src/Settings/Builders/Make.hs
@@ -1,6 +1,5 @@
 module Settings.Builders.Make (makeBuilderArgs) where
 
-import Builder
 import Rules.Gmp
 import Rules.Libffi
 import Settings.Builders.Common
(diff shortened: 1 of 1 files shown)


snowleopard:
@alpmestan Thanks! I love commits with such stats: `+0 -8` :-)

src/Settings/Builders/Make.hs
```diff
@@ -10,16 +9,9 @@ makeBuilderArgs = do
     threads    <- shakeThreads <$> expr getShakeOptions
     gmpPath    <- expr gmpBuildPath
     libffiPath <- expr libffiBuildPath
-    ghcPath    <- expr $
-      (-/-) <$> topDirectory <*> builderPath (Ghc CompileHs Stage2)
-    perlPath   <- expr $ builderPath Perl
     let t = show $ max 4 (threads - 2) -- Don't use all Shake's threads
```
snowleopard: @alpmestan By the way, do you think `nofib` could benefit from using parallelism? Perhaps, we could pass `t` to its Make invocations in the examples below?

alpmestan: @snowleopard I thought about it but I don't think we want to do that. First, [the nofib trac page](https://ghc.haskell.org/trac/ghc/wiki/Building/RunningNoFib) does not mention any parallelism variable/knob. Second, nofib is after all a benchmark suite, so we quite likely don't want to run many of them simultaneously as this could have quite an impact on their respective performance and therefore would make the measurements "useless".

snowleopard: @alpmestan Thanks, I see.

snowleopard:
Merged!

Issue #611, opened 2018-06-06
Cannot build RTS on Windows


snowleopard:
After #610, the Windows build fails with the following error:
```
[...]

PR #613, opened 2018-06-12
Update RTS flags

snowleopard:
See #611


alpmestan:
So CI is green on Linux and Windows? Progress! :-)

snowleopard:
@alpmestan Yes! :-) Hopefully, someone with OS X can figure out a fix for it.

Issue #615, opened 2018-06-12
Nix: /nix/store/.../bin/ld: cannot find -lgmp


sgillespie:
There appears to be a missing dependency on `gmp` in `shell.nix`:

```
/nix/store/j75dgadrff2d1fyc4fczmcgqkid2imdx-binutils-2.30/bin/ld: cannot find -lgmp
collect2: error: ld returned 1 exit status
`cc' failed in phase `Linker'. (Exit code: 1)
shakeArgsWith     0.000s    0%
Function shake    0.484s    0%
Database read     0.000s    0%
With database     0.002s    0%
Running rules  2199.161s   99%  =========================
Total          2199.647s  100%
Error when running Shake build system:
* _build/stage1/lib/bin/unlit
user error (Development.Shake.cmd, system command failed
Command: _build/stage0/bin/ghc -Wall -hisuf hi -osuf o -hcsuf hc -static -hide-all-packages -no-user-package-db '-package-db _build/stage1/lib/package.conf.d' -i -i_build/stage1/utils/unlit/build -i_build/stage1/utils/unlit/build/unlit/autogen -iutils/unlit/. -Iincludes -I_build/generated -I_build/stage1/utils/unlit/build -I/nix/store/70g2i2v6dqm5ss9p2m1jb8f2swsz46qh-gmp-6.1.2-dev/include -I_build/generated -optc-I_build/generated -optP-include -optP_build/stage1/utils/unlit/build/unlit/autogen/cabal_macros.h -optc-fno-stack-protector -odir _build/stage1/utils/unlit/build -hidir _build/stage1/utils/unlit/build -stubdir _build/stage1/utils/unlit/build -no-auto-link-packages -no-hs-main -optl-lgmp -Wnoncanonical-monad-instances -optc-Werror=unused-but-set-variable -optc-Wno-error=inline _build/stage1/utils/unlit/build/c/unlit.o _build/stage1/utils/unlit/build/c/fs.o -o _build/stage1/lib/bin/unlit -O2 -H32m -XHaskell2010 -ghcversion-file=/home/sgillespie/dev/ghc-hadrian/_build/generated/ghcversion.h
Exit code: 1
Stderr:
/nix/store/j75dgadrff2d1fyc4fczmcgqkid2imdx-binutils-2.30/bin/ld: cannot find -lgmp
collect2: error: ld returned 1 exit status
`cc' failed in phase `Linker'. (Exit code: 1)
)
```

snowleopard:
[...]

PR #616, opened 2018-06-13
Add `gmp` and `ncurses` as nix dependencies (#615)

sgillespie:
Lifted from: https://github.com/alpmestan/ghc.nix/blob/master/default.nix

I'm not sure why I had to specify `.dev` and `.out`, but this is what worked for me

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -76,6 +76,8 @@ in
         [ hadrianPackages.hadrian
           nixpkgs.arcanist
           nixpkgs.git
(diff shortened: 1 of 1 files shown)


snowleopard:
@sgillespie Many thanks, looks good to me.

@alpmestan @mpickering Could one of you please have a look as well?

snowleopard:
Merged. Thanks @sgillespie and @alpmestan!

PR #621, opened 2018-06-13
Add test compiler option to test

chitrak7:
I have applied an approach similar to the one used by boilerplate.mk to find other packages. 
This is not yet complete. I just wish to review the steps taken here.

### DIFF
--- a/src/CommandLine.hs
+++ b/src/CommandLine.hs
@@ -53,9 +53,12 @@ data TestArgs = TestArgs
     , testSpeed    :: TestSpeed
     , testSummary  :: Maybe FilePath
 <<<<<<< HEAD
+<<<<<<< HEAD
 =======
     , testThreads  :: Maybe String
 >>>>>>> Added support for more testing features
+=======
+>>>>>>> Removed TestThread argument
     , testVerbosity:: Maybe String
     , testWays     :: [String] }
     deriving (Eq, Show)
@@ -71,9 +74,12 @@ defaultTestArgs = TestArgs
     , testSpeed    = Average
     , testSummary  = Nothing
 <<<<<<< HEAD
+<<<<<<< HEAD
 =======
     , testThreads  = Nothing
 >>>>>>> Added support for more testing features
+=======
+>>>>>>> Removed TestThread argument
     , testVerbosity= Nothing
     , testWays     = [] }
 
@@ -210,21 +216,9 @@ optDescrs =
       "fast, slow or normal. Normal by default"
     , Option [] ["summary"] (OptArg readTestSummary "TEST_SUMMARY")
       "Where to output the test summary file."
-<<<<<<< HEAD
     , Option [] ["test-verbose"] (OptArg readTestVerbose "TEST_VERBOSE")
       "A verbosity value between 0 and 5. 0 is silent, 4 and higher activates extra output."
     , Option [] ["test-way"] (OptArg readTestWay "TEST_WAY")
-=======
-    , Option [] ["test-threads"] (OptArg readTestThreads "TEST_THREADS")
-      "Number of concurrent parallel jobs"
-    , Option [] ["test-verbose"] (OptArg readTestVerbose "TEST_VERBOSE")
-      "A verbosity value between 0 and 5. 0 is silent, 4 and higher activates extra output."
-<<<<<<< HEAD
-    , Option [] ["test-way"] (OptArg readTestWays "TEST_WAY")
->>>>>>> Added support for more testing features
-=======
-    , Option [] ["test-way"] (OptArg readTestWay "TEST_WAY")
->>>>>>> Minor Revision
       "only run these ways" ]
     
 -- | A type-indexed map containing Hadrian command line arguments to be passed
--- a/src/CommandLine.hs
+++ b/src/CommandLine.hs
@@ -53,13 +53,6 @@ data TestArgs = TestArgs
     , testSkipPerf :: Bool
     , testSpeed    :: TestSpeed
     , testSummary  :: Maybe FilePath
-<<<<<<< HEAD
-<<<<<<< HEAD
-=======
-    , testThreads  :: Maybe String
->>>>>>> Added support for more testing features
-=======
->>>>>>> Removed TestThread argument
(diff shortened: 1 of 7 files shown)


src/Settings/Builders/RunTest.hs
```diff
@@ -81,53 +72,76 @@ runTestBuilderArgs = builder RunTest ? do
...
 setTestSpeed Fast    = "2"
 setTestSpeed Average = "1"
 setTestSpeed Slow    = "0"
 
+-- | Returns parent path of test compiler 
+parentPath :: String -> String
+parentPath path = let upPath = init $ splitOn "/" path
+                  in  intercalate "/" upPath
+
+fullpath :: Stage -> Package -> Action FilePath
```
chitrak7: fullpath has been used a lot of times. Maybe we should move it to Hadrian.Utilities?

src/CommandLine.hs
```diff
@@ -121,6 +123,11 @@ readProgressInfo ms =
 readSplitObjects :: Either String (CommandLineArgs -> CommandLineArgs)
 readSplitObjects = Right $ \flags -> flags { splitObjects = True }
 
+readTestCompiler :: Maybe String -> Either String (CommandLineArgs -> CommandLineArgs)
+readTestCompiler compiler = maybe (Left "Cannot parse compiler") (Right . set) (lower <$> compiler)
```
chitrak7: TODO: remove lower

src/Settings/Builders/RunTest.hs
```diff
@@ -81,53 +72,76 @@ runTestBuilderArgs = builder RunTest ? do
...
-                         Nothing -> Nothing
-                         Just verbosity -> Just $ "--verbose=" ++ verbosity
-        wayArgs    = map ("--way=" ++) (testWays args) 
-    pure $  testOnlyArg
-         ++ speedArg 
+                           Nothing -> Just $ "--verbose=" ++ show (fromEnum globalVerbosity)
+                           Just verbosity -> Just $ "--verbose=" ++ verbosity
+        wayArgs      = map ("--way=" ++) (testWays args) 
+        compilerArg  = ["--config", "compiler=" ++ show (compiler)]
+        ghcPkgArg    = ["--config", "ghc-pkg=" ++ show (bindir -/- "ghc-pkg")]
```
chitrak7: typo: ghc_pkg

snowleopard:
Thanks @chitrak7! I'll have a look a bit later. By the way, it would be much better if instead of adding comments here, you actually add comments to the code, explaining what each function does. Comments made here will be lost and forgotten, but comments in the code will live!

chitrak7:
Hi @snowleopard I believe this PR is complete. Stage1 gives multiple failures but that is because many bootstrap libraries are used and testsuite requires the same version of ghc and library. I have tested stage0, stage1, stage2, bindist compiler and a binary downloaded.

src/Settings/Builders/RunTest.hs
```diff
@@ -81,53 +72,76 @@ runTestBuilderArgs = builder RunTest ? do
...
+setCompiler compiler = pure compiler 
 
 -- | Set speed for test
 setTestSpeed :: TestSpeed -> String
 setTestSpeed Fast    = "2"
 setTestSpeed Average = "1"
 setTestSpeed Slow    = "0"
 
+-- | Returns parent path of test compiler 
+parentPath :: String -> String
```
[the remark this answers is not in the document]
snowleopard: Such questions should also go in the source code in the form of `TODO`s. For example:
```haskell
-- TODO: The current implementation is not perfect, because X, Y and Z. Can we find a better one?
```

snowleopard:
@chitrak7 Thanks, I've asked @alpmestan to review this, since he is more familiar with the requirements for this part of Hadrian.

alpmestan:
FWIW, this is what I get with this branch:

```
SUMMARY for test run started at Thu Jun 14 13:24:13 2018 CEST
 0:12:38 spent to go through
    6416 total tests, which gave rise to
   15870 test cases, of which
    9186 were skipped

      36 had missing libraries
    5627 expected passes
     114 expected failures

      63 caused framework failures
       0 caused framework warnings
      54 unexpected passes
```

I'm working on some of the problematic tests already.

src/CommandLine.hs
```diff
@@ -188,6 +195,8 @@ optDescrs =
       "Progress info style (None, Brief, Normal or Unicorn)."
     , Option [] ["split-objects"] (NoArg readSplitObjects)
       "Generate split objects (requires a full clean rebuild)."
+    , Option [] ["test-compiler"] (OptArg readTestCompiler "TEST_COMPILER")
+      "Use given compiler [Default=Stage2]."
```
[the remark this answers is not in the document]
chitrak7: Yeah, a typographical error. I will update it in a couple hours after I finish with bindist rule.

chitrak7:
@snowleopard I will be sending a slightly revised version accounting for typographical mistakes and better comments.

snowleopard:
Thanks @alpmestan!

@chitrak7 Sure, please ping me when you are done.

chitrak7:
@snowleopard I have made the required changes

src/Settings/Builders/RunTest.hs
```diff
@@ -81,53 +72,85 @@ runTestBuilderArgs = builder RunTest ? do
...
+setCompiler compiler = pure compiler 
 
 -- | Set speed for test
 setTestSpeed :: TestSpeed -> String
 setTestSpeed Fast    = "2"
 setTestSpeed Average = "1"
 setTestSpeed Slow    = "0"
 
+-- | Returns parent path of test compiler 
+-- | TODO : Can this be implemented in a better way
```
snowleopard: Could you be more specific here? Describe why the current approach is not good enough, so that it is clear to the person comes across this `TODO` what the end end goal is.

chitrak7: Actually, I just find it a bit hacky. I cannot think of any reason why this will fail. Just the approach is not satisfying in my opinion.

snowleopard:
OK, looks good to me -- merged.

@chitrak7 Thanks again!

PR #623, opened 2018-06-14
Added featues to binary distribution

chitrak7:
The code is working on linux now.
@angeman @alpmestan @snowleopard

### DIFF
--- a/src/Rules/BinaryDist.hs
+++ b/src/Rules/BinaryDist.hs
@@ -66,7 +66,7 @@ bindistRules = do
       writeFile' makefilePath bindistMakefile
     
     root -/- "bindist" -/- "ghc-*" -/- "wrappers/*" %> \wrapperPath -> 
-      writeFile' wrapperPath $ wrapper (basename wrapperPath)
+      writeFile' wrapperPath $ wrapper (takeFileName wrapperPath)
 
     -- copy over the various configure-related files needed for a working
     -- './configure [...] && make install' workflow
@@ -213,9 +213,6 @@ bindistMakefile = unlines
   , "# END INSTALL"
   , "# -----------------------------------------------------------------------------"
   ]
-
-basename :: FilePath -> String
-basename path = last $ splitOn "/" path
 
 wrapper :: String -> String
 wrapper "ghc"        = ghcWrapper
--- a/src/Rules/BinaryDist.hs
+++ b/src/Rules/BinaryDist.hs
@@ -37,6 +37,9 @@ bindistRules = do
       -- other machine.
       need $ map (bindistFilesDir -/-)
                  (["configure", "Makefile"] ++ bindistInstallFiles)
+      need $ map ((bindistFilesDir -/- "wrappers") -/-) ["check-api-annotations"
+                 , "check-ppr", "ghc", "ghc-iserv", "ghc-pkg", "ghc-split"
+                 , "ghci", "haddock", "hpc", "hp2ps", "hsc2hs", "runhaskell"]
 
       -- finally, we create the archive, at
       -- <root>/bindist/ghc-X.Y.Z-platform.tar.xz
@@ -61,6 +64,9 @@ bindistRules = do
     -- generate the Makefile that enables the "make install" part
     root -/- "bindist" -/- "ghc-*" -/- "Makefile" %> \makefilePath ->
       writeFile' makefilePath bindistMakefile
+    
+    root -/- "bindist" -/- "ghc-*" -/- "wrappers/*" %> \wrapperPath -> 
+      writeFile' wrapperPath $ wrapper (basename wrapperPath)
 
     -- copy over the various configure-related files needed for a working
     -- './configure [...] && make install' workflow
@@ -83,7 +89,7 @@ bindistInstallFiles :: [FilePath]
 bindistInstallFiles =
   [ "config.sub", "config.guess", "install-sh"
   , "mk" -/- "config.mk.in", "mk" -/- "install.mk.in"
-  , "settings.in", "README", "INSTALL"
+  , "mk" -/- "project.mk.in", "settings.in", "README", "INSTALL"
   ]
 
 -- | Auxiliary function that gives us a 'Filepath' we can 'need' for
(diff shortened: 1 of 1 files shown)


src/Rules/BinaryDist.hs
```diff
@@ -61,6 +63,27 @@ bindistRules = do
...
+      writeFile' wrapperHaddockPath haddockWrapper
+
+    root -/- "bindist" -/- "ghc-*" -/- "wrappers/common" %> \wrapperCommonPath ->
+      writeFile' wrapperCommonPath commonWrapper
+
+    root -/- "bindist" -/- "ghc-*" -/- "wrappers/hsc2hs" %> \wrapperHsc2hsPath ->
+      writeFile' wrapperHsc2hsPath hsc2hsWrapper
+
+    root -/- "bindist" -/- "ghc-*" -/- "wrappers/runhaskell" %> \wrapperRunHaskellPath ->
+      writeFile' wrapperRunHaskellPath runhaskellWrapper
```
alpmestan: I think you could substantially shorten this bit of code by having an function to abstract over the general pattern:

``` haskell
  bindistWrapperRule root "ghc" wrapperGhcPath ghcWrapper
  [...]
  bindistWrapperRule root "hsc2hs" wrapperHsc2hsPath hsc2hsWrapper
  bindistWrapperRule root "runhaskell" wrapperRunHaskellPath runhaskellWrapper
```

src/Rules/BinaryDist.hs
```diff
@@ -83,7 +106,7 @@ bindistInstallFiles :: [FilePath]
 bindistInstallFiles =
   [ "config.sub", "config.guess", "install-sh"
   , "mk" -/- "config.mk.in", "mk" -/- "install.mk.in"
-  , "settings.in", "README", "INSTALL"
+  , "mk" -/- "project.mk.in", "settings.in", "README", "INSTALL"
```
alpmestan: I don't see any occurence of `project.mk` or `project.mk.in` besides this one, does the code as it is now somehow implicitly requires it? Or something will, in an upcoming patch? Or is it just a leftover from an experiment of yours?

chitrak7: project.mk is a dependency of config.mk. I am using config.mk.

chitrak7:
I have made the necessary changes @alpmestan @angerman @snowleopard an tested it on linux. 
Also, we are not shipping ghci, should I create a separate PR to do so?

chitrak7:
@alpmestan Should we ship c header files as well. It is probably required by hsc2hs. If yes, then I can create a separate PR.

alpmestan:
We do need to ship some C header files yes. This is what my bindist of 8.4.3 comes with:

``` sh
$ ls /nix/store/a5z79ybg660ymjdq8mh4mcr7p6cbpxg4-ghc-8.4.3-with-packages/lib/ghc-8.4.3/include/
Cmm.h               HsFFI.h     Rts.h     Stg.h  ffitarget.h    ghcconfig.h    ghcversion.h  stg/
DerivedConstants.h  MachDeps.h  RtsAPI.h  ffi.h  ghcautoconf.h  ghcplatform.h  rts/
```

chitrak7:
Hi @alpmestan, I have figured out how to ship includes. But since this PR is already pretty big, I think I should send a separate PR once this is merged.

src/Rules/BinaryDist.hs
alpmestan: OK, as long as it doesn't pull anything else with it, that should be fine :-)

alpmestan:
@chitrak7 As you prefer. Most of your PR (in terms of # of lines changed) is about your Makefile changes, so it's still quite easily reviewable, so you can feel free to use this PR too unless @snowleopard objects.

src/Rules/BinaryDist.hs
```diff
@@ -106,16 +112,174 @@ bindistMakefile = unlines
...
+  , "\t$(INSTALL_DIR) \"$(bindir)\""
+  , "\t$(foreach p, $(BINARY_NAMES),\\"
+  , "\t\t$(call installscript,$p,$(bindir)/$p,$(bindir),$(GHCBINDIR),$(GHCBINDIR)/$p,$(libdir),$(docdir),$(includedir)))"
+  , ""
+  , "# END INSTALL"
+  , "# -----------------------------------------------------------------------------"
   ]
+
+basename :: FilePath -> String
+basename path = last $ splitOn "/" path
```
snowleopard: Should this be replaced with `takeFileName`?

src/Rules/BinaryDist.hs
```diff
@@ -106,16 +112,174 @@ bindistMakefile = unlines
...
+
+wrapper :: String -> String
```
snowleopard: Should the first argument be given type `FilePath` instead? It's not just any `String`, it's a file name.

src/Rules/BinaryDist.hs
snowleopard: I mean the standard function:

https://hackage.haskell.org/package/filepath-1.4.2/docs/System-FilePath-Posix.html#v:takeFileName

src/Rules/BinaryDist.hs
chitrak7: @snowleopard It is not exactly a filename. It is the name of the program we are installing program for.

src/Rules/BinaryDist.hs
chitrak7: Resolved. Thanks @snowleopard

src/Rules/BinaryDist.hs
snowleopard: Yes, but `takeFileName` returns a `FilePath`, not a `String`. I know it's just a synonym, but still it is worth following standard convention, where file/program names are given type `FilePath`.

src/Rules/BinaryDist.hs
chitrak7: Yes, project.mk is independent.

src/Rules/BinaryDist.hs
chitrak7: @snowleopard Resolved

snowleopard:
@chitrak7 Thanks, looks good to me.

@alpmestan Any further comments from you?

alpmestan:
No, looks good to me.

snowleopard:
Thanks, merged!

PR #624, opened 2018-06-15
Added more Testing features

chitrak7:
Added config file field in test arguments.
Added option to specify speed with validate.
@snowleopard @angerman @alpmestan


src/CommandLine.hs
```diff
@@ -45,31 +45,33 @@ defaultCommandLineArgs = CommandLineArgs
...
-    , testJUnit    = Nothing
-    , testOnly     = Nothing
-    , testOnlyPerf = False
-    , testSkipPerf = False
-    , testSpeed    = Average
-    , testSummary  = Nothing
-    , testVerbosity= Nothing
-    , testWays     = [] }
+    { testCompiler   = "stage2"
+    , testConfigFile = Nothing
```
alpmestan: Shuoldn't this be `Just "testsuite/config/ghc"` ?

src/Settings/Builders/RunTest.hs
```diff
@@ -85,7 +84,10 @@ getTestArgs = do
     bindir          <- expr $ setBinaryDirectory (testCompiler args)
     compiler        <- expr $ setCompiler (testCompiler args)
     globalVerbosity <- shakeVerbosity <$> expr getShakeOptions 
-    let testOnlyArg  = case testOnly args of
+    let configFileArg= case testConfigFile args of
+                           Just filepath -> ["--config-file=" ++ filepath]
+                           Nothing       -> ["--config-file=testsuite/config/ghc"]
```
alpmestan: I'd recommend just using testsuite/config/ghc as the default value, as pointed out in my other comment. If an arg is given, use it, otherwise use the default value.

chitrak7: Resolved

snowleopard:
@chitrak7 @alpmestan Thanks for the PR and the review!

PR #625, opened 2018-06-15
attempt at setting some more test arguments correctly

alpmestan:
This patch adds 2 more packages (both required for at least some tests) and tries to set some more test options correctly, using the information we have from the `Flavour`. Feedback welcome! I'm not sure how reliable that is.

### DIFF
--- a/src/GHC.hs
+++ b/src/GHC.hs
@@ -107,6 +107,8 @@ stage2Packages = return [haddock]
 testsuitePackages :: Action [Package]
 testsuitePackages = return [ checkApiAnnotations
                            , checkPpr
+                           , ghcPkg
+                           , parallel
                            , hp2ps              ]
 
 -- | Given a 'Context', compute the name of the program that is built in it
--- a/src/Rules/Test.hs
+++ b/src/Rules/Test.hs
@@ -71,7 +71,12 @@ needTestsuiteBuilders = do
     need targets
   where
     needfile :: Stage -> Package -> Action FilePath
-    needfile stage pkg = programPath =<< programContext stage pkg
+    needfile stage pkg
+      -- TODO (Alp): we might sometimes need more than vanilla!
+      -- This should therefore depend on what test ways
+      -- we are going to use, I suppose?
+      | isLibrary pkg = pkgConfFile (Context stage pkg (read "v"))
+      | otherwise = programPath =<< programContext stage pkg
 
 needTestBuilders :: Action ()
 needTestBuilders = do
--- a/src/Rules/Test.hs
+++ b/src/Rules/Test.hs
@@ -75,7 +75,7 @@ needTestsuiteBuilders = do
       -- TODO (Alp): we might sometimes need more than vanilla!
       -- This should therefore depend on what test ways
       -- we are going to use, I suppose?
-      | isLibrary pkg = pkgConfFile (Context stage pkg (read "v"))
+      | isLibrary pkg = pkgConfFile (vanillaContext stage pkg)
       | otherwise = programPath =<< programContext stage pkg
 
 needTestBuilders :: Action ()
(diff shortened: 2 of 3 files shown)


src/GHC.hs
```diff
@@ -107,6 +107,8 @@ stage2Packages = return [haddock]
 testsuitePackages :: Action [Package]
 testsuitePackages = return [ checkApiAnnotations
                            , checkPpr
+                           , ghcPkg
```
chitrak7: We have ghcPkg in stage1. Do we need it here also?

alpmestan: Well, it was not getting built here with `build.sh [...] test`. After adding it there, it was.

src/Rules/Test.hs
```diff
@@ -71,7 +71,12 @@ needTestsuiteBuilders = do
     need targets
   where
     needfile :: Stage -> Package -> Action FilePath
-    needfile stage pkg = programPath =<< programContext stage pkg
+    needfile stage pkg
+      -- TODO (Alp): we might sometimes need more than vanilla!
+      -- This should therefore depend on what test ways
+      -- we are going to use, I suppose?
```
chitrak7: I am currently working on this. Think a PR will be ready in shortly.

alpmestan: Good. Thanks!

src/GHC.hs
chitrak7: In this case, maybe there are other stage1 packages that are not built properly. Should we change the need rule implemented there to need all stage1 packages?

alpmestan:
This is what I get with this patch:

```
SUMMARY for test run started at Fri Jun 15 11:50:07 2018 CEST
 0:15:43 spent to go through
    6416 total tests, which gave rise to
   23725 test cases, of which
   16974 were skipped

      40 had missing libraries
    5714 expected passes
      92 expected failures

      67 caused framework failures
       0 caused framework warnings
       1 unexpected passes
     896 unexpected failures
       1 unexpected stat failures
```

Back when I tried #621, the results that I got were:

```
SUMMARY for test run started at Thu Jun 14 13:24:13 2018 CEST
 0:12:38 spent to go through
    6416 total tests, which gave rise to
   15870 test cases, of which
    9186 were skipped

      36 had missing libraries
    5627 expected passes
     114 expected failures

      63 caused framework failures
       0 caused framework warnings
      54 unexpected passes
```

So we're skipping **a lot less** tests now, and I could have increased that number I suppose if we were building `dyn` etc variants of things by default here. And shouldn't we be? The default flavour (`perf`) surely builds a whole bunch of things.

src/Rules/Test.hs
```diff
@@ -71,7 +71,12 @@ needTestsuiteBuilders = do
...
+      | isLibrary pkg = pkgConfFile (Context stage pkg (read "v"))
```
snowleopard: Shall we change `Context stage pkg (read "v")` to `vanillaContext stage pkg`?

src/GHC.hs
snowleopard: I think `testsuitePackages` should list all packages that are requried for the testsuite. It doesn't matter what we build in Stage0 or Stage1.

snowleopard:
@alpmestan Thanks! Looks good, so happy to merge. Just a minor comment above.

src/Rules/Test.hs
alpmestan: Oh, yes, right.

src/GHC.hs
alpmestan: Yeah I think it's fine to put those in `testsuitePackages`, as at least for the two ones that I added, they actually are (required and) used by some tests.

alpmestan:
>  Just a minor comment above

Addressed :-)

snowleopard:
Thanks, merged!

PR #627, opened 2018-06-17
For executables, we should read the `main-is` field from the cabal file.

sighingnow:
Previously, we simply treat file name for `Main` module as `Main.hs` to build executables. That doesn't work for the `timeout` program. This patch fixes the problem.

### DIFF
--- a/src/Oracles/ModuleFiles.hs
+++ b/src/Oracles/ModuleFiles.hs
@@ -124,10 +124,15 @@ moduleFilesOracle = void $ do
     void . addOracle $ \(ModuleFiles (stage, package)) -> do
         let context = vanillaContext stage package
         srcDirs <- interpretInContext context (getPackageData PD.srcDirs)
+        mainIs  <- interpretInContext context (getPackageData PD.mainIs)
+        let removeMain = case mainIs of
+                              Just (mod, _) -> delete mod
+                              Nothing       -> id
         modules <- fmap sort $ interpretInContext context (getPackageData PD.modules)
         autogen <- autogenPath context
         let dirs = autogen : map (pkgPath package -/-) srcDirs
-            modDirFiles = groupSort $ map decodeModule modules
+            -- Don't resolve the file path for module `Main` twice.
+            modDirFiles = groupSort $ map decodeModule $ removeMain modules
         result <- concatForM dirs $ \dir -> do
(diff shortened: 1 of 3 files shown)


src/Oracles/ModuleFiles.hs
```diff
@@ -136,7 +141,16 @@ moduleFilesOracle = void $ do
                 let cmp f = compare (dropExtension f)
                     found = intersectOrd cmp files mFiles
                 return (map (fullDir -/-) found, mDir)
-        let pairs = sort [ (encodeModule d f, f) | (fs, d) <- result, f <- fs ]
+
+        mainpairs <- case mainIs of
+            Just (mod, filepath) ->
+                concatForM dirs $ \dir -> do
+                    found <- doesFileExist (dir -/- filepath)
+                    return $ if found then [(mod, unifyPath $ dir -/- filepath)]
```
snowleopard: `[ (mod, unifyPath $ dir -/- filepath) | found ]` might be less awkward.

src/Oracles/ModuleFiles.hs
```diff
@@ -136,7 +141,16 @@ moduleFilesOracle = void $ do
                 let cmp f = compare (dropExtension f)
                     found = intersectOrd cmp files mFiles
                 return (map (fullDir -/-) found, mDir)
-        let pairs = sort [ (encodeModule d f, f) | (fs, d) <- result, f <- fs ]
+
+        mainpairs <- case mainIs of
```
snowleopard: Could you add a comment here, explaining why `main`s are treated differently?

snowleopard:
@sighingnow Many thanks! Just a couple of minor comments above. Please add a comment to clarify what's going on.

src/Oracles/ModuleFiles.hs
sighingnow: Fixed.

src/Oracles/ModuleFiles.hs
sighingnow: Done.

sighingnow:
@snowleopard Could we land this now ?

snowleopard:
@sighingnow Yes, thanks! I'm just waiting for at least one green light from Travis (I believe 8.2.2 was fine).

snowleopard:
Merged.

sighingnow:
Thanks !

Issue #559, opened 2018-04-07
Using NUMA requires -lnuma to be passed to GHC


izgzhen:
```
collect2: error: ld returned 1 exit status
`gcc' failed in phase `Linker'. (Exit code: 1)
shakeArgsWith    0.000s    0%                           
Function shake  12.998s    4%  =                        
Database read    0.251s    0%                           
With database    0.012s    0%                           
Running rules  267.347s   95%  =========================
Total          280.609s  100%                           
Error when running Shake build system:
* _build/stage1/bin/runhaskell
user error (Development.Shake.cmd, system command failed
[...]

PR #628, opened 2018-06-17
Remove validation errors

### LINKED ISSUE #626: Build Profiling and Dynamic Ways for libraries
We need Profiling and Dynamic Ways to pass validation. These ways are required for ghci, base and rts.
@snowleopard  @alpmestan

chitrak7:
I have copied some files as required by testsuite.

Also, I have unsuccessfully attempted to resolve #626.

@sighingnow @snowleopard @alpmestan

### DIFF
--- a/src/GHC.hs
+++ b/src/GHC.hs
@@ -110,8 +110,10 @@ testsuitePackages = return [ checkApiAnnotations
                            , ghci
                            , ghcPkg
                            , hp2ps
+                           , iserv
                            , parallel
-                           , timeout         ]
+                           , runGhc
+                           , timeout           ]
 
 -- | Given a 'Context', compute the name of the program that is built in it
 -- assuming that the corresponding package's type is 'Program'. For example, GHC
--- a/src/GHC.hs
+++ b/src/GHC.hs
@@ -107,9 +107,10 @@ stage2Packages = return [haddock]
 testsuitePackages :: Action [Package]
 testsuitePackages = return [ checkApiAnnotations
                            , checkPpr
+                           , ghci
                            , ghcPkg
-                           , parallel
                            , hp2ps
+                           , parallel
                            , timeout         ]
 
 -- | Given a 'Context', compute the name of the program that is built in it
--- a/src/Oracles/TestSettings.hs
+++ b/src/Oracles/TestSettings.hs
@@ -1,3 +1,7 @@
+-- | We create a file <root>/test/ghcconfig containing configuration of test
+-- | compiler. We need to search this file for required keys and setting 
+-- | required for testsuite e.g. WORDSIZE, HOSTOS etc.  
+
 module Oracles.TestSettings (
     TestSetting (..), testSetting, testRTSSettings
     ) where
@@ -8,7 +12,7 @@ import Base
 testConfigFile :: Action FilePath
 testConfigFile = buildRoot <&> (-/- "test/ghcconfig")
 
--- | Test settings obtained from ghcConfig
+-- | Test settings that are obtained from ghcconfig file.
 data TestSetting = TestHostOS
                  | TestWORDSIZE
                  | TestTARGETPLATFORM
@@ -33,6 +37,7 @@ data TestSetting = TestHostOS
                  deriving (Show)
 
 -- | Lookup for testsettings in ghcconfig file
+-- | To obtain RTS Ways supported in ghcconfig file, use testRTSSettings.
 testSetting :: TestSetting -> Action String
 testSetting key = do
     file <- testConfigFile
--- a/src/Oracles/TestSettings.hs
+++ b/src/Oracles/TestSettings.hs
@@ -1,0 +1,66 @@
+module Oracles.TestSettings (
+    TestSetting (..), testSetting, testRTSSettings
+    ) where
+
+import Hadrian.Oracles.TextFile
+import Base
+
+testConfigFile :: Action FilePath
+testConfigFile = buildRoot <&> (-/- "test/ghcconfig")
+
+-- | Test settings obtained from 
+data TestSetting = TestHostOS
+                 | TestWORDSIZE
+                 | TestTARGETPLATFORM
+                 | TestTargetOS_CPP
+                 | TestTargetARCH_CPP
+                 | TestGhcStage
+                 | TestGhcDebugged
+                 | TestGhcWithNativeCodeGen
+                 | TestGhcWithInterpreter
+                 | TestGhcUnregisterised
+                 | TestGhcWithSMP
+                 | TestGhcDynamicByDefault
+                 | TestGhcDynamic
+                 | TestGhcProfiled
+                 | TestAR
+                 | TestCLANG
+                 | TestLLC
+                 | TestTEST_CC
+                 | TestGhcPackageDbFlag
+                 | TestMinGhcVersion711
+                 | TestMinGhcVersion801
+                 deriving (Show)
+
+testSetting :: TestSetting -> Action String
+testSetting key = do
+    file <- testConfigFile
+    lookupValueOrError file $ case key of
+        TestHostOS                -> "HostOS"
+        TestWORDSIZE              -> "WORDSIZE" 
+        TestTARGETPLATFORM        -> "TARGETPLATFORM"
+        TestTargetOS_CPP          -> "TargetOS_CPP"
+        TestTargetARCH_CPP        -> "TargetARCH_CPP"
+        TestGhcStage              -> "GhcStage" 
+        TestGhcDebugged           -> "GhcDebugged"
+        TestGhcWithNativeCodeGen  -> "GhcWithNativeCodeGen"
+        TestGhcWithInterpreter    -> "GhcWithInterpreter"
+        TestGhcUnregisterised     -> "GhcUnregisterised"
+        TestGhcWithSMP            -> "GhcWithSMP"
+        TestGhcDynamicByDefault   -> "GhcDynamicByDefault"
+        TestGhcDynamic            -> "GhcDynamic"
+        TestGhcProfiled           -> "GhcProfiled"
+        TestAR                    -> "AR"
+        TestCLANG                 -> "CLANG"
+        TestLLC                   -> "LLC"
+        TestTEST_CC               -> "TEST_CC"
+        TestGhcPackageDbFlag      -> "GhcPackageDbFlag"
+        TestMinGhcVersion711      -> "MinGhcVersion711"
+        TestMinGhcVersion801      -> "MinGhcVersion801"
+    
+
+testRTSSettings :: Action [String]
+testRTSSettings = do 
+    file <- testConfigFile
+    fmap words $ lookupValueOrError file "GhcRTSWays"
+
--- a/src/Oracles/TestSettings.hs
+++ b/src/Oracles/TestSettings.hs
@@ -8,7 +8,7 @@ import Base
 testConfigFile :: Action FilePath
 testConfigFile = buildRoot <&> (-/- "test/ghcconfig")
 
--- | Test settings obtained from 
+-- | Test settings obtained from ghcConfig
 data TestSetting = TestHostOS
                  | TestWORDSIZE
                  | TestTARGETPLATFORM
@@ -32,6 +32,7 @@ data TestSetting = TestHostOS
                  | TestMinGhcVersion801
                  deriving (Show)
 
+-- | Lookup for testsettings in ghcconfig file
 testSetting :: TestSetting -> Action String
 testSetting key = do
     file <- testConfigFile
@@ -59,6 +60,7 @@ testSetting key = do
         TestMinGhcVersion801      -> "MinGhcVersion801"
     
 
+-- | Get the RTS ways of the test compiler
 testRTSSettings :: Action [String]
 testRTSSettings = do 
     file <- testConfigFile
--- a/src/Rules/Test.hs
+++ b/src/Rules/Test.hs
@@ -20,13 +20,13 @@ testRules = do
     -- | Using program shipped with testsuite to generate ghcconfig file.
     root -/- ghcConfigProgPath ~> do
         ghc             <- builderPath $ Ghc CompileHs Stage0
-        command [] ghc [ghcConfigHsPath, "-o" , root -/- ghcConfigProgPath]
+        cmd ghc [ghcConfigHsPath, "-o" , root -/- ghcConfigProgPath]
  
     -- | TODO : Use input test compiler and not just stage2 compiler.  
     root -/- ghcConfigPath ~> do
         ghcPath         <- needfile Stage1 ghc
         need [ root -/- ghcConfigProgPath]
-        command [FileStdout $ root -/- ghcConfigPath] (root -/- ghcConfigProgPath)
+        cmd [FileStdout $ root -/- ghcConfigPath] (root -/- ghcConfigProgPath)
             [ ghcPath  ] 
 
     "validate" ~> do
@@ -66,8 +66,8 @@ testRules = do
         buildWithCmdOptions env $ target (vanillaContext Stage2 compiler) RunTest [] []
 
 -- | Build extra programs and libraries required by testsuite
-needTestsuiteLibraries :: Action ()
-needTestsuiteLibraries = do
+needTestsuitePackages :: Action ()
+needTestsuitePackages = do
     targets        <- mapM (needfile Stage1) =<< testsuitePackages
     binPath        <- stageBinPath Stage1
     libPath        <- stageLibPath Stage1
@@ -109,7 +109,7 @@ needTestBuilders = do
     needBuilder Hpc
     needBuilder (Hsc2Hs Stage1)
     timeoutProgBuilder
-    needTestsuiteLibraries
+    needTestsuitePackages
 
 -- | Extra flags to send to the Haskell compiler to run tests.
 runTestGhcFlags :: Action String
@@ -160,3 +160,4 @@ needfile stage pkg
 -- we are going to use, I suppose?
     | isLibrary pkg = pkgConfFile (Context stage pkg profilingDynamic)
     | otherwise = programPath =<< programContext stage pkg
+
--- a/src/Rules/Test.hs
+++ b/src/Rules/Test.hs
@@ -29,12 +29,12 @@ testRules = do
             , builderEnvironment "TEST_HC" $ Ghc CompileHs Stage2
             , AddEnv "TEST_HC_OPTS" <$> runTestGhcFlags ]
 
-        makePath       <- builderPath $ Make ""
-        top            <- topDirectory
-        ghcPath        <- (top -/-) <$> builderPath (Ghc CompileHs Stage2)
-        ghcFlags       <- runTestGhcFlags
-        checkPprPath   <- (top -/-) <$> expr (needfile Stage1 checkPpr)
-        checkPprPath   <- (top -/-) <$> expr (needfile Stage1 checkApiAnnotations)
+        makePath        <- builderPath $ Make ""
+        top             <- topDirectory
+        ghcPath         <- (top -/-) <$> builderPath (Ghc CompileHs Stage2)
+        ghcFlags        <- runTestGhcFlags
+        checkPprPath    <- (top -/-) <$> needfile Stage1 checkPpr
+        annotationsPath <- (top -/-) <$> needfile Stage1 checkApiAnnotations
 
         -- Set environment variables for test's Makefile.
         liftIO $ do
@@ -50,8 +50,14 @@ testRules = do
 -- | Build extra programs required by testsuite
 needTestsuiteBuilders :: Action ()
 needTestsuiteBuilders = do
-    targets <- mapM (needfile Stage1) =<< testsuitePackages
+    targets        <- mapM (needfile Stage1) =<< testsuitePackages
+    binPath        <- stageBinPath Stage1
+    libPath        <- stageLibPath Stage1
+    iservPath      <- needfile Stage1 iserv 
+    runhaskellPath <- needfile Stage1 runGhc
     need targets
+    copyFile iservPath $ libPath -/- "bin/ghc-iserv"
+    copyFile runhaskellPath $ binPath -/- "runghc"
 
 -- | Build the timeout program.
 -- See: https://github.com/ghc/ghc/blob/master/testsuite/timeout/Makefile#L23
--- a/src/Rules/Test.hs
+++ b/src/Rules/Test.hs
@@ -15,12 +15,30 @@ import System.Environment
 -- TODO: clean up after testing
 testRules :: Rules ()
 testRules = do
+    root <- buildRootRules
+
+    -- | Using program shipped with testsuite to generate ghcconfig file.
+    root -/- ghcConfigProgPath ~> do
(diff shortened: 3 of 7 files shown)


src/GHC.hs
```diff
@@ -107,9 +107,10 @@ stage2Packages = return [haddock]
 testsuitePackages :: Action [Package]
 testsuitePackages = return [ checkApiAnnotations
```
snowleopard: We have this list of packages needed for the testsuite, but then do a lot of ad hoc `need`s in the testsuite instead of fully relying on this list. Can we do better?

src/Rules/Test.hs
```diff
@@ -50,33 +50,35 @@ testRules = do
...
 
 -- | Build extra programs required by testsuite
 needTestsuiteBuilders :: Action ()
 needTestsuiteBuilders = do
-    targets <- mapM (needfile Stage1) =<< testsuitePackages
+    targets        <- mapM (needfile Stage1) =<< testsuitePackages
+    binPath        <- stageBinPath Stage1
+    libPath        <- stageLibPath Stage1
+    iservPath      <- needfile Stage1 iserv 
+    runhaskellPath <- needfile Stage1 runGhc
```
snowleopard: Shall `iserv` and `runGhc` be added to `testsuitePackages`?

src/Rules/Test.hs
```diff
@@ -50,33 +50,35 @@ testRules = do
...
             setEnv "MAKE" makePath
             setEnv "TEST_HC" ghcPath
             setEnv "TEST_HC_OPTS" ghcFlags
+            setEnv "CHECK_PPR" checkPprPath
+            setEnv "CHECK_API_ANNOTATIONS" annotationsPath 
 
         -- Execute the test target.
         buildWithCmdOptions env $ target (vanillaContext Stage2 compiler) RunTest [] []
 
 -- | Build extra programs required by testsuite
```
snowleopard: I've already commented on this: the name `needTestsuiteBuilders` is misleading -- this function deals not only with "builders" but also with libraries. It should therefore be renamed, perhaps to `needTestsuitePackages`.

src/Rules/Test.hs
```diff
@@ -50,33 +50,35 @@ testRules = do
...
     need targets
-  where
-    needfile :: Stage -> Package -> Action FilePath
-    needfile stage pkg
-      -- TODO (Alp): we might sometimes need more than vanilla!
-      -- This should therefore depend on what test ways
-      -- we are going to use, I suppose?
-      | isLibrary pkg = pkgConfFile (vanillaContext stage pkg)
-      | otherwise = programPath =<< programContext stage pkg
+    copyFile iservPath $ libPath -/- "bin/ghc-iserv"
```
snowleopard: Why do we need to do the copying? We should add a comment here to clarify the logic.

snowleopard:
@chitrak7 Thank you! Please see my comments above.

src/Rules/Test.hs
chitrak7: Yes, It we need iserv in testsuite packages as we need to build its profiling way. 
runGhc is already in stage1 and I dont't think we need to build any other way here. But we can add it just to be sure. What are your suggestions @snowleopard .

src/Rules/Test.hs
chitrak7: Yes, updating in a revision.

src/Rules/Test.hs
snowleopard: The `testsuitePackages` should contain *all* packages required for the testsuite. That's what the name suggests!

snowleopard: Aha, I see where the confusion can be coming from. The comment says: "Packages that are built only for the testsuite", but I think it should actually say "Packages that are required for the testsuite".

src/GHC.hs
chitrak7: Actually, default libraries are built vanilla way, but the libraries listed here are needed in profiling and dynamic ways also. Can we create a separate list of packages needed only  for testsuite and have to built in profiling and dynamic ways.

src/Rules/Test.hs
chitrak7: Resolved. There are still few questions regarding this issue but I think I should create a separate issue for that.

src/GHC.hs
snowleopard: So, maybe this list should list ways too?

I think someone needs to think about the overall architecture of the testsuite. So far there are a bunch of PRs that seem to introduce various ad hoc solutions, but I do not see a clear design of how everything should be tied together. Can you write up such a design document?

src/Settings/Builders/GhcCabal.hs
```diff
@@ -57,21 +57,25 @@ ghcCabalBuilderArgs = mconcat
 -- TODO: Isn't vanilla always built? If yes, some conditions are redundant.
 -- TODO: Need compiler_stage1_CONFIGURE_OPTS += --disable-library-for-ghci?
 -- TODO: should `elem` be `wayUnit`?
+-- This approach still doesn't work. Previously libraries were build only in the 
+-- Default flavours and not using context.
```
snowleopard: This might actually be dead code, since we no longer call `ghc-cabal`. Are you sure this function is actually executed?

chitrak7: Oh, maybe this is why this approach is not working. Can you please tell me how exactly we are configuring cabal files then!

snowleopard: @chitrak7 I'll try to find time today to remove dead code. You should look into the module `Hadrian.Haskell.Cabal.Parse`.

src/GHC.hs
chitrak7: Yes, I will be creating an Issue in this regards soon enough.

chitrak7:
@snowleopard does ghc always make vanilla?

snowleopard:
> does ghc always make vanilla?

@chitrak7 Yes, I think it does.

chitrak7:
@snowleopard I find the logic testsuite uses to find which ways libraries are built quite weak. It simply checks for  the presence/absence of PrimoWrappers.hi/dyn_hi/p_hi. Should I use the same logic also??
Is there any other way to do this?

snowleopard:
@chitrak7 Why don't you look at the contents of `ways`?

alpmestan:
Ideally I think we really want to examine the `libraryWays` and `rtsWays` of the `Flavour` we are running with. If you find that for some reason this is not reliable enough (or falls short in some cases), well first please let me know because I'm curious about this, but we can then simply go with the slightly more "hacky" strategy of checking the existence of build products with all those way-specific extension to figure out what we have actually built. Sounds good to everyone?

snowleopard:
@alpmestan As I mentioned elsewhere, I think we might want to add a new build flavour, specifically for the testsuite, listing precisely all the ways that we'd like to test (and therefore build).

In other words, the build for testing is different from a usual (default) build: not only we build a different collection of packages, we might also want to specify precisely which ways we want.

chitrak7:
@snowleopard I have added fore features to this PR. 

> Why don't you look at the contents of ways?

Can you please explain what you mean when you say this. Exactly whose ways are we supposed to look?

alpmestan:
> @alpmestan As I mentioned elsewhere, I think we might want to add a new build flavour, specifically for the testsuite, listing precisely all the ways that we'd like to test (and therefore build).

I'm not sure I agree with this. The behaviour of the Make build system here, when you do say `make slowtest`, is to run all the tests that are not skipped given the configuration/build artifacts available. So if you don't have profiling libs, all tests that require them will be skipped (which is different from failing, they're just not considered for execution, like Windows tests are not considered for execution on Linux). I would therefore have thought that the `--flavour` could dictate what we produce and that the testsuite could just go with that, skipping all the tests that require other things.

This by the way does not exclude having a dedicated flavour that just has _everything_ the testsuite needs.

chitrak7:
@snowleopard I think I agree with Alp on this one. We should instead modify Hadrian to read input test ways and build libraries based on them. This is what I was trying to fix through when I encountered #626. Aside from that, I favor using input flavors and build flavors to build libraries instead of chec

chitrak7:
@snowleopard I think I am finished with major part of this issue. I will handle library ways in a later PR. Can you review this one at present.

src/GHC.hs
alpmestan: Yes we need to be a _lot_ more careful about those uses of `vanillaContext` and default ways for packages etc. Most of the time we just want to follow what `libraryWays` and `rtsWays` dictate, I think. I'm expecting this single line of work to fix a _lot_ of test failures.

src/Settings/Builders/RunTest.hs
```diff
@@ -23,23 +28,27 @@ runTestBuilderArgs = builder RunTest ? do
...
-    unregisterised    <- getFlag GhcUnregisterised
-    withSMP           <- expr ghcWithSMP
+    hasProfiling  <- expr $ fmap stringToBool (testSetting TestGhcProfiled) 
+    hasDynamic    <- expr $ fmap stringToBool (testSetting TestGhcDynamic)
+    hasDebugged   <- expr $ fmap stringToBool (testSetting TestGhcDebugged)
+    hasDynamicByDefault <- expr $ fmap stringToBool (testSetting TestGhcDynamicByDefault)
+    withNativeCodeGen <- expr $ fmap stringToBool (testSetting TestGhcWithNativeCodeGen)
+    withInterpreter   <- expr $ fmap stringToBool (testSetting TestGhcWithInterpreter)
+    unregisterised    <- expr $ fmap stringToBool (testSetting TestGhcUnregisterised)
+    withSMP           <- expr $ fmap stringToBool (testSetting TestGhcWithSMP)
```
alpmestan: I _think_ it's safe to say that we might want a shorter name for `expr $ fmap stringToBool (testSetting ...)` (or just the `fmap stringToBool (testSetting ...)` part?) =)

src/Rules/Test.hs
```diff
@@ -15,47 +15,70 @@ import System.Environment
...
+    root -/- ghcConfigProgPath ~> do
+        ghc             <- builderPath $ Ghc CompileHs Stage0
+        command [] ghc [ghcConfigHsPath, "-o" , root -/- ghcConfigProgPath]
+ 
+    -- | TODO : Use input test compiler and not just stage2 compiler.  
+    root -/- ghcConfigPath ~> do
+        ghcPath         <- needfile Stage1 ghc
+        need [ root -/- ghcConfigProgPath]
+        command [FileStdout $ root -/- ghcConfigPath] (root -/- ghcConfigProgPath)
+            [ ghcPath  ]
```
sighingnow: I think the variable `ghcPath` may be incorrect. In ghc's `testsuite/mk/boilerplate.mk` we pass the `$(TEST_HC)` as argument to `ghc-config`. However in the `"test"` rule, we use stage2 compiler as `TEST_HC`. I think here the consistency does matter.

Ref: https://github.com/ghc/ghc/blob/master/testsuite/mk/boilerplate.mk#L240

src/Oracles/TestSettings.hs
```diff
@@ -0,0 +1,68 @@
...
+        TestGhcPackageDbFlag      -> "GhcPackageDbFlag"
+        TestMinGhcVersion711      -> "MinGhcVersion711"
+        TestMinGhcVersion801      -> "MinGhcVersion801"
+    
+
+-- | Get the RTS ways of the test compiler
+testRTSSettings :: Action [String]
+testRTSSettings = do 
+    file <- testConfigFile
+    fmap words $ lookupValueOrError file "GhcRTSWays"
```
sighingnow: I think the `testRTSSettings` is just a case of `testSetting`.

How about add a `TestGhcRTSWays` constructor to `TestSetting` and move the `lookupValueOrError file "GhcRTSWays"` part into `testSetting`, then the `testRTSSettings` will be:

```
testRTSSettings :: Action [String]
testRTSSettings = words <$> testSetting TestGhcRTSWays
```

More concise.

src/Rules/Test.hs
chitrak7: Here also, the stage2 compiler is used only. We just build stage2 compiler in stage1 and this is why I have used stage1.

src/Oracles/TestSettings.hs
chitrak7: @sighingnow I just refrained from using RTSSetting with testSetting because someone might use testSetting instead of testRTSSetting and get errors. I wanted to separate RTSWays completely from testSetting so as no confusion is created.

sighingnow: I see. Thanks for making this point clear to me.

src/Rules/Test.hs
sighingnow: In `need needTestBuilders` we ensure the stage2 ghc is ready before run test. The stage1 and stage2 compiler are often built in different ways (for example, with or without `-prof`). We are testing again the stage2 compiler, I think it does matter to use stage2 compiler to get these test settings (for example, `GhcWithInterpreter`, `GhcProfiled` and `GhcDebugged`).

> We just build stage2 compiler in stage1 and this is why I have used stage1.

I think we could `needBuilder $ Ghc CompileHs Stage2` in rule `root -/- ghcConfigPath`, before executing the `ghc-config` command.

chitrak7: @sighingnow when I call needfile Stage1 ghc, I call the path of stage2 compiler only. Now, stage2 compiler is built in _build/stage1/bin/ghc. Hence, I am still pointing to stage2 compiler only.

sighingnow: Yes you are correct. Thanks for the explanation. I just found that hadrain places the stage2 ghc inside the `stage1` folder....

Now this PR looks good to me.

snowleopard: Thank you for the review @sighingnow!

snowleopard:
@alpmestan @chitrak7 OK, you convinced me :)

Then, what about producing a file, e.g. `_build/flavour`, after each successful build that essentially provides all necessary information for the testsuite (or other possible consumers in the toolchain)? This will make it possible for the user to run Hadrian using their custom build flavour and then run the testsuite, which will pick up the correct build configuration without relying on hacks like checking for `p_dyn` objects? This will also not tie the testsuite only to the default build flavour.

snowleopard:
An alternative is then, which you seem to suggest, is to require the user to run the testsuite with exactly the same flavour as during the build.

For example, `build --flavour=user` must be followed by `build test --flavour=user`. This might work too.

src/Oracles/TestSettings.hs
```diff
@@ -0,0 +1,68 @@
...
+                 | TestAR
+                 | TestCLANG
+                 | TestLLC
+                 | TestTEST_CC
+                 | TestGhcPackageDbFlag
+                 | TestMinGhcVersion711
+                 | TestMinGhcVersion801
+                 deriving (Show)
+
+-- | Lookup for testsettings in ghcconfig file
```
snowleopard: Please provide the full path to the `ghcconfig` file.

src/Rules/Test.hs
```diff
@@ -15,47 +15,70 @@ import System.Environment
 -- TODO: clean up after testing
 testRules :: Rules ()
 testRules = do
+    root <- buildRootRules
+
+    -- | Using program shipped with testsuite to generate ghcconfig file.
+    root -/- ghcConfigProgPath ~> do
+        ghc             <- builderPath $ Ghc CompileHs Stage0
+        command [] ghc [ghcConfigHsPath, "-o" , root -/- ghcConfigProgPath]
```
snowleopard: Could you switch to using `cmd` instead of `command` as in the rest of Hadrian?

src/Rules/Test.hs
```diff
@@ -15,47 +15,70 @@ import System.Environment
...
         -- Execute the test target.
         buildWithCmdOptions env $ target (vanillaContext Stage2 compiler) RunTest [] []
 
--- | Build extra programs required by testsuite
-needTestsuiteBuilders :: Action ()
-needTestsuiteBuilders = do
-    targets <- mapM (needfile Stage1) =<< testsuitePackages
+-- | Build extra programs and libraries required by testsuite
+needTestsuiteLibraries :: Action ()
+needTestsuiteLibraries = do
```
snowleopard: But `needTestsuiteLibraries` is as misleading as `needTestsuiteBuilders`!

Why not `needTestsuitePackages`?

src/Oracles/TestSettings.hs
```diff
@@ -0,0 +1,68 @@
+module Oracles.TestSettings (
+    TestSetting (..), testSetting, testRTSSettings
+    ) where
+
+import Hadrian.Oracles.TextFile
+import Base
+
+testConfigFile :: Action FilePath
```
snowleopard: It would be great to add more comments here about the purpose of this module.

Please describe the purpose of the file `test/ghcconfig`, give a couple of examples of the settings that you read from it, and how they are used by the testsuite.

snowleopard:
@chitrak7 I've added a couple of more comments. Please have a look.

alpmestan:
> An alternative is then, which you seem to suggest, is to require the user to run the testsuite with exactly the same flavour as during the build.
>
> For example, `build --flavour=user` must be followed by `build test --flavour=user`. This might work too.

Right. Or really you'd just do `build test --flavour=something` (where something is an existing or user defined flavour). This is a quite common need. Say you investigate some ticket, write a patch and want to see if that fixes the failure of test `foo`. You can then just go ahead and do `build test --flavour=suitable_flavour --build-root=_test-foo --only=foo` and have a cup of tea/coffee, where `suitable_flavour` should be whatever is appropriate: quickest if the test doesn't need any fancy _way_, devel2/perf/prof/etc depending on what's needed.

One reason why I think this works rather nicely is that this is the most flexible approach. I see two alternatives. The first one, at the opposite end of the spectrum, is to just require all the ways of everything to be built before we can run any test. This is annoying as it will _significantly_ increase the duration of the build. The other one, sitting in the middle, is only conceptual but is the best one. It'd consist in building just what we need for just those tests the user is asking to run (as determined by command line options etc). However the testsuite doesn't really give us a way to discover this ahead of time, as far as I know. I mean, we could scan the .T files and try to inspect the source code to figure some things out but... I don't think we want to do this :-)

Now... in this scheme that I'm suggesting, what if we call the `test` rule again on the same build root but with a different flavour than the one we called it on right before? I don't think this would rebuild things with the way our library/executable rules are implemented at the moment. It feels like we should rebuild everything so as to actually produce all the build artefacts variants that the _new_ flavour implies. A bit like how `cabal-install` does it. Thoughts?

snowleopard:
@alpmestan Yes, I agree, this looks like the right approach indeed.

> Now... in this scheme that I'm suggesting, what if we call the `test` rule again on the same build root but with a different flavour than the one we called it on right before? I don't think this would rebuild things with the way our library/executable rules are implemented at the moment. 

Hmm, are you sure the rebuilds won't happen? We track all command line arguments, so if changing the flavour means a command line changes (e.g. you add `-prof`) then all affected targets will be rebuilt.

sighingnow:
> what if we call the test rule again on the same build root but with a different flavour than the one we called it on right before?

I agree with this idea. `make test` doesn't trigger rebuild event if there are changes in sources. I think we should **decouple the dependency** between the stage1/2 ghc and the `hadrain test` command. We only need to make sure the corresponding ghc compiler exists. If not, we can report error to user and suggest the `hadrian build --flavour=xxx` command.

src/Oracles/TestSettings.hs
chitrak7: @snowleopard Why do we need the full path. The provided path is relative to buildroot.

snowleopard: I'm referring to the comment. You should keep in mind the readers that have no idea what this code is supposed to do. The full path will be helpful for them.

alpmestan:
> Hmm, are you sure the rebuilds won't happen? We track all command line arguments, so if changing the flavour means a command line changes (e.g. you add -prof) then all affected targets will be rebuilt.

@snowleopard I'm not entirely sure, and I'm in the middle of something right now so I can't really give it a shot.

> I agree with this idea. make test doesn't trigger rebuild even if there are changes in sources. I think we should decouple the dependency between the stage1/2 ghc and the hadrain test command. We only need to make sure the corresponding ghc compiler exists. If not, we can report error to user and suggest the hadrian build --flavour=xxx command.

@sighingnow Right. I'll be happy as long as I can start from a fresh source tree and just call `build -c test` and have this command build _a_ GHC. Adding `--flavour` should then refine what gets built and possibly rebuild things, and the test rules should just react accordingly, picking up or dropping tests depending on what build products are available.

Note that this is the default behaviour of `make test` indeed but not of the `validate` script, which by default uses a pre-determined `.mk` file but lets you supply your own. It might be interesting to map this pre-determined build specification to a `Flavour` in hadrian land and use it for the `validate` rule (and just that one, not the `test` rule) when no `--flavour` is specified? We would therefore have `build validate ~ ./validate` and `build test ~ make [slow|fast|full]test`. And supply a `--flavour` explicitly would amount to writing a custom `validate.mk` to be picked up by the `./validate` script.

src/Oracles/TestSettings.hs
chitrak7: Resloved

src/Rules/Test.hs
chitrak7: Resolved

src/Rules/Test.hs
chitrak7: Resolved

src/Oracles/TestSettings.hs
chitrak7: resolved

chitrak7:
@snowleopard I incurred an error which I think is due to #559 and added an additional flag to resolve this. I accidentally committed the change but think might be helpful. Should I remove the argument?

src/Rules/Test.hs
alpmestan: h I just found that hadrain places the stage2 ghc inside the stage1 folder....

Right, because it is _built_ by stage1. It is however indeed confusing at first, but it makes sense to me now.

src/Settings/Builders/Ghc.hs
```diff
@@ -79,6 +79,7 @@ commonGhcArgs = do
             , arg "-hcsuf", arg $ hcsuf way
             , wayGhcArgs
             , packageGhcArgs
+            , arg "-lnuma"
```
chitrak7: @snowleopard I incurred an error which I think is due to #559 and added an additional flag to resolve this. I accidentally committed the change but think might be helpful. Should I remove the argument?

alpmestan: Shouldn't this be dependent on whether we enable NUMA support in `./configure` ?

chitrak7: I supplied no such flag, neither is there any such flag with configure script help. I think it is default setting. Reverting this change and looking further into the matter.

snowleopard: We shouldn't add such flags unconditionally. Better do this in a separate PR.

chitrak7: Resolved

snowleopard:
OK, looks like all comments have been addressed. Shall I merge this?

chitrak7:
@alpmestan @snowleopard Is this pull ready to land.

snowleopard:
@chitrak7 Merged, thanks!

Issue #463, opened 2017-11-16
build.cabal.sh breaks badly if hadrian's dependencies do not successfully install.


duog:
running build.cabal.sh with Cabal <2.1 triggers this code path:
```bash
    if ! ( "$CABAL" sandbox hc-pkg list > /dev/null 2>&1); then
        "$CABAL" sandbox init
        "$CABAL" sandbox add-source ../libraries/Cabal/Cabal
        "$CABAL" install                \
            --dependencies-only         \
            --disable-library-profiling \
            --disable-shared
    fi
```

If dependencies do not cleanly install one must manually delete the sandbox to proceed.

snowleopard:
Thanks @duog! I suppose the right solution is to detect the failure and delete the sandbox?

I'm not familiar with Cabal sandboxes so it will take me some time to dig out the right commands to run. Presumably this is an easy fix, so a PR is very welcome.

chitrak7:
@snowleopard @izgzhen deleting .sandbox file and then rerunning build works whenever I get this error. I believe this should be a good fix here.

izgzhen:
related to #524

PR #633, opened 2018-06-22
Fix cabal sandbox init rule

### LINKED ISSUE #524: build.cabal.sh's condition on sandbox init is too brittle
https://github.com/snowleopard/hadrian/blob/9eef6a7c7cce150f39740beb8570f4cbc38b947e/build.cabal.sh#L61-L68

If I CTRL-C during `install` phase, then continue, some packages will be left uninstalled.

chitrak7:
Fixes #524 #463

@snowleopard @angerman @izgzhen 

Test: Remove the local cabal sandbox, then run ./build.sh. Stop the process mid-installation, and then run the command again. Setup will start by installing the program where we previously stopped.


chitrak7:
Rebased from master.

PR #639, opened 2018-06-27
Getting information of input test compiler

chitrak7:
Earlier, we would get configuration from stage2 compiler irrespective of the test compiler in the input . This PR fixes this.

### DIFF
--- a/src/Rules/Test.hs
+++ b/src/Rules/Test.hs
@@ -1,6 +1,7 @@
-module Rules.Test (testRules, runTestGhcFlags, timeoutProgPath) where
+module Rules.Test (testRules, runTestGhcFlags, timeoutProgPath, compilerPath) where
 
 import Base
+import CommandLine
 import Expression
 import GHC
 import GHC.Packages (timeout)
@@ -22,9 +23,8 @@ testRules = do
         ghc             <- builderPath $ Ghc CompileHs Stage0
         cmd ghc [ghcConfigHsPath, "-o" , root -/- ghcConfigProgPath]
  
-    -- | TODO : Use input test compiler and not just stage2 compiler.  
     root -/- ghcConfigPath ~> do
-        ghcPath         <- needfile Stage1 ghc
+        ghcPath         <- getCompiler
         need [ root -/- ghcConfigProgPath]
         cmd [FileStdout $ root -/- ghcConfigPath] (root -/- ghcConfigProgPath)
             [ ghcPath  ] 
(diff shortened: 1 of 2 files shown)


src/Rules/Test.hs
```diff
@@ -160,4 +160,15 @@ needfile stage pkg
...
     | otherwise = programPath =<< programContext stage pkg
-
+ 
+getCompiler :: Action FilePath
+getCompiler = do
+    args            <- userSetting defaultTestArgs
+    compilerPath $ testCompiler args
+
+-- | Set Test Compiler
+compilerPath :: String -> Action FilePath
```
snowleopard: I'm a bit unhappy about using strings here. Could the first argument be turned into a `Stage`, ideally during parsing of test arguments?

src/Rules/Test.hs
```diff
@@ -22,9 +23,8 @@ testRules = do
         ghc             <- builderPath $ Ghc CompileHs Stage0
         cmd ghc [ghcConfigHsPath, "-o" , root -/- ghcConfigProgPath]
  
-    -- | TODO : Use input test compiler and not just stage2 compiler.  
     root -/- ghcConfigPath ~> do
-        ghcPath         <- needfile Stage1 ghc
+        ghcPath         <- getCompiler
```
snowleopard: Strange whitespace

src/Settings/Builders/RunTest.hs
```diff
@@ -17,6 +16,7 @@ oneZero lbl True = lbl ++ "=1"
 stringToBool :: String -> Bool
 stringToBool "YES"  = True
 stringToBool "NO"   = False
+stringToBool _      = error "Cannot parse string"
```
snowleopard: Let's rename this function to something more obvious, e.g. `parseYesNo`? Also I would prefer to make it type safe by returning `Maybe Bool` and handling possible failures at the use site.

Also, note that we already have something like this here:

https://github.com/snowleopard/hadrian/blob/master/src/Oracles/Flag.hs#L39-L40

As you can see, handling such errors at the use site allows for more informative error messages.

src/Rules/Test.hs
alpmestan: Well, I suppose this could be turned into `compilerPath :: Either Stage FilePath -> Action FilePath`  ?

snowleopard: @alpmestan Indeed, I missed the last case.

snowleopard: @chitrak7 The benefit is in keeping parsing, string manipulation and associated errors in one place. The rest of the codebase should preferably be strongly typed.

snowleopard:
@chitrak7 Are you planning to complete this PR?

PR #646, opened 2018-07-19
Fixes building with stack within recent versions of ghc repo

mgsloan:


mgsloan:
@snowleopard Makes sense!  I've updated it

snowleopard:
Thanks @mgsloan! Merged.

Issue #642, opened 2018-07-12
Support GHCi debugger with -fexternal-interpreter


alpmestan:
See https://phabricator.haskell.org/D4955#inline-38448.

alpmestan:
Should we pass that flag to the `CabalFlags` builder or to the `GhcCabal Conf` one?

snowleopard:
@alpmestan I'd like to get rid of `CabalFlags` at some point (hope to find time for this soon!), so I'd prefer to start switching over to `GhcCabal Conf` which is more consistent with the rest of the infrastructure. But if doing it using `CabalFlags` is much easier, go for it :)

alpmestan:
Done in #648.

PR #648, opened 2018-07-20
implement the equivalent of D4955 for hadrian

alpmestan:
This addresses #642.

@snowleopard I'm using CabalFlags for now to keep it consistent. I also think it's not necessarily a good idea to throw everything at `GhcCabal Conf` instead, since we do not use ghc-cabal anymore. Since it's at least worth discussing, I just went with uniformity and passed the flags through `CabalFlags`.


snowleopard:
@alpmestan Thanks! I'll let CI finish and merge.

Issue #635, opened 2018-06-23
Remove GhcCabal buider


chitrak7:
With #531, GhcCabal builder has become redundant. We have to remove the builder and generate arguments separately.

snowleopard:
Done in #658.

PR #531, opened 2018-03-19
Use Cabal directly in place of ghc-cabal + make build root configurable

alpmestan:
This commit implements two significant changes (that were not easy to
separate):

- Don't use ghc-cabal anymore for getting information about Haskell packages.
  We now instead directly use Cabal-the-library.

- Make the build root configurable. This effectively gets rid of the inplace
  logic and allows us to place _all_ build artefacts in some directory of
  our choice, by passing '--build-root <some path>' to hadrian. The GHCs we
  produce are now relocatable.

The code for this was mostly taken from #445. I have successfully built functional quick/prof/perf/devel2/quick-cross builds of GHC with this branch. I might commit 
another tweak or two as I'm now testing the test/docs rules. But I'm not expecting 
_major_ changes there.

I'm also not clear on what we want to do with the install/wrapper rules, with 
relocatable GHC builds.

PR #658, opened 2018-08-25
Clean up remaining bits related to ghc-cabal

snowleopard:
After merging #531, we still have a few occurrences of the `ghc-cabal` package, as well as the `GhcCabal` builder, which we need to clean up. See #635.


snowleopard:
@alpmestan @angerman @chitrak7 I think I found a good way to merge `GhcCabal` and `CabalFlags`. I replaced them with a new builder `Cabal` with `ConfigurationInfo = Setup | Flags` to distinguish between the two pieces of information -- see the latest commit. Any comments/suggestions?

angerman:
Look good. I’m still stumbling over the `Setup` name though; can’t think of a better name right now though. Something along the line of `SetupHsArgs` might be more descriptive.

Issue #653, opened 2018-08-09
Build with --quickest didn't produce stage1/touchy


ndmitchell:
I built with `hadrian\build.bat -j --flavour=quickest --configure`. I then tried to start `ghci` and got:

```
GHCi, version 8.7.20180808: http://www.haskell.org/ghc/  :? for help
package flags have changed, resetting and loading new packages...
Loaded GHCi configuration from ./utils/ghc-in-ghci/settings.ghci
[  8 of 493] Compiling Fingerprint      ( _build\stage1\compiler\build\Fingerprint.hs, .ghci-objects\Fingerprint.o )
[  9 of 493] Compiling Config           ( _build\stage1\compiler\build\Config.hs, .ghci-objects\Config.o )
[ 10 of 493] Compiling Constants        ( compiler\main\Constants.hs, .ghci-objects\Constants.o )
[ 11 of 493] Compiling Ar               ( compiler\main\Ar.hs, .ghci-objects\Ar.o )
[ 12 of 493] Compiling Hooks[boot]      ( compiler\main\Hooks.hs-boot, .ghci-objects\Hooks.o-boot )
ghc.exe: could not execute: C:\Neil\ghc\sep2018\ghc\_build\stage1\lib/bin/touchy.exe
Loaded GHCi configuration from ./utils/ghc-in-ghci/load-main.ghci
```

I copied `touchy` from `stage0` to `stage1` and it worked. Does Hadrian ever build touchy in stage1? Should it do so in quickest?

snowleopard:
Aha, I didn't realise we need both binaries. We can surely build both, it should be a simple change in `GHC`/`GHC.Packages`.

There is also this issue: https://github.com/snowleopard/hadrian/issues/595. Do we need to build `unlit` twice too? If yes, we should close it.

snowleopard:
Fixed in #659.

Issue #540, opened 2018-03-30
Follow-up to #531


snowleopard:
Here is a list of things that I'd like to look at after #531 has been merged:
- [x] Try to break import loops by adding more `.Type` modules.
- [x] Work on #538 to fix `-c`.
- [x] Figure out where we can replace `buildDir` with `contextDir`. 
- [x] Finish the `Settings/Packages.hs` story.
- [x] Move `ghctags` and `haddock`back to Stage2.
- [x] Clean up remaining `ghc-cabal` bits.
- [x] Clean up remaining installation infrastructure.

alpmestan:
`-c` is fixed (#571).

snowleopard:
@alpmestan Yep, checked the box.

snowleopard:
I finished merging all package-specific arguments into `Settings/Packages.hs` in #622.

snowleopard:
I'm moving Haddock and `ghctags` to Stage2, since this is what Make build system does. I haven't found a justification for moving Haddock to Stage2 in #531. 

@alpmestan If this is wrong, please shout.

alpmestan:
There probably isn't a justification. I don't think this should be a problem. As long as you make sure everything still happily builds, we should be fine :-)

snowleopard:
@alpmestan Thanks, done :)

snowleopard:
I think I'm done with this issue. 

> Figure out where we can replace `buildDir` with `contextDir`.

The answer to this appears to be "No, we need to keep `buildDir`", since the Cabal library insists on keeping build artefacts in the `build` subdirectory.

PR #659, opened 2018-08-26
Fix defaut top level targets

snowleopard:
This does the following:
* Fix the top-level targets, which appear to have been restricted to Stage1 targets only in #531.
* Restore the `stage1Only` logic, remove outdated comments.
* Build `touchy` in Stage1, as requested in #653.
* Move `ghcTags` and `haddock` packages to Stage2, as in the Make build system (#540).
* Keep only Stage0 and Stage1 package databases (we do not need Stage2/Stage3 package databases).

### DIFF
--- a/src/Base.hs
+++ b/src/Base.hs
@@ -82,10 +82,10 @@ generatedDir = "generated"
 generatedPath :: Action FilePath
 generatedPath = buildRoot <&> (-/- generatedDir)
 
--- | Path to the package database for the given stage of GHC,
---   relative to the build root.
+-- | Path to the package database for a given build stage, relative to the build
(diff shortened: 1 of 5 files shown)


alpmestan:
Does this mean cross-compilers don't get their `ghc-tags` (cross-compilers are stage 1 GHCs) anymore ?

snowleopard:
@alpmestan According to [this commit](https://github.com/ghc/ghc/commit/5fb72555f7b7ab67a33583f33ad9160761ca434f) `ghctags` should be built with Stage2 compiler.

snowleopard:
Here is the issue: #661.

PR #663, opened 2018-08-28
Simplify Package data type

snowleopard:
I noticed that all packages we build have Cabal files now, which means we no longer need to distinguish between C and Haskell packages. This leads to a lot of simplifications and removal of unsafe functions.


src/Rules/Documentation.hs
```diff
@@ -131,7 +131,7 @@ allHaddocks :: Action [FilePath]
 allHaddocks = do
     pkgs <- stagePackages Stage1
     sequence [ pkgHaddockFile $ vanillaContext Stage1 pkg
-             | pkg <- pkgs, isLibrary pkg, isHsPackage pkg ]
```
alpmestan: Have you tried building the docs since this removal?

alpmestan:
Given that we test very little on CI, I'm a bit "anxious" about changing a whole lot of things, testing very little of it and just merging without anyone's review.

snowleopard:
Thanks @alpmestan! Yes, you are right, our CI is weak and I should have waited for some feedback on such large refactoring. I will build other targets and report back. What do you mean by "stage2 ghc" target? We do build Stage2 GHC by default.

We should probably put together a list of things-to-test, which cover all Hadrian targets.

snowleopard:
So, the `docs` target is broken, but in a way which doesn't seem to be related to this PR:
```
| Copy directory: utils/haddock/haddock-api/resources/html => _build/docs/html
| Run Sphinx Html: docs/users_guide => _build/docs/html/users_guide
* docs
* _build/docs/html/index.html
* _build/docs/html/libraries/index.html
* _build/docs/html/haddock-bundle.min.js
Error, rule finished running but did not produce file:
  _build/docs/html/haddock-bundle.min.js
```
What's weird is that the file in question `_build/docs/html/haddock-bundle.min.js` does exist! So this looks like some concurrency issue. I've created #664 to track this.

alpmestan:
I think I have an idea for the insufficient CI bit.

Issue #664, opened 2018-08-29
Documentation broken


snowleopard:
As discovered in #663, the documentation target `docs` is currently broken:
```
| Copy directory: utils/haddock/haddock-api/resources/html => _build/docs/html
| Run Sphinx Html: docs/users_guide => _build/docs/html/users_guide
Error when running Shake build system:
* docs
* _build/docs/html/index.html
* _build/docs/html/libraries/index.html
* _build/docs/html/haddock-bundle.min.js
Error, rule finished running but did not produce file:
  _build/docs/html/haddock-bundle.min.js
```
What's weird is that the file in question `_build/docs/html/haddock-bundle.min.js` does exist! So this looks like some concurrency issue.

snowleopard:
Fixed in #665.

PR #665, opened 2018-08-29
Fix documentation rules

snowleopard:
See #664.

I still have some issues locally, but this seems to be related to my LaTeX setup.


snowleopard:
@alpmestan This fixes documentation rules for me. There were a few issues, and one of them was indeed introduced by me in a recent refactoring, so thanks for reminding the check the documentation target!

Issue #667, opened 2018-08-30
Symlink traversing logic in build scripts


snowleopard:
Hadrian build scripts `build.sh`, `build.cabal.sh` etc contain mysterious (to me) symlink-traversing logic, which may be confusing in certain setups, as reported in https://ghc.haskell.org/trac/ghc/ticket/15576.

@hvr I think this logic was introduced by you. I couldn't recall why. Could you please clarify? Do we still need it? If yes, how do we avoid the confusion as described in the above ticket?

I'm tempted to classify this as a bug, since it leads to producing build results in an unexpected place.

alpmestan:
@hvr on IRC:

```
<hvr> I'm trying to rememebr the rationale, but it's likely that we need to know the normalised pathname of where the script lives
<hvr> as cabal needs to be able to locate the .cabal files and all that
<hvr> and passing relative pathnames to cabal is often not a good idea
<hvr> so, if we can ditch support for cabal < 2.2
<hvr> stuff gets more uniform
<hvr> as then we only have to figure out how to support the
<hvr> "$CABAL" new-run hadrian --  --directory "$absoluteRoot/.." "$@"
<hvr> invocation
```

hvr:
To elaborate a bit; the important bit is that we want `build.cabal.sh` to do the right thing independently on what $PWD currently points to; in order to do that, we need 

 1. to make sure that `cabal new-{run,build}` are able to find their (cabal) project root (NB: `cabal new-run` intentionally does *not* change the CWD before invoking the executable)
 2. make sure `hadrian` is able to locate the GHC project root

It's also worth noting that the script could effectively be reduced to a single invocation of `new-run` (assuming we have a trivial `cabal.project` file to help `cabal` out):

```
"$CABAL" --project-file="$absoluteRoot/cabal.project"  new-run hadrian -- --directory "$absoluteRoot/.." "$@"   
```

No need to `cd` anywhere anymore, assuming `hadrian` can (like `cabal`) handle it if it's told where the project root is.

snowleopard:
> we want `build.cabal.sh` to do the right thing independently on what $PWD currently points to 

@hvr Aha, I see. This makes sense, although Hadrian currently doesn't support $PWD-based build targets.

> It's also worth noting that the script could effectively be reduced to a single invocation of `new-run`

Let's do this! I'm happy to sacrifice the support of older versions of Cabal if need be. 

Could you put send a PR?

hvr:
[...]

PR #668, opened 2018-08-30
Fix build scripts

snowleopard:
See #667.


alpmestan:
Which CI job are expecting to be green if this commit doesn't break anything compared to the previous incarnation of the build scripts?

snowleopard:
@alpmestan Windows and Linux jobs should be green, OS should fail with linking.

PR #674, opened 2018-09-04
remove unwanted '-O0's in Settings.Packages

alpmestan:
It turns out we don't see the bugs if we just keep the `-fno-omit-...` etc flags. This also "resolves" several failing tests (see comment in #669) !


alpmestan:
Uh, well, yes, it makes sense. We're taking a compilation time hit in exchange of better execution time.

sighingnow:
If we don't pass `-O0`, we don't need to add the extra flags `-fno-omit-interface-pragmas` and `-fno-ignore-interface-pragmas`.

PR #675, opened 2018-09-05
remove superfluous -fno-{ignore, omit}-interface-pragmas

alpmestan:
As suggested by @sighingnow in https://github.com/snowleopard/hadrian/pull/674#issuecomment-418656150. I ran the testsuite after I wrote this patch, and this doesn't affect _anything_.

### DIFF
--- a/src/Settings/Packages.hs
+++ b/src/Settings/Packages.hs
@@ -27,13 +27,6 @@ packageArgs = do
           -- This fixes the 'unknown symbol stat' issue.
           -- See: https://github.com/snowleopard/hadrian/issues/259.
           , builder (Ghc CompileCWithGhc) ? arg "-optc-O2"
-
-          -- See https://ghc.haskell.org/trac/ghc/ticket/15286 and
-          -- https://phabricator.haskell.org/D4880
-          , builder (Ghc CompileHs) ? mconcat
-             [ input "//Natural.hs" ? pure ["-fno-omit-interface-pragmas"]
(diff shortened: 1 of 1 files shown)


alpmestan:
Hmm, this does re-introduce the "can't use Natural in base" error when built with `quickest` (I forgot to try that, I've mostly been building `perf`-flavoured GHCs lately for the testsuite work).

Should we make this conditional on the flavour for now, until the underlying problem is solved in GHC land, if ever?

https://github.com/snowleopard/hadrian/blob/master/doc/flavours.md#arguments says we are likely to encounter this problem with `quick`, `quickest` and `prof` (since they build with `-O0`). Should emit those flags that this patch removes whenever the flavour is one of those? It's quite a hack, anyone's got a better idea?

snowleopard:
Should we add the workarounds directly into flavour descriptions, e.g. here?

https://github.com/snowleopard/hadrian/blob/master/src/Settings/Flavours/Quickest.hs#L17

alpmestan:
Yes, that would be nice!

We can still filter on the input file etc there, right? We could this way define the list of flags at the toplevel somewhere, with proper docs pointing to the trac ticket, and use that in all the relevant flavour definitions. That does sound better.

snowleopard:
Yep, we can use all predicates there too.

alpmestan:
OK, I'm closing this PR since the solution we're going for is "different enough" to justify a new one.

alpmestan:
Better patch implementing the idea from above in https://github.com/snowleopard/hadrian/pull/676.

PR #676, opened 2018-09-05
Work around the "can't use natural in base" problem on a per-flavour basis

alpmestan:
The only flavours that need the workaround are the ones that build
GHC/{Natural, Num}.hs with -O0, namely 'quick', 'quickest' and 'prof'.
This patches defines the necessary arguments in one place and uses them
in all the aforementionned flavour definitions.

This will allow us to have both quick/quickest/prof builds that come through
as well as an efficient compiler when we want it (with e.g perf), which wasn't
the case before my series of patches for this problem.

### DIFF
--- a/src/Settings/Flavours/Common.hs
+++ b/src/Settings/Flavours/Common.hs
@@ -4,8 +4,8 @@ import Expression
 
 -- See https://ghc.haskell.org/trac/ghc/ticket/15286 and
 -- https://phabricator.haskell.org/D4880
-naturalInBaseFixArgs :: [Args]
-naturalInBaseFixArgs =
+naturalInBaseFixArgs :: Args
+naturalInBaseFixArgs = mconcat
   [ input "//Natural.hs" ? pure ["-fno-omit-interface-pragmas"]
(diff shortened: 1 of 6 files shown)


src/Settings/Flavours/Common.hs
```diff
@@ -0,0 +1,11 @@
+module Settings.Flavours.Common where
```
snowleopard: Shall we have a module dedicated to various temporary hacks and workarounds? It could live at the top `src/ToDo.hs`, and it will be the go-to place for Hadrian maintainers for introducing and -- importantly -- finding forgotten temporary fixes that can now be removed?

alpmestan: I hesitated, but imagine it gets empty, we delete it? And then we have new hacks, we recreate it? So I just went with the very obvious `Common`, where we can put anything that's shared between flavours, hacks or not :)

snowleopard: OK, let's stick with your naming. I agree, this module may be useful for other common flavour-related settings.

alpmestan: OK great. I'll let you review everything now, I addressed your other bit of feedback.

PR #677, opened 2018-09-06
Make the test rule depend on ghc-compact, to make this library's tests pass

alpmestan:
This commit also removes some leftover from my previous PR, sorry about that.


snowleopard:
Will merge when CI finishes.

alpmestan:
Oh, I didn't state it explicitly, so here it is: this fixes all the failing `ghc-compact` tests (quite likely all of them, since they presumably all need the library).

snowleopard:
CI still timeouts, so merging as is.

@alpmestan Thanks, I've attached the description to the commit.

PR #679, opened 2018-09-07
Fix clean so it works on Windows

ndmitchell:
Otherwise you get a failure that it can't delete the `.shake.lock` file since Shake has it open.


snowleopard:
@ndmitchell I should learn to use `clean` instead of deleting the `_build` folder!

PR #680, opened 2018-09-07
Move to GHC 8.4.3 by upgrading the Stackage snapshot

ndmitchell:
At least compiles Hadrian itself, so this seems like a much better prospect - as it is Hadrian is close to getting outside the GHC support Window of last 2 versions.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,7 +1,7 @@
 # For more information, see: http://docs.haskellstack.org/en/stable/yaml_configuration.html
 
 # Specifies the GHC version and set of packages available (e.g., lts-3.5, nightly-2015-09-21, ghc-7.10.2)
-resolver: lts-10.8
+resolver: lts-12.9
 
 # Local packages, usually specified by relative directory name
 packages:


snowleopard:
Thanks @ndmitchell! GHC should be buildable with the previous two major GHC releases, and I guess Hadrian should be buildable by the previous two major GHC releases too. This currently includes GHC 8.2.*. 

Should we bump to `lts-11.22` instead until GHC 8.6.1 is out?

ndmitchell:
Yes, Hadrian should be buildable with 8.2, and it is. But why make people use an older GHC than is necessary? Using the latest GHC seems the right default.

ndmitchell:
Shouldn't newer GHC's compile faster and print out less warnings? I guess I can see it both ways so whatever you prefer.

snowleopard:
> Shouldn't newer GHC's compile faster and print out less warnings?

I'm not sure why it would compile faster (newer GHCs aren't always faster), but I agree about warnings -- that's a good point.

Issue #671, opened 2018-09-02
Reduce the number of times a Cabal file is parsed


snowleopard:
When I do zero build with Hadrian in verbose mode, I see the following:
```
| Executable found: bash => C:/msys/usr/bin/bash.exe
| KeyValue oracle: reading 'hadrian/cfg/system.config'...
| Building Stage1 libraries: binary, Cabal, ghc, ghc-boot, ghc-boot-th, ghc-heap, ghci, hpc, mtl, parsec, template-haskell, text, transformers, array, base, bytestring, containers, deepseq, directory, filepath, ghc-compact, ghc-prim, haskeline, integer-simple, pretty, process, rts, stm, time, xhtml, Win32
| Building Stage1 programs : ghc, ghc-pkg, hsc2hs, unlit, hpc, runghc, touchy

| Windows path mapping: /mingw64/bin/ => C:/msys/mingw64/bin/
| CabalFile oracle: reading 'libraries/binary/binary.cabal' (Stage: stage0)...
| CabalFile oracle: reading 'libraries/Cabal/Cabal/Cabal.cabal' (Stage: stage0)...
| CabalFile oracle: reading 'compiler/ghc.cabal' (Stage: stage0)...
| CabalFile oracle: reading 'libraries/ghc-boot/ghc-boot.cabal' (Stage: stage0)...
| CabalFile oracle: reading 'libraries/ghc-boot-th/ghc-boot-th.cabal' (Stage: stage0)...
| CabalFile oracle: reading 'libraries/ghc-heap/ghc-heap.cabal' (Stage: stage0)...
| CabalFile oracle: reading 'libraries/ghci/ghci.cabal' (Stage: stage0)...
| CabalFile oracle: reading 'libraries/hpc/hpc.cabal' (Stage: stage0)...
| CabalFile oracle: reading 'libraries/mtl/mtl.cabal' (Stage: stage0)...
| CabalFile oracle: reading 'libraries/parsec/parsec.cabal' (Stage: stage0)...
| CabalFile oracle: reading 'libraries/template-haskell/template-haskell.cabal' (Stage: stage0)...
| CabalFile oracle: reading 'libraries/text/text.cabal' (Stage: stage0)...
Warning: libraries\text\text.cabal:4:1: The field "bug-reports" is specified
more than once at positions 4:1, 42:1
| CabalFile oracle: reading 'libraries/transformers/transformers.cabal' (Stage: stage0)...
[...]

Issue #550, opened 2018-04-03
Cache package metadata parsing


snowleopard:
When I run Hadrian on my machine, the first minute or so is spent on parsing Cabal files:

```bash
$ ./build.bat -j --flavour=quickest --integer-simple --verbose
hadrian-0.1.0.0: configure (exe)
hadrian-0.1.0.0: build (exe)
hadrian-0.1.0.0: copy/register
Log files have been written to: C:\msys\home\nam83\ghc\hadrian\.stack-work\logs\
| KeyValue oracle: reading 'hadrian/cfg/system.config'...
Building stage2
| Building Programs:  ghc, ghc-pkg, hsc2hs, unlit, haddock, runhaskell, hpc
| Building Libraries: binary, Cabal, ghc, ghc-boot, ghc-boot-th, ghci, hpc, mtl, parsec, template-haskell, text, transformers, array, base, bytestring, containers, deepseq, directory, filepath, ghc-compact, ghc-prim, haskeline, integer-simple, pretty, process, rts, stm, time, xhtml, Win32

| CabalFile oracle: reading 'libraries/binary/binary.cabal' (Stage: stage1)...
| CabalFile oracle: reading 'libraries/Cabal/Cabal/Cabal.cabal' (Stage: stage1)...
| CabalFile oracle: reading 'compiler/ghc.cabal' (Stage: stage1)...
...
```

This used to be cached in `package-data.mk` files substantially speeding up subsequent builds, but now this seems to be a constant cost for any build, unless I'm mistaken.

(The build soon fails on my machine due to an unrelated issue, so I can't try a zero build, but I assume there is no caching at the moment.)

@angerman @alpmestan Do you have any plans on how to cache this?

alpmestan:
@snowleopard I think both of us were hoping there was caching going on here. Maybe it happens within a build but not accross builds? I'm really not sure -- the oracle system comes with caching no? So if we ask twice for the Cabal data of a package, hadrian will only hit the `.cabal` file once, the first time, right?

(I'm talking about these oracles: https://github.com/snowleopard/hadrian/blob/master/src/Hadrian/Oracles/TextFile.hs#L101 -- "instantiated" below in `textFileOracle`.)

snowleopard:
@alpmestan I think the `newCache` combinator only caches computations within a single build. 

I remember that Shake now provides some new/nicer infrastructure for storing arbitrary information in the Shake database. @ndmitchell Could you give a pointer?

[...]

PR #692, opened 2018-09-18
[WIP] Rework parsing Cabal metadata

snowleopard:
The current implementation of parsing Cabal data is a big mess, which affects performance as described in #671 and #550. In this PR I intend to simplify the implementation and avoid unnecessary reparsing.

This is in a very draft state right now and not ready for review. I'm pushing to do a CI run.


alpmestan:
Since this is what we're after here, do you think you could quantify the improvement that this patch makes, **once it's ready for review** (i.e: no rush, I understand it's not quite ready yet) ? Perhaps give the timings for no-op runs? If you don't mind.

snowleopard:
> do you think you could quantify the improvement that this patch makes, 

@alpmestan Yes, absolutely! I'll try to quantify the impact both on the full build and on the zero build. Presumably the impact on the full build will be negligible, and I hope to get back to a few seconds for the zero build (compared to current 30 seconds).

I'll aim to finalise this PR by end of today.

snowleopard:
@alpmestan Current status: zero build down to 10 seconds. Changes to full build time seem insignificant.

There is still some work to be done though.

alpmestan:
@snowleopard Ah, that's already much more reasonable, good work! I certainly won't stand in your way if you want to reduce this further. :)

alpmestan:
We certainly didn't see those on my last few PRs, did we? In which case it'd quite likely be related to commits recently merged to HEAD.

Any chance https://github.com/ghc/ghc/commit/ce3897ffd6e7c8b8f36b8e920168bac8c7f836ae is causing them?

snowleopard:
> any chance ghc/ghc@ce3897f is causing them?

This was my first guess too, but it doesn't seem to touch any sensitive code -- it essentially fixes a bug in the configure scripts, and in the worst case the build system would fail during the configuration.

alpmestan:
Any chance this fixes your problem: https://phabricator.haskell.org/D5163 ?

snowleopard:
@alpmestan Aha, that's it! Thank you.

snowleopard:
P.S.: Sorry, got distracted from finishing this PR at work. Will be back in action tomorrow.

snowleopard:
Got zero build down to 8 seconds but at the cost of one extra oracle. Looks a bit messy...

alpmestan:
Well then you're apparently saving enough time to get us below the max duration :-)

I'll quite likely review this on monday. Can't wait to see it in action.

src/Hadrian/Oracles/TextFile.hs
```diff
@@ -73,12 +72,37 @@ lookupDependencies depFile file = do
...
+    deriving (Binary, Eq, Hashable, NFData, Show, Typeable)
+type instance RuleResult KeyValue = Maybe String
+
+newtype KeyValues = KeyValues (FilePath, String)
+    deriving (Binary, Eq, Hashable, NFData, Show, Typeable)
+type instance RuleResult KeyValues = Maybe [String]
+
+-- | This oracle reads and parses text files to answer various queries, caching
+-- and tracking the results.
+textFileOracle :: Rules ()
```
alpmestan: Same as for the other caches, it'd be nice to get a sliiiightly more detailed overview. "Text files" just seems a little too vague. It'd probably be a good idea to mention which text files this actually gets used on. (.cabal files I suppose, but maybe others?)

snowleopard:
@alpmestan Many thanks for the review -- I fully agree about comments. Will address all your comments tomorrow!

src/Hadrian/Oracles/Cabal/Type.hs
```diff
@@ -23,10 +23,22 @@ import Hadrian.Haskell.Cabal.Type
 import Hadrian.Package
 import Stage
 
+-- | This type of oracle key is used by 'Hadrian.Oracles.Cabal.readPackageData'
+-- to cache reading and parseing of 'Hadrian.Haskell.Cabal.Type.PackageData'.
```
alpmestan: parsing

src/Hadrian/Oracles/Cabal/Type.hs
```diff
@@ -23,10 +23,22 @@ import Hadrian.Haskell.Cabal.Type
...
 newtype PackageDataKey = PackageDataKey Package
     deriving (Binary, Eq, Hashable, NFData, Show, Typeable)
 type instance RuleResult PackageDataKey = PackageData
 
+-- | This type of oracle key is used by 'Hadrian.Oracles.Cabal.readContextData'
+-- to cache reading and parseing of 'Hadrian.Haskell.Cabal.Type.ContextData'.
```
alpmestan: parsing

snowleopard: Thanks, fixed both typos.

snowleopard:
@alpmestan Thank you for the review! Merged.

PR #693, opened 2018-09-21
unlit: build only in stage 0, copy that binary for later stages (fixes #595)

### LINKED ISSUE #595: Don't build unlit twice
Currently we build `unlit` utility twice, [which is probably redundant](https://github.com/snowleopard/hadrian/pull/591#issuecomment-388482754).

Let's build it only in `Stage0`. This is an easy issue, I'll leave it here for a newcomer to try.

alpmestan:


alpmestan:
This seems to cause an error, so it's not ready just yet.

src/Rules/Program.hs
```diff
@@ -55,6 +55,9 @@ buildProgram rs = do
                         (True, s) | s > Stage0 -> do
                             srcDir <- buildRoot <&> (-/- (stageString Stage0 -/- "bin"))
                             copyFile (srcDir -/- takeFileName bin) bin
+                        (False, s) | s > Stage0 && takeFileName bin == "unlit" -> do
```
snowleopard: Should we instead compare packages, i.e. `s > Stage0 && package == unlit`? This should be more robust.

alpmestan: Done.

alpmestan:
So I can just change `package == unlit` to `package == unlit || package == touchy` then, no? That should do the trick, right?

snowleopard:
> So I can just change `package == unlit` to `package == unlit || package == touchy` then, no? That should do the trick, right?

Yes! Or ``package `elem` [unlit, touchy]``.

alpmestan:
Done.

snowleopard:
@alpmestan Great! I'll see what CI says and merge.

PR #694, opened 2018-09-29
Simplify oracles

snowleopard:
As discussed in #550, we can remove some unnecessary occurrences of `newCache`.


snowleopard:
I tested this on my machine and the simplified oracles behave as expected.

PR #703, opened 2018-10-20
Minor revision

snowleopard:
Bumping QuickCheck upper bound, fixing warnings, improving documentation.


snowleopard:
I'm going to merge this PR, since it compiles fine on all CIs and appears to be benign

